PR #22, opened 2020-10-09
Rename Control+ Hub to Technic Hub

laurensvalk:
See discussion in https://github.com/pybricks/support/issues/119


laurensvalk:
Looks like I caught most things, but I wonder if this is stored/cached elsewhere?

> cplushub firmware Expected — Waiting for status to be reported

dlech:
> I wonder if this is stored/cached elsewhere?

I think on Pull Requests that GitHub actions doesn't use the `.github/...` from the checked out code (probably for security reasons). So I think we can just ignore this.

It looks like there are a few mentions of Control+ left in `README.md`, `lib/lego/lego_uart.h` and `lib/pbio/sys/battery.c`.

dlech (approved):

Issue #2, opened 2020-06-23
UART hotplug is broken


dlech:
https://github.com/pybricks/support/issues/43

Currently UART devices can only be detected once. If it is unplugged or goes out of sync, then the hub has to be rebooted.

We need to add some events to communicate between the uartdev and the lpf2_port.

laurensvalk:
This seems to be working fine on the SPIKE Prime Hub. It is still confirmed on Move Hub. (Have not tried the others.)

So if it is partially there already, maybe we don't need a significant overhaul to `uartdev` to fix this after all.

laurensvalk:
Update: This only works if the new device is also a UART device. Hotplugging passive <---> passive is also OK, but passive <---> UART is not. Still, just making hotplugging UART/UART work on all hubs is probably still worth it.

laurensvalk:
UART to UART hotplugging is fixed :tada:.

I'll suggest a patch for UART to passive hotplugging also later today.

laurensvalk:
Patch for passive sensors is pushed also. Pbio tests fail since there isn't a test for `ioport_lpf2` at the moment.

laurensvalk:
@dlech , would you like to review these changes?

dlech:
I made a PR for easier review.

laurensvalk:
A couple of things turned up that I'd like to address also.

- [ ] On Technic Hub, sensors always fail at first, then succeed on the second sync. The easiest way to see this is with the ColorDistanceSensor. I think the error is always `Bad device type id`, so we should be able to deterministically figure out why, and then fix it.
- [ ] I'm currently resetting the ioport to check for passive sensors after every error in `uartdev.c` which is obviously not needed. It would be better to do this only on disconnect or persistent failure to sync up.
- [ ] Avoid using "processes.h" --- see @dlech's comments in the PR. As a first step, I've dropped the commits that messed with the tests and the debug port, instead wrapping the call to the process in an `#if` block for now. This too will be removed and replaced by a call to the right namespace instead.

laurensvalk:
I was looking at fixing the baudrate on syncup, but this requires more changes than I thought, so I think we can defer that to the more rigorous revisions after this release. For now it's probably better if the sensors just work, even if people have to wait a few seconds.

laurensvalk:
Making some good progress here, but new things tend to pop up as we go along

- [ ] Do not auto-power off all power ports (we do this in the motor drivers) because it turns off all sensors too
[...]

PR #32, opened 2020-12-09
Hotplug

dlech:
PR to review changes suggested in #2 .

### DIFF
--- a/lib/pbio/drv/ioport/ioport_lpf2.c
+++ b/lib/pbio/drv/ioport/ioport_lpf2.c
@@ -517,9 +517,9 @@ PROCESS_THREAD(pbdrv_ioport_lpf2_process, ev, data) {
                         iodev->info = info;
 
                         const lump_mode_flags_t *flags = &info->mode_info[0].flags;
-                        iodev->motor_flags = PBIO_IODEV_MOTOR_FLAG_NONE;
+                        iodev->capability_flags = PBIO_IODEV_CAPABILITY_FLAG_NONE;
                         if (flags->flags0 & LUMP_MODE_FLAGS0_MOTOR_POWER) {
-                            iodev->motor_flags |= PBIO_IODEV_MOTOR_FLAG_IS_MOTOR;
+                            iodev->capability_flags |= PBIO_IODEV_CAPABILITY_FLAG_IS_MOTOR;
                         }
                         ioport->iodev = iodev;
                     }
--- a/lib/pbio/drv/ioport/ioport_lpf2.c
+++ b/lib/pbio/drv/ioport/ioport_lpf2.c
@@ -481,35 +481,48 @@ PROCESS_THREAD(pbdrv_ioport_lpf2_process, ev, data) {
     }
 
     while (true) {
-        PROCESS_WAIT_EVENT_UNTIL(ev == PROCESS_EVENT_TIMER && etimer_expired(&timer));
-        etimer_reset(&timer);
+        PROCESS_WAIT_EVENT();
 
-        for (int i = 0; i < PBDRV_CONFIG_IOPORT_LPF2_NUM_PORTS; i++) {
-            ioport_dev_t *ioport = &ioport_devs[i];
+        // If pbio_uartdev_process tells us the uart device was removed, reset
+        // ioport id so the next timer event will take care of resetting the ioport
+        if (ev == PROCESS_EVENT_SERVICE_REMOVED) {
+            for (int i = 0; i < PBDRV_CONFIG_IOPORT_LPF2_NUM_PORTS; i++) {
+                ioport_dev_t *ioport = &ioport_devs[i];
 
-            if (ioport->connected_type_id != PBIO_IODEV_TYPE_ID_LPF2_UNKNOWN_UART) {
-                poll_dcm(ioport);
+                if (ioport->iodev == (pbio_iodev_t *)data) {
+                    ioport->connected_type_id = PBIO_IODEV_TYPE_ID_NONE;
+                }
             }
+        } else if (ev == PROCESS_EVENT_TIMER && etimer_expired(&timer)) {
+            etimer_reset(&timer);
 
-            if (ioport->connected_type_id != ioport->prev_type_id) {
-                ioport->prev_type_id = ioport->connected_type_id;
-                if (ioport->connected_type_id == PBIO_IODEV_TYPE_ID_LPF2_UNKNOWN_UART) {
-                    ioport_enable_uart(ioport);
-                    pbio_uartdev_get(i, &ioport->iodev);
-                } else if (ioport->connected_type_id == PBIO_IODEV_TYPE_ID_NONE) {
-                    ioport->iodev = NULL;
-                } else {
-                    assert(ioport->connected_type_id < PBIO_IODEV_TYPE_ID_LPF2_UNKNOWN_UART);
-                    pbio_iodev_t *iodev = &basic_devs[i];
-                    const pbio_iodev_info_t *info = &basic_infos[ioport->connected_type_id].info;
-                    iodev->info = info;
+            for (int i = 0; i < PBDRV_CONFIG_IOPORT_LPF2_NUM_PORTS; i++) {
+                ioport_dev_t *ioport = &ioport_devs[i];
 
-                    const lump_mode_flags_t *flags = &info->mode_info[0].flags;
-                    iodev->motor_flags = PBIO_IODEV_MOTOR_FLAG_NONE;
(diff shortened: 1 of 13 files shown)


lib/pbio/drv/uart/uart_stm32f0.c
```diff
@@ -178,6 +178,19 @@ pbio_error_t pbdrv_uart_set_baud_rate(pbdrv_uart_dev_t *uart_dev, uint32_t baud)
     return PBIO_SUCCESS;
 }
 
+void pbdrv_uart_flush(pbdrv_uart_dev_t *uart_dev) {
```
dlech, line 181: Since this also cancels read and write, I would be inclined to call it "reset" rather than "flush". When talking about input flush usually means "discard the buffered data" but for output it usually means "make sure it gets sent right now".

Or maybe the canceling doesn't really belong here?

lib/pbio/platform/debug/pbdrvconfig.h
```diff
@@ -48,11 +48,18 @@
 #define PBDRV_CONFIG_UART_STM32F4_LL_IRQ_NUM_UART   (1)
 
 #define PBDRV_CONFIG_HAS_PORT_1 (1)
+#define PBDRV_CONFIG_HAS_PORT_A (1)
```
dlech, line 51: Is port A replacing port 1? If so, we should remove all of the port 1 stuff.

lib/pbio/drv/ioport/ioport_lpf2.c
```diff
@@ -481,7 +481,20 @@ PROCESS_THREAD(pbdrv_ioport_lpf2_process, ev, data) {
     }
 
     while (true) {
-        PROCESS_WAIT_EVENT_UNTIL(ev == PROCESS_EVENT_TIMER && etimer_expired(&timer));
+        PROCESS_WAIT_EVENT_UNTIL(ev == PROCESS_EVENT_POLL || (ev == PROCESS_EVENT_TIMER && etimer_expired(&timer)));
+
+        // If pbio_uartdev_process process polls us, we need to run dcm again,
+        // so set the iport state to disconnected.
+        if (ev == PROCESS_EVENT_POLL) {
```
dlech, line 488: `PROCESS_EVENT_POLL` has special meaning, so this isn't the right event for an event that gets queued and takes a data argument. How about `PROCESS_EVENT_SERVICE_REMOVED`?

lib/pbio/drv/ioport/ioport_lpf2.c
```diff
@@ -481,7 +481,20 @@ PROCESS_THREAD(pbdrv_ioport_lpf2_process, ev, data) {
     }
 
     while (true) {
-        PROCESS_WAIT_EVENT_UNTIL(ev == PROCESS_EVENT_TIMER && etimer_expired(&timer));
+        PROCESS_WAIT_EVENT_UNTIL(ev == PROCESS_EVENT_POLL || (ev == PROCESS_EVENT_TIMER && etimer_expired(&timer)));
```
dlech, line 484: This can be simplified to `PROCESS_WAIT_EVENT()`

Then wrap the timer handler section in `if (ev == PROCESS_EVENT_TIMER && etimer_expired(&timer)) { }`

lib/pbio/platform/debug/platform.c
```diff
@@ -248,16 +252,6 @@ void USART2_IRQHandler(void) {
     pbdrv_uart_stm32f4_ll_irq_handle_irq(UART_ID_0);
 }
 
-// HACK: we don't have a generic ioport interface yet so defining this function
-// in platform.c
-pbio_error_t pbdrv_ioport_get_iodev(pbio_port_t port, pbio_iodev_t **iodev) {
```
dlech, line 253: Is this no longer used?

laurensvalk: ioport_lpf2.c defines it too, so there's duplicate compilation error if we kept this one. The note make it suggest it was going to be removed at that point.

lib/pbio/drv/uart/uart_stm32f0.c
laurensvalk: Ack, will rename to reset. I wasn't sure about the read/write cancel -- that was one reason for requesting the review.

lib/pbio/platform/debug/pbdrvconfig.h
laurensvalk: I'm not actually sure wat the debug port should or shouldn't be like, or whether we still use it. This was the minimal change to get the CI to pass it.

lib/pbio/drv/ioport/ioport_lpf2.c
laurensvalk: Thanks, will try that.

lib/pbio/drv/ioport/ioport_lpf2.c
laurensvalk: :+1:

lib/pbio/drv/uart/uart_stm32f0.c
dlech: If it works without the cancel in here, then I think I would prefer to leave them out and leave this as a flush function.

lib/pbio/platform/debug/pbdrvconfig.h
dlech: This was set up to use a UART on the dev board so that we could actually connect sensors to the dev board and debug them.

lib/pbio/platform/debug/platform.c
dlech: I think this is because the debug board was just using uartdev directly without the ioport. If we can, it would be nice to leave ioport disabled in the config and leave this here.

laurensvalk: The reason to enable `PBDRV_CONFIG_IOPORT` is so that `pbdrv_ioport_lpf2_process` is available from `uartdev`, but we could just wrap the whole service/poll part inside a `PBDRV_CONFIG_IOPORT` `#if` block instead.

dlech: Another way to work around this would be to wrap `process_post(&pbdrv_ioport_lpf2_process, PROCESS_EVENT_xxx, data);` in a function in the `pbdrv_ioport_lpf2_` namespace and use the usual static inline patter if PBDRV_CONFIG_IOPORT is disabled.

dlech: I've been trying to get rid of `processes.h` anyway since it leads to issues like this.

lib/pbio/drv/uart/uart_stm32f0.c
laurensvalk: It seems we can still get UART/UART to work right without them, which is good. I'm not sure about the passive case, but that should be unrelated, so I'll probably address that as part of the other comments.

lib/pbio/platform/debug/pbdrvconfig.h
laurensvalk: I revised it to not touch debug/tests at all, so this is resolved.

lib/pbio/platform/debug/platform.c
laurensvalk: Resolved by moving remaining key point to the issue.

lib/pbio/drv/ioport/ioport_lpf2.c
```diff
@@ -481,12 +481,26 @@ PROCESS_THREAD(pbdrv_ioport_lpf2_process, ev, data) {
...
 
+            // If pbio_uartdev_process tells us the uart device was removed, reset
+            // ioport id so the next timer event will take care of resetting the ioport
+            if (ev == PROCESS_EVENT_SERVICE_REMOVED) {
+                if (ioport->iodev == (pbio_iodev_t *)data) {
+                    ioport->connected_type_id = PBIO_IODEV_TYPE_ID_NONE;
+                }
+                continue;
+            }
+
```
dlech, line 503: `poll_dcm(ioport)` must be called at 2 ms intervals to work correctly. So we need to add something to ensure that it is only called when the timer event was received.

laurensvalk: That's what this is doing, right?

dlech: It looks like the timer is reset is guarded by `if (ev == PROCESS_EVENT_TIMER && etimer_expired(&timer))` but the call to `poll_dcm(ioport)` is outside of that, so it will be called when any event other than `PROCESS_EVENT_SERVICE_REMOVED` happens. There are other events that broadcast to all processes.

PR #34, opened 2020-12-11
pbio/platform/prime_hub: change dual boot button

laurensvalk:
Previously, we required Center + Right buttons to be pressed to boot Pybricks.

This simplifies it to use *only* the center button:
Short press = stock firmware; long press = Pybricks.

This makes it easier to operate the buttons with one hand, which is especially convenient if the hub is built into a large robot. Also, most hubs (Technic / City) have only one button, so this theoretically lets us keep it consistent across hubs, even though we probably won't ever have enough free space on those hubs.

### DIFF
--- a/lib/pbio/platform/prime_hub/dual_boot.c
+++ b/lib/pbio/platform/prime_hub/dual_boot.c
@@ -44,9 +44,9 @@ static void jump_to_other_firmware() {
 }
 
 /**
- * Checks if the right button is pressed during early boot.
+ * Checks if the center button is pressed for a long time (~2 s) or just briefly (<< 2 s).
  */
-static bool check_for_right_button_pressed() {
+static bool check_for_long_press() {
     __HAL_RCC_ADC1_CLK_ENABLE();
     __HAL_RCC_GPIOA_CLK_ENABLE();
     __HAL_RCC_GPIOB_CLK_ENABLE();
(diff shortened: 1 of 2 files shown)


laurensvalk:
This is a work in progress since it is currently not working as intended. 

This works (currently implemented as such):
Short press = Pybricks; long press = stock firmware.

But this does not (but this is the desired behavior):
Short press = stock firmware; long press = Pybricks.

It probably does not work because releasing the button means releasing power which is not yet on at that point.

dlech:
We probably need to remove `HAL_DeInit();` from `jump_to_other_firmware()` and `__HAL_RCC_GPIOx_CLK_DISABLE()` from `check_for_long_press()` so that the GPIOs are not reset.

(in addition to moving the following in `platform.c` before `pbio_platform_dual_boot()`

```C
    // Keep main power on (PA13 == POWER_EN)
    GPIOA->BSRR = GPIO_BSRR_BS_13;
    GPIOA->MODER = (GPIOA->MODER & ~GPIO_MODER_MODER13_Msk) | (1 << GPIO_MODER_MODER13_Pos);
```

dlech:
Merged in https://github.com/pybricks/pybricks-micropython/commit/524f06cccfd3ed5ea65fadcb0fd910936398ad70

PR #57, opened 2021-06-21
Reduce flash and RAM usage by uartdev.

laurensvalk:
This is a first step towards simplifying `pbio/uartdev` to conserve flash and RAM.

This series of commits saves 1Kb of flash and 1732 bytes of .bss, which seems quite substantial.

I have yet to run the sensor tests, but basic REPL tests look fine.

Most of it is just omitting information we don't use. Is there anything I shouldn't have dropped here?

One change is https://github.com/pybricks/pybricks-micropython/commit/5d847aa995e13f218d57b95a6ef052769dc96037, which I am still contemplating. I was experimenting if we could work towards an absolute minimum, dropping `pbio_iodev_mode_t` entirely to leave only a simple array of data types. (This is not done yet.)


laurensvalk:
All of the dropped information are constants that we could just document somewhere, so technically there is no loss of generality.

We can expand `iodevices.PUPDevice` with a method to read the entire sync data sequence. We can disable `iodevices` on Move Hub so this won't add any extra size. (In fact, save another 560 bytes).

laurensvalk:
>> We can expand `iodevices.PUPDevice` with a method to read the entire sync data sequence.
>
> I'm not sure I see any usefulness in doing this if we have everything about the device capabilities and modes documented.

Whoever documents it first has to get the information first, ourselves included :wink: This would be a way to get that info.

And it would just be the raw bytes, not any of the decoding stuff. That can be done offline --- or in an end-user script.

dlech:
> This would be a way to get that info.

FYI, I already have a logic analyzer script that decodes all of the messages. 



> And it would just be the raw bytes

So another way to do this would be to just make the generic UART device class and read the incoming data instead of adding a special method to the PUPDevice class.

laurensvalk:
> FYI, I already have a logic analyzer script that decodes all of the messages.

Yeah, it would be nice if users who have a new sensor (e.g. third party, or to their own) could do this without special hardware or waiting for us to do it, so doing it through `UARTDevice`  is nice indeed. Either way, that means we can remove everything we don't need from `uartdev`, so it sounds like we're on the same page for this PR.

dlech:
I don't have any objections to pulling all of the mode info out.

laurensvalk:
> I have yet to run the sensor tests

These are passing, so I'm going to go ahead and merge so we get the space savings for Move Hub.

As mentioned, it has mainly been about removing unused features, with only minor other changes.

It may still be beneficial to refactor `uartdev`, `ioport_lpf2` and `iodev` more thoroughly in the future to make it more robust and efficient.

PR #60, opened 2021-07-28
Enabling additional MicroPython features

laurensvalk:
Opening this for discussion.

Documenting (and finalizing some of the) MicroPython functionality and modules is part of the [v3.1](https://github.com/pybricks/support/issues/407) roadmap. It would be nice if we can do anticipated features while we are at it.

We don't need to enable all of these, but it's worth having a look. On Prime Hub, these are the size increases:

base: 228292

- bytearray: +3232
- set: +2396
- ustruct: +1388

total: +7K

`bytearray` and `ustruct` would seem useful for hub-to-hub communication which will rely heavily on byte streams.

Set is probably less critical.

This intends to address https://github.com/pybricks/support/issues/140


dlech:
:shipit: 

Would we change `buttons.pressed()` to return a `set` on platforms that support it?

laurensvalk:
Perhaps unsurprisingly, but archiving for future reference, enabling `array` is only 276 more bytes if bytearray is enabled. I personally don't really use it much.

laurensvalk:
> Would we change buttons.pressed() to return a set on platforms that support it?

Perhaps not now, so let's leave out set as it may not be worth the `2396` bytes. We can reconsider it later when we know how much space Bluetooth will require.

PR #64, opened 2021-08-08
drv/bluetooth: updates to connect generic LWP3 device instead of only the remote

NStrijbosch:
Changes required to connect to generic LWP3 devices.
A hack is included to connect to the MARIO hub as discussed here: https://github.com/pybricks/support/discussions/416

### DIFF
--- a/npm/firmware/CHANGELOG.md
+++ b/npm/firmware/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.12.0 - 2021-06-108
+### Changed
+- Updated firmware to v3.0.0.
+
 ## 4.11.0 - 2021-05-11
 ### Changed
 - Updated firmware to v3.0.0c1.
--- a/npm/firmware/package.json
+++ b/npm/firmware/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@pybricks/firmware",
-    "version": "4.11.0+firmware.v3.0.0c1",
+    "version": "4.12.0+firmware.v3.0.0",
     "description": "Pybricks firmware",
     "main": "./build/index.js",
     "types": "./build/index.d.ts",


lib/pbio/drv/bluetooth/bluetooth_stm32_bluenrg.c
```diff
@@ -472,12 +470,18 @@ static PT_THREAD(scan_and_connect_task(struct pt *pt, pbio_task_t *task)) {
...
 
-    PT_WAIT_UNTIL(pt, {
-        if (task->cancel) {
-            goto cancel_disconnect;
-        }
-        remote_lwp3_char_handle;
-    });
+    // HACK: Chararacteristics of LEGO Mario are not properly found by GATT_DiscCharsByUUID(). 
+    // remote_lwp3_char_handle for mario is hard coded for now 
+    if(context->hub_kind == LWP3_HUB_KIND_MARIO){
```
dlech, line 475: We use an automatic code format tool to try to keep things somewhat consistent. See https://github.com/pybricks/pybricks-micropython/pull/64/checks?check_run_id=3281639134

There should be spaces before and after the `()` here. 

There are also trailing spaces at the end of the line in a few other places.

lib/pbio/include/pbdrv/bluetooth.h
```diff
@@ -151,6 +153,7 @@ typedef struct {
     uint8_t status;
     uint8_t bdaddr[6];
     char name[20];
+    lwp3_hub_kind_t hub_kind;
```
dlech, line 156: Moving this to be the first item in the struct should make the compiled code size a bit smaller.

pybricks/pupdevices/pb_type_pupdevices_remote.c
```diff
@@ -89,12 +89,12 @@ STATIC void remote_connect(const char *name, mp_int_t timeout) {
     // we are using static memory
     memset(remote, 0, sizeof(*remote));
 
+    remote->context.hub_kind = LWP3_HUB_KIND_HANDSET;
```
dlech, line 92: Not a big deal, but it looks like this line got moved in the second commit for some reason.

dlech (changes requested):
Thanks for working on this! Overall, look great.

Have you tested connecting to Mario from SPIKE Prime or MINDSTORMS RI? I guess it doesn't need the hack since it probably enumerates the characteristics differently than the other hubs.

See inline comments for some small things that need to be fixed before we merge.

lib/pbio/drv/bluetooth/bluetooth_stm32_bluenrg.c
```diff
@@ -472,12 +470,18 @@ static PT_THREAD(scan_and_connect_task(struct pt *pt, pbio_task_t *task)) {
     PT_WAIT_UNTIL(pt, hci_command_status);
     context->status = aci_gatt_disc_charac_by_uuid_end();
 
-    PT_WAIT_UNTIL(pt, {
-        if (task->cancel) {
-            goto cancel_disconnect;
-        }
-        remote_lwp3_char_handle;
-    });
+    // HACK: Chararacteristics of LEGO Mario are not properly found by GATT_DiscCharsByUUID().
```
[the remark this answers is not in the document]
NStrijbosch: > Have you tested connecting to Mario from SPIKE Prime or MINDSTORMS RI?

Yes! Tested and indeed it didn't need the hack. I tested the updates for all three drivers on: technichhub; movehub and inventorhub. All three can connect to at least the remote/cityhub/mario running standard firmware.

lib/pbio/drv/bluetooth/bluetooth_stm32_bluenrg.c
NStrijbosch: > See inline comments for some small things that need to be fixed before we merge.

Thanks for the feedback! I will apply the required updates.

dlech:
Thanks!

I apparently I squashed the fixups into the other commits. Then I botched something when trying to push it back to this pull request, which automatically closed it. But the changes are merged in 2bcf69e8df5cae1b1980c31d48e32797d09510a4, 450ab1516ab1c476f31078602bce26eb3af8279d and 1235b35f148900db131bf946e54a383d2fca6460

PR #67, opened 2021-08-20
Add feature: LWP3 Device as discussed in [#262](https://github.com/pybricks/support/issues/262)

NStrijbosch:
Given that bluetooth communication with other devices than the remote was pushed forward on your timeline, I gave this a shot. 

These commits will add a starting point to the LWP3Device class as mentioned here: [https://github.com/pybricks/support/issues/262#issuecomment-800903523](https://github.com/pybricks/support/issues/262#issuecomment-800903523)

This is tested on: Technic Hub/City Hub/Prime Hub, each connects/reads/writes successfully in combination with a City hub and Remote running the stock LEGO firmware. Mario was previously working but can currently not connect see: [https://github.com/pybricks/support/issues/442](https://github.com/pybricks/support/issues/442).

I am curious what your view is on this new feature, any feedback is welcome!

I am happy to provide some code examples and/or docs (I didn't know where to place them in this repo)

### DIFF
--- a/pybricks/pupdevices/pb_type_pupdevices_remote.c
+++ b/pybricks/pupdevices/pb_type_pupdevices_remote.c
@@ -77,6 +77,45 @@ STATIC void handle_notification(pbdrv_bluetooth_connection_t connection, const u
     }
 }
 
+STATIC void remote_assert_connected(void) {
+    if (!pbdrv_bluetooth_is_connected(PBDRV_BLUETOOTH_CONNECTION_PERIPHERAL_LWP3)) {
+        mp_raise_OSError(MP_ENODEV);
+    }
+}
+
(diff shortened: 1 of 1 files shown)


laurensvalk:
Thanks for this contribution!

> I am happy to provide some code examples and/or docs (I didn't know where to place them in this repo)

Sharing it in a message here would be a good start, to facilitate discussion.

dlech:
I think Laurens made the same mistake I did on #64. @NStrijbosch can you `git push -f` your `master` branch back to GitHub? We might avoid this mistake in the future if you use `git checkout -b <name>` to give your feature branch a unique name.

laurensvalk:
I broke this pull request in an attempt to fix it. It seems like this can happen when trying to push back to a fork's master branch. So it seems to help to make a pull request from a branch by a different name, like `Mario`.

PR #74, opened 2021-10-04
pbio/drv/bluetooth_stm32_cc2640: scan timout inf

NStrijbosch:
Set the scan timeout to infinite. To connect with bluetooth devices after 10 seconds, see issue [https://github.com/pybricks/support/issues/485](https://github.com/pybricks/support/issues/485)
Additionally the scanning intervals and scanning windows are set to 30ms, to match the scan interval on the prime/essential hubs.

### DIFF
--- a/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c
+++ b/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c
@@ -1280,6 +1280,20 @@ static PT_THREAD(gap_init(struct pt *pt)) {
     PT_WAIT_UNTIL(pt, hci_command_status);
     // ignoring response data
 
+    // Set scan timeout on btchip to infinite. Seperate timout
+    // is implemented to stop program and thus scan
+    PT_WAIT_WHILE(pt, write_xfer_size);
+    GAP_SetParamValue(TGAP_GEN_DISC_SCAN, 0);
+    PT_WAIT_UNTIL(pt, hci_command_status);
+    // ignoring response data
+
+    // Set scan timeout on btchip to infinite. Seperate timout
+    // is implemented to stop program and thus scan
+    PT_WAIT_WHILE(pt, write_xfer_size);
+    GAP_SetParamValue(TGAP_LIM_DISC_SCAN, 0);
+    PT_WAIT_UNTIL(pt, hci_command_status);
+    // ignoring response data
+
     PT_WAIT_WHILE(pt, write_xfer_size);
     GAP_SetParamValue(TGAP_GEN_DISC_ADV_INT_MIN, 40);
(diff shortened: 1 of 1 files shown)


NStrijbosch:
Although this is a fix for the issue [https://github.com/pybricks/support/issues/485] (scanning indefinitely instead of 10 seconds) it exposes a new issue that still prevents connecting the remote more than 10-20 seconds after calling `Remote(timeout=None)`. 

This might be worth to be a new issue, but I wanted to summarize it here for you before considering merging this pull request. 

It seems like the hub gets in a bad state when waiting 10-20 seconds before trying to connect the remote. 

Some of my observations:
* the fix in this pull request succesfully prevents the gap event `GAP_DEVICE_DISCOVERY_DONE` being called. Hence I suspect the bt chip is still scanning
* when I use the advertisement data in a different context (just scanning for data and not connecting) shows that the btchip itself can scan for at least 45 minutes without getting into this blocking state.

I unfortunately haven't been able to identify the exact location in the code where the connection process is blocked.

lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c
```diff
@@ -1280,6 +1280,20 @@ static PT_THREAD(gap_init(struct pt *pt)) {
     PT_WAIT_UNTIL(pt, hci_command_status);
     // ignoring response data
 
+    // Set scan timeout on btchip to infinite. Seperate timout
```
dlech, line 1283:
```suggestion
    // Set scan timeout on btchip to infinite. Seperate timeout
```

lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c
```diff
@@ -1280,6 +1280,20 @@ static PT_THREAD(gap_init(struct pt *pt)) {
...
+    // is implemented to stop program and thus scan
+    PT_WAIT_WHILE(pt, write_xfer_size);
+    GAP_SetParamValue(TGAP_GEN_DISC_SCAN, 0);
+    PT_WAIT_UNTIL(pt, hci_command_status);
+    // ignoring response data
+
+    // Set scan timeout on btchip to infinite. Seperate timout
```
dlech, line 1290:
```suggestion
    // Set scan timeout on btchip to infinite. Seperate timeout
```

NStrijbosch:
> Hi Nard. Have you had any more insight on this? It sounds like it is probably OK to still merge this as-is.

I do not have new insights yet. It is indeed OK to merge this, as it fixes at least the scan time-out and for normal use the remote connects just fine. I just wanted to mention the other issue, since it is preventing an easy test to see if this fix works.

dlech:
Merged, thanks!

PR #75, opened 2021-10-08
Save programs on Prime Hub as they get downloaded

laurensvalk:
The Prime Hub, Inventor Hub, and Essential Hub have a separate chip to store user programs and other files. This is sometimes referred to as "external" storage, as opposed to the "internal" storage on the STM32 device, on which the firmware is stored.

This PR is a first attempt at:
- Accessing the external storage via SPI. Access is blocking. It calls the pbio update loop in between reading/writing chunks in order to keep everything else (more or less) running.
- Using `littlefs` as the file system layer. This uses (the somewhat outdated) V1.7 in order to be compatible with the file system used by the stock firmware of the Prime Hub and Inventor Hub. If we can manage to keep using this, we'd be able to avoid formatting the external storage. This would be nice even if we'd discontinue dual boot support.
- Saving `main.mpy` on external storage when it is downloaded to the hub via Pybricks Code or Pybricksdev. The program can be restarted by pressing the button. This makes the user experience a whole lot nicer, and it is a key feature to making Pybricks useable in First LEGO League. It's part of the [#V3.2](https://github.com/pybricks/support/issues/406) roadmap.

There are currently also a few experimental MicroPython functions to read and write files. I've used them for debugging, but I'd like to remove them before merging this. Since file access is blocking, this isn't something we should support in user programs. As long as Bluetooth keeps working well enough, we can probably get away with this for download-and-run since no other critical services like motors are running at that time.

The build size does increase, particularly due to `littlefs`, getting us quite close to our self-imposed dual boot size limit. So we might want to rethink the dual boot approach at some point. But since we have access to external flash now, there are a few other creative ways we can back up and restore the original firmware, all without DFU. We could also have a closer look at where the boundary really should be; perhaps we're making it harder than we need to.

It is currently not activated on the Essential Hub. It should work with very little modifications, but I just didn't get around to working on this yet. We'll want to check that it uses the same `littlefs` version and/or file system properties.

Before we can even think about merging this, we'll need to carefully review that we aren't doing any unwanted writes, in order to preserve the integrity of user files and the bootloader partition.


bricks/stm32/main.c
```diff
@@ -328,6 +359,12 @@ static void run_user_program(uint32_t len, uint8_t *buf, uint32_t free_len) {
 }
 
 static void stm32_main(void) {
+
+    #if PYBRICKS_HUB_PRIMEHUB
+    mp_hal_delay_ms(500);
```
dlech, line 364: what is this delay for?

pybricks/util_pb/pb_flash.c
```diff
@@ -0,0 +1,450 @@
...
+
+#include <pybricks/util_pb/pb_flash.h>
+
+SPI_HandleTypeDef hspi2;
+
+// Whether the SPI and filesystem have been initialized once after boot.
+static bool flash_initialized = false;
+
+// True sets the notCS pin, False resets it
+static void flash_enable(bool enable) {
```
dlech, line 30: It there a reason we do this manually instead of setting the pin to AF5 and letting the SPI hardware take care of it automatically?

pybricks/util_pb/pb_flash.c
```diff
@@ -0,0 +1,450 @@
...
+
+    return PBIO_SUCCESS;
+}
+
+static HAL_StatusTypeDef flash_status_read(uint8_t *status) {
+
+    uint32_t timeout = 100;
+
+    // Read status command
+    uint8_t command = 0x05;
```
dlech, line 92: are there macros/enums that we can use for these sorts of values?

pybricks/util_pb/pb_flash.c
```diff
@@ -0,0 +1,450 @@
...
+    address_bytes[1] = (address & 0xff000000) >> 24;
+    address_bytes[2] = (address & 0x00ff0000) >> 16;
+    address_bytes[3] = (address & 0x0000ff00) >> 8;
+    address_bytes[4] = address & 0x000000ff;
+
+    // Write the read command with address
+    flash_enable(true);
+    HAL_StatusTypeDef err = HAL_SPI_Transmit(&hspi2, address_bytes, sizeof(address_bytes), 100);
+    if (err != HAL_OK) {
+        return err;
```
dlech, line 148: this returns without disabling the gpio

pybricks/util_pb/pb_flash.c
```diff
@@ -0,0 +1,450 @@
...
+    // First 1MiB is reserved for firmware updates with official firmware.
+    address += 1024 * 1024;
+
+    uint8_t address_bytes[5];
+
+    // Read command
+    address_bytes[0] = 0x13;
+
+    // Address as big endian
+    address_bytes[1] = (address & 0xff000000) >> 24;
```
dlech, line 139: Applying bit masks is not necessary here (increases firmware size). Assigning to a byte takes care if it already.

pybricks/util_pb/pb_flash.c
```diff
@@ -0,0 +1,450 @@
...
+    flash_enable(true);
+    err = HAL_SPI_Transmit(&hspi2, address_bytes, sizeof(address_bytes), 100);
+    flash_enable(false);
+
+    err = flash_wait_ready();
+
+    return err;
+}
+
+int block_device_read(const struct lfs_config *c, lfs_block_t block, lfs_off_t off, void *buffer, lfs_size_t size) {
```
dlech, line 230: static?

pybricks/util_pb/pb_flash.c
```diff
@@ -0,0 +1,450 @@
...
+uint8_t lfs_lookahead_buf[256];
+uint8_t lfs_file_buf[256];
+
+struct lfs_config cfg = {
+    .read = block_device_read,
+    .prog = block_device_prog,
+    .erase = block_device_erase,
+    .sync = block_device_sync,
+
+    .read_size = 256,
```
dlech, line 313: Does this info come from the data sheet? If so, we should have a comment explaining which chip we have.

pybricks/util_pb/pb_flash.c
dlech: also applies to most functions and global variables below

pybricks/util_pb/pb_flash.c
```diff
@@ -0,0 +1,450 @@
...
+
+lfs_t lfs;
+lfs_file_t file;
+
+uint8_t lfs_read_buf[256];
+uint8_t lfs_prog_buf[256];
+uint8_t lfs_lookahead_buf[256];
+uint8_t lfs_file_buf[256];
+
+struct lfs_config cfg = {
```
dlech, line 307: can this be static and/or const?

pybricks/util_pb/pb_flash.c
```diff
@@ -0,0 +1,450 @@
...
+    }
+
+    // write file contents
+    if (lfs_file_write(&lfs, &file, buf, size) != size) {
+        return PBIO_ERROR_FAILED;
+    }
+
+    // Close the file
+    if (lfs_file_close(&lfs, &file) != LFS_ERR_OK) {
+        return PBIO_ERROR_FAILED;
```
dlech, line 445: Should we map LFS errors to PBIO errors to get a more specific error?

dlech (commented):
Nice work. Any reason we couldn't make this [non-blocking](https://bluscape.blog/2019/10/21/no-os-non-blocking-littlefs/) and put it in pbdrv right away instead of having this temporary implementation?

pybricks/util_pb/pb_flash.c
dlech: could be nice to make a `pbio_set_uint16_be()` function too

bricks/stm32/main.c
laurensvalk: For some reason Bluetooth wouldn't come up without it.

This initial read/write operation takes fairly long, so maybe there's still a bit too much blocking going on while everything else is initializing.

laurensvalk:
> Nice work. Any reason we couldn't make this [non-blocking](https://bluscape.blog/2019/10/21/no-os-non-blocking-littlefs/) and put it in pbdrv right away instead of having this temporary implementation?

No particular reason against it, just lack of skill at this point. Getting this to work seemed like a good learning opportunity and the possibility to try out the resulting user experience (which is positive.)

I made sure to keep everything out of `pbdrv` and with only minimal changes to `stm32/main.c` so it would be easy to remove when we get to the proper implementation later.

But now that we've figured out the essentials, maybe doing the right way isn't too bad in terms of time.

pybricks/util_pb/pb_flash.c
laurensvalk: No. I think we can read an arbitrary lengths from the device.  Since reads are blocking for now, this provides a reasonable balance between read time and giving the run loop enough cycles.

laurensvalk:
We've decided to merge this so we can get an early beta version for Prime Hub and Inventor Hub out. :rocket: 

We'll need to make this non-blocking later.

Issue #77, opened 2021-10-21
Calling motors in tight loop triples loop time


laurensvalk:
Calling motor commands in a tight loop affects loop time more than it should.

Relevant code (full script will be pushed shortly.)

```python
# Run for the duration of the logger.
watch = StopWatch()
while watch.time() < DURATION + 500:
    # Get error from the line.
    err = sensor.reflection() - threshold

    # Set steering proportional to reflection error.
    steering = err * 3
    drive_base.drive(SPEED, steering)
```

![image](https://user-images.githubusercontent.com/12326241/138281158-80368c12-afc4-4b42-995a-ca458b27f1f5.png)

After inserting `wait(10)` we get:

![image](https://user-images.githubusercontent.com/12326241/138281790-51136a44-d52c-422e-8b6c-18a8c6298d9c.png)

laurensvalk:
It can still happen with a `wait` in the loop; it probably just makes the occurence much less likely. But this means we should fix it.

![image](https://user-images.githubusercontent.com/12326241/138284688-cc49bc7f-d1fe-41d1-890a-0da7df5e15d0.png)

laurensvalk:
To reproduce with a single motor attached to port A, just run:

```
poetry run ./tests/motors/run_test.py tests/motors/loop_time.py --show
```

(Have to be on `motors` branch for now)

laurensvalk:
Yeah, that looks like it. All the peaks are gone with this hack:

![image](https://user-images.githubusercontent.com/12326241/138339087-cefb2394-6b64-4c6b-b746-18d343f18833.png)

Probably not how it should be done, but maybe this provides some insight.

```diff
diff --git a/py/gc.c b/py/gc.c
index 8284c435b..5f6a5a3a0 100644
--- a/py/gc.c
+++ b/py/gc.c
@@ -34,6 +34,8 @@
 
 #if MICROPY_ENABLE_GC
 
+extern int pbio_do_one_event(void);
+
 #if MICROPY_DEBUG_VERBOSE // print debugging info
 #define DEBUG_PRINT (1)
 #define DEBUG_printf DEBUG_printf
@@ -213,15 +215,18 @@ STATIC void gc_mark_subtree(size_t block) {
     // Start with the block passed in the argument.
     size_t sp = 0;
     for (;;) {
+        pbio_do_one_event();
         // work out number of consecutive blocks in the chain starting with this one
         size_t n_blocks = 0;
         do {
[...]

PR #80, opened 2021-11-30
BLE broadcast: communication between pybricks hubs

NStrijbosch:
This PR is a starting point to support communication between Primehubs, Inventorhubs, Essentialhubs, Technichubs and Cityhubs. 

The protocol is based on the protocol used by the `hub to hub` wordblocks of the official LEGO MINDSTORMS app. This implies communication is possible between a hub running pybricks and an inventor hub programmed using wordblock using the official LEGO app. 

**Code example**
When running this code on multiple prime/inventor hubs the screen will sync. 
``` python
from pybricks.ble import Broadcast
from pybricks.parameters import Button
from ustruct import pack, unpack

# Initialize the hub.
hub = ThisHub()

# Initialize broadcast
bc = Broadcast(signals=["signal_1"])

pressed = []
i=0
hub.display.number(0)
while True:
    pressed = hub.buttons.pressed()

    # Update screen if updated value is received
    if bc.received("signal_1")!=b'':
        i = unpack('<h', bc.received("signal_1"))[0] # unpack latest received data
        hub.display.number(i)
    
    # Update value if button is pressed
    if Button.LEFT in pressed:
        i=i-1
        hub.display.number(i)
        bc.transmit("signal_1",message=pack("<h",i)) # send updated value of i
        wait(500)
    elif Button.RIGHT in pressed:
        i=i+1
        hub.display.number(i)
        bc.transmit("signal_1",message=pack("<h",i)) # send updated value of i
        wait(500)

    wait(10)
``` 

**Known issues:**
Although this PR contains the essentials to get the communication working. There are a few minor issues:
*  To send data a hub needs to advertise data. To get full support for communication with inventor hubs running official firmware this requires advertising data with the `adv_type=ADV_SCAN_IND`. Currently, this does not work on the Technic Hub, hence it only sends with `adv_type=ADV_IND`. Communication between pybricks hubs works fine at the moment, but I am still looking for a fix to advertise with `ADV_SCAN_IND`. EDIT: BTChip reports error: `bleIncorrectMode` when sending with `ADV_SCAN_IND`
*  ~~To have bidirectional communication hubs need to both scan(receive) and advertise (transmit). I am not sure how this is exactly implemented on BT-chip on the Technic Hub/City but the implementation  in this PR is as follows:~~
    *  ~~when a broadcast is initiated the hub starts scanning~~
    *  ~~when `transmit()` is called the hub starts advertising. I think the BT-chip stops scanning temporarily.~~
    *  ~~1000 ms after calling `transmit()` the hub stops advertising. Scanning restarts automatically (no explicit call is necessary)~~
  ~~I implemented the 1000ms delay using an `etimer`, and it seems to work when starting a program from code.pybricks.com. However when I store the program in main.py the timer never expires, i.e., the hub advertises indefinitely and never restarts scanning. This prevents bidirectional communication when storing the program on the hub. I am not sure what all the process are doing in the background that cause this behaviour. Maybe you have some insights to fix this bug.~~
*  Some weird behaviour when programs are not started at the same time. This is caused by the following: an index in the advertisement data is used to compare if the received advertisement data is new or old. The current implementation uses the simplest possible comparison. Consequently, my rule of thumb is to restart the programs on each hub at all times to get reproducible behaviour. In the future an update might be useful to allow for a scenario where one hub keeps running the same program while the program on another hub is restarted in the mean time. (The official MINDSTORMS app has similar issues)
*  The current implementation on code.pybricks.com/beta.pybricks.com is (obviously) not yet suitable for multi-hub projects. Hence, when testing my advice is to (where possible considering the above issues) run the program from main.py.  

Anyhow, thanks in advance for all feedback you may have!


laurensvalk:
Nice work! This could be a great first step in bluetooth communication, and compatibility with the official MINDTORMS app is a nice touch.

It could be a nice addition/alternative to the BLE UART in https://github.com/pybricks/support/issues/262 since both approaches have different use cases.

As you point out, there's still some work to be done before we can merge it, but a PR is just the right spot for that :)

laurensvalk:
I've taken a closer look and reworked parts of it to address some of your questions and some stability issues:
https://github.com/pybricks/pybricks-micropython/tree/broadcast2

The aim was to move all the platform-agnostic logic to `pbio`, while leaving only driver-specific parts in `pbdrv`. Except for removing the callback setter, nothing has changed in `pbdrv` yet, so the actual Bluetooth performance should be the same.

The `pybricks.ble.Broadcast` class is now mostly just a wrapper around the `pbio` functionality. All it does is convert the `qstrs` to crc32 hashes on first use, and then it just does a look up when they are used again. `pbio` uses only the hashes.

With the original PR, the hubs would crash (watchdog timer) quite frequently. While I haven't ruled out all possible causes, I think part of it may be related to how data is allocated and used. When transmitting, I think the process could be trying to access data that no longer exists. When handling advertisement data, it could sometimes write data to heap even after the user program ends.

This update resolves this by allocating a number of read-only signals statically. There is one static object for transmission.

A next step could be to look at how the various processes should work, especially considering (re)starting of scanning and advertising.

laurensvalk:
Thanks for the great (offline) discussion so far @NStrijbosch. I've added my changes on top of your branch.

One of the next steps would be to convert the transmission into a task which updates and waits for the advertising data to be updated. It could (should?) also start another (cancellable) task to stop advertising after a timeout, and not wait for it at the user level.

Combined, this ensures that we 1) don't have to wait a second between transmissions and 2) can safely update the transmitted data without overriding ongoing tasks.

laurensvalk:
I've added an update with two additional methods (`send` and `receive`) to make communication easier.

You can now send and receive a tuple with up to 8 values at once, as follows:

```python
radio = Broadcast(topics=["info"])
radio.send("data", ("Hello", 1.234, "world!"))
```

The tuple may contain integers (2 bytes or 4 bytes), floats (4 bytes), or strings ( N + 1 bytes), as long as the total fits within 20 bytes.

This facilitates many use cases, including things like remote control of a vehicle, e.g:

```python
radio.send(steering, speed, light_status)
```

On the receiving end, you could do:

```python

data = radio.receive("info")
if data is not None:
    steering, speed, light_status = data
    # Apply steering, speed, etc.
```

The original methods to send raw bytes are still available, though they are renamed to `send_bytes` and `receive_bytes`. This can be used with hubs running the official MINDSTORMS firmware, or any other use case where raw bytes are required.

antonvh:
Great stuff going on here. Following with interest...

NStrijbosch:
> To have bidirectional communication hubs need to both scan(receive) and advertise (transmit). I am not sure how this is exactly implemented on BT-chip on the Technic Hub/City but the implementation in this PR is as follows:
> * when a broadcast is initiated the hub starts scanning
> * when `transmit()` is called the hub starts advertising. I think the BT-chip stops scanning temporarily.
> * 1000 ms after calling `transmit()` the hub stops advertising. Scanning restarts automatically (no explicit call is necessary)
>   I implemented the 1000ms delay using an `etimer`, and it seems to work when starting a program from code.pybricks.com. However when I store the program in main.py the timer never expires, i.e., the hub advertises indefinitely and never restarts scanning. This prevents bidirectional communication when storing the program on the hub. I am not sure what all the process are doing in the background that cause this behaviour. Maybe you have some insights to fix this bug.

This is fixed through: https://github.com/pybricks/pybricks-micropython/pull/80/commits/7fb0166613b6cedb5589309c36675c6e3addc926

laurensvalk:
We've now added a broadcast process that ensures we don't que up too much stuff if you send things in a tight loop. This makes things quite a bit smoother and more reliable.

That also means some additional work is required to start and stop this background process cleanly. This isn't done yet, so you might see some unexpected behavior *after* running your first program if you try this build.

JorgePe:
Thanks for this great work!
With to some hints from Laurens I got a way to listen to the broadcasts and event broadcast data myself with just the linux Bluez tools (bluetoothctl + btmon on my Ubuntu and hcitool + hcidump on ev3dev because Bluez stack is much older there). Not pretty, not stable, not fast (at least not on EV3) but it works and gives a good idea of what it's possible with connectionless BLE communication.
... but I still think there is a place for Nordic UART Service somewhere on this project ;)

dlech:
FYI, LEGO has fixed the advertising data bugs where the length was missing and the byte order of the company ID was wrong in the MINDSTORMS app v10.3.0. Unfortunately, they are still using ADV_SCAN_IND instead of ADV_NONCONN_IND.

lib/pbio/src/broadcast.c
```diff
@@ -152,7 +153,8 @@ void pbio_broadcast_parse_advertising_data(const uint8_t *data, uint8_t size) {
     }
 
     // We only process data with the right header
-    if (memcmp(data, &transmit_signal.header[0], 3)) {
+    if ((data[0] != size - 1) && data[1] != transmit_signal.AD_type &&
+        memcmp(&data[2], &transmit_signal.company_id, 2)) {
```
laurensvalk, line 157:
```c
memcmp(&data[2], &transmit_signal.company_id, 2)
```

Can probably be simplified to something like

```c
pbio_get_uint16_le(&data[2]) != transmit_signal.company_id
```

NStrijbosch:
I will just leave some issues I am trying to fix here:

When running a program on the technic hub from `main.py`:

1. once every about 500 `send()` commands the hub gets in a blocking state. Some observations in this blocking state:
    * Exiting the program with a press on the button does not work
    * Shutting down the hub with a long press does work
    * When attaching a motor and using the command `run_target(50,0)` shows the controller is still active when the hub reaches this blocking state.
2. when using bidirectional communication receiving does not work in some (1 out of 3) tests. 
    * using another device I can see that messages are send by the other hub (so something is going wrong on the technic hub side)
    * after stopping and restarting the scan at the start of the program does not seem to improve the success rate. 
    * this is something that goes wrong the whole program or goes well the whole program. 

Both issues are hard to debug since I am not able to find a case with a 100% chance on either of the issues above. Besides this  I only observed the issues when running from main.py, so I cannot print messages to the terminal to see where the program is stuck... 

Anyhow, I still have some ideas to test. I will update the observations above accordingly.

dlech:
> 1. the hub gets in a blocking state.

Does this make the problem go away?

```diff
diff --git a/pybricks/util_pb/pb_task.c b/pybricks/util_pb/pb_task.c
index 73b58558..84e20878 100644
--- a/pybricks/util_pb/pb_task.c
+++ b/pybricks/util_pb/pb_task.c
@@ -42,7 +42,7 @@ void pb_wait_task(pbio_task_t *task, mp_int_t timeout) {
         pbio_task_cancel(task);
 
         while (task->status == PBIO_ERROR_AGAIN) {
-            MICROPY_VM_HOOK_LOOP
+            MICROPY_EVENT_POLL_HOOK
         }
 
         nlr_jump(nlr.ret_val);
```

This isn't a proper fix because it would cause tasks to keep running after an exception and lead to memory corruption, etc But it would tell us that the problem is that a task was requested to be canceled but never properly canceled.

NStrijbosch:
> Does this make the problem go away?

The problem is not gone. But it does change the behaviour in the blocking state: With this change a single press will shutdown the hub, instead of the long press before. So at least `pb_wait_task` is related to the issue. 

One of my own test was to change `pb_wait_task(&task,-1` to `pb_wait_task(&task,1000)`  to trigger an error if the task wasn't finished within 1 second. Since I expect that the BTchip is not sending the HCI command the driver is waiting for when setting the advertising data. However this timeout error was never triggered, but maybe I interpret the `pb_wait_task` wrong.

dlech:
If the task times out, it still has to be cancelled and wait for cancellation to complete. But some tasks are not currently cancellable at all or in some regions of code, so will cause a lock up. Since the hub powers off/reboots when you make the change I suggested, that indicates that it caused a hard crash, which was expected.

NStrijbosch:
@dlech You are correct. Some of the tasks I added were not cancellable. I am trying to understand what is going on. 

I suspect that the assumption that the BTchip will perform all commands without an error is not justified anymore when stretching the limit of fast advertisement updates. 

To debug the status of the BTchip I use a `printf()`. But I also found the following piece of code, and was wondering how to read the debug messages that are made using `DBG()`? I do not see them in my terminal when starting a program using pybricksdev or do I need other configuration changes other than setting `DEBUG==2`? Or is extra hardware required?

https://github.com/pybricks/pybricks-micropython/blob/da8fd73420ffa0950be92d1ce069e36e6c75d28e/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c#L49-L59

dlech:
> I suspect that the assumption that the BTchip will perform all commands without an error is not justified anymore when stretching the limit of fast advertisement updates.

Yes. And in the case of actually connecting to other devices, we need to consider the handle of the remote device as well. Basically, we need to replace all instances of `PT_WAIT_UNTIL(pt, hci_command_status);` and `PT_WAIT_UNTIL(pt, hci_command_complete);` with something like https://github.com/pybricks/pybricks-micropython/blob/da8fd73420ffa0950be92d1ce069e36e6c75d28e/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c#L549-L576. This does more matching to ensure that the response was actually for this specific request. We can probably get away with leaving the init sequences the way they are for now since no other commands should be running at that time. But anything that can run after that needs this more careful handling.

> To debug the status of the BTchip I use a `printf()`.

This is dangerous since printing also uses Bluetooth. It will likely change the behavior of what you are trying to inspect. Also, using printf in any pbio code is dangerous because it can cause reentrany issues because it runs the contiki event loop while we are already in the contiki event loop. One trick I use sometimes is to make the print buffer very big so that printf doesn't block when called. This avoids the reentrancy problems in most cases.

> But I also found the following piece of code, and was wondering how to read the debug messages that are made using `DBG()`?

This uses a UART to print debug messages. This works around the problem of printf using Bluetooth. Although it still isn't perfect because it is blocking and can break things by changing the timing. To use it, you have to change the configuration to make the hub only have 3 ports instead of 4 and then add some code in the early init in platform.c to configure the UART on port D. If you look way back in git history, you can probably find a time where this was implemented.

NStrijbosch:
> This uses a UART to print debug messages.

This sounds like exactly what I hoped for! I look back in the GIT history to check the implementation.

NStrijbosch:
> * once every about 500 `send()` commands the hub gets in a blocking state. Some observations in this blocking state:
>   
>   * Exiting the program with a press on the button does not work
>   * Shutting down the hub with a long press does work
>   * When attaching a motor and using the command `run_target(50,0)` shows the controller is still active when the hub reaches this blocking state.

Using the UART hack on port D I was able to trace the cause of the blocking state. It seems that the `pbdrv_bluetooth_spi_process` can get stuck in the `PROCESS_WAIT_UNTIL(spi_xfer_complete);` here: 
https://github.com/pybricks/pybricks-micropython/blob/da8fd73420ffa0950be92d1ce069e36e6c75d28e/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c#L1703-L1709

Consequently, there are no new events registered anymore. When waiting, e.g., for an event that indicates that advertisement data is updated, this event is not triggered. Leading to the blocking state.

dlech:
I wonder if we are getting an SPI error. If you override the weak function `void HAL_SPI_ErrorCallback(SPI_HandleTypeDef *hspi)`, is it called?

NStrijbosch:
> I wonder if we are getting an SPI error. If you override the weak function `void HAL_SPI_ErrorCallback(SPI_HandleTypeDef *hspi)`, is it called?

I added the following to lib/pbio/platform/technic_hub/platform.c: 
```c++
void HAL_SPI_ErrorCallback(SPI_HandleTypeDef *hspi) {
    DBG("SPI error");
}
```
including the required definition of `DBG()` (I tested this function in another callback and printing over port D from this file works). 

But unfortunately this DBG print is not exectued, i.e., this function is not called... Or should it be placed somewhere else?

dlech:
Hmm... it sounds as if we could be just missing an interrupt then. I don't see how else `pbdrv_bluetooth_stm32_cc2640_spi_xfer_irq()` would not be called.

NStrijbosch:
Could it be just a single interrupt? 

The hub is scanning, so advertising data should be received by the BTchip all the time. So shouldn't this interrupt be triggered again after missing one? At the moment when it is stuck, the interrupt handler is not called anymore.

dlech:
The interrupt is for the end of a DMA transfer, so there will never be another interrupt until a new DMA transfer is started. But that doesn't happen because the loop that starts the transfers is locked waiting for the previous transfer to complete.

On the SPI bus, unlike UART, the MCU has to initiate all communications.

dlech:
Another thought. We could check the return value of `HAL_SPI_TransmitReceive_DMA()` to verify that the DMA transfer is actually started.

NStrijbosch:
> The interrupt is for the end of a DMA transfer, so there will never be another interrupt until a new DMA transfer is started. But that doesn't happen because the loop that starts the transfers is locked waiting for the previous transfer to complete.
> 
> On the SPI bus, unlike UART, the MCU has to initiate all communications.

Clear!

dlech:
There is some evidence for buggy DMA hardware implementation here: https://community.st.com/s/question/0D50X00009XkfAr/dmausart-on-stm32f407vg-tc-interrupt-sometimes-not-triggered

The person there way using UART w/ DMA but was also missing the DMA transmission complete (TC) interrupt occasionally.

But hopefully this is something simple like `xfer_size == 0`.

NStrijbosch:
> Another thought. We could check the return value of `HAL_SPI_TransmitReceive_DMA()` to verify that the DMA transfer is actually started.

There is indeed an error reported in the return value of `HAL_SPI_TransmitReceive_DMA()`. It returns 0 all the time. However when it is stuck it reports 1. So this return value can at least help us identify that we are stuck.

NStrijbosch:
> But hopefully this is something simple like `xfer_size == 0`.

This is the case when the process is stuck and consequently causes the `HAL_ERROR` if I understand the code correctly.

dlech:
Yes, that looked like the only code path that could trigger an error which is why I guessed it. :smile: 

So now the question is what is the state of things that causes it to be 0?

NStrijbosch:
I do not fully comprehend the SPI process. So I just document my findings here. 

The following piece of code determines the largest of the read and write buffer:
https://github.com/pybricks/pybricks-micropython/blob/da8fd73420ffa0950be92d1ce069e36e6c75d28e/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c#L1696-L1702
In the state where `xfer_size=0' both `read_xfer_size=0` and `real_write_xfer_size=0`. 

The function `get_npi_rx_size(&read_xfer_size)`  returns `false` since `read_buf[0] = 00` instead of `FE`, leading to `read_xfer_size=0`. 

`real_write_xfer_size=0` since `write_xfer_size=0` indicating there is nothing to write. 

To sum up the weird state: there is nothing to write or read....

NStrijbosch:
Continuing the path of `xfer_size=0` :

 https://github.com/pybricks/pybricks-micropython/blob/da8fd73420ffa0950be92d1ce069e36e6c75d28e/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c#L1664-L1669

The process continues after the `PROCESS_WAIT_UNTIL` in line 1668 if either:
*  `write_xfer_size!=0` we want to write something`
*  `spi_srdy=true` i am not sure what the exact meaning of this state is, but I am guessing it is set to `true` if there is something to be read?

In the situation where we end up with `xfer_size=0` the process continued after `spi_srdy` was set to  true. 

Following up on my expectation that `spi_srdy` is set to `true` if there is something to be read, I checked the return value of `get_npi_rx_size()`. I observe that this value is always `true` if `spi_srdy` is set to true. However, when we end up with `xfer_size=0` we have `spi_srdy=true` and `get_npi_rx_size()=false`.

DietrichChristopeit:
Dear Mr. Strijbosch,The first thought when I read the abbreviation was 'state ready' or 'send ready'...Maybe it helps.Thank you for all your efforts for pybricks, it's great.Best regards,Dietrich. 
-------- Ursprüngliche Nachricht --------Von: NStrijbosch ***@***.***> Datum: 16.01.22  11:51  (GMT+01:00) An: pybricks/pybricks-micropython ***@***.***> Cc: Subscribed ***@***.***> Betreff: Re: [pybricks/pybricks-micropython] BLE broadcast: communication between pybricks hubs (PR #80) 
Continuing the path of xfer_size=0 :
https://github.com/pybricks/pybricks-micropython/blob/da8fd73420ffa0950be92d1ce069e36e6c75d28e/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c#L1664-L1669
The process continues after the PROCESS_WAIT_UNTIL in line 1668 if either:

write_xfer_size!=0 we want to write something`
spi_srdy=true i am not sure what the exact meaning of this state is, but I am guessing it is set to true if there is something to be read?

In the situation where we end up with xfer_size=0 the process continued after spi_srdy was set to  true.
Following up on my expectation that spi_srdy is set to true if there is something to be read, I checked the return value of get_npi_rx_size(). I observe that this value is always true if spi_srdy is set to true. However, when we end up with xfer_size=0 we have spi_srdy=true and get_npi_rx_size()=false.

dlech:
> * `spi_srdy=true` i am not sure what the exact meaning of this state is, but I am guessing it is set to `true` if there is something to be read?

This reflect the state of the SRDY signal from the Bluetooth chip. It means that the Bluetooth chip has a message ready for us to read. It is connected to a GPIO interrupt handler.

> However, when we end up with `xfer_size=0` we have `spi_srdy=true` and `get_npi_rx_size()=false`.

Well, this is the point where we need to put a logic analyzer on the Bluetooth chip to see what is really going on. But that is much easier said than done. 

Making a blind guess here, I wonder if we make the 2 ms timeout hack that skips waiting for the SRDY signal to change longer if the problem goes away?

NStrijbosch:
> I wonder if we make the 2 ms timeout hack that skips waiting for the SRDY signal to change longer if the problem goes away?

Unfortunately not. Both with 4ms and 10ms I have observed the problem. When setting it 1 ms getting in a blocking state seems to have a higher likelihood (within 50 sends instead of 500), but on first sight this blocking state seems to be somewhere else in the process.

dlech:
> When setting it 1 ms getting in a blocking state seems to have a higher likelihood (within 50 sends instead of 500), but on first sight this blocking state seems to be somewhere else in the process.

As you probably say in the comments in the code, this is expected.

---

Here is another hack to try:

```diff
diff --git a/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c b/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c
index 639905ae..f508cc27 100644
--- a/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c
+++ b/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c
@@ -1701,11 +1701,16 @@ start:
             xfer_size = real_write_xfer_size;
         }
 
-        // read the remaining message
-        spi_xfer_complete = false;
-        pdata->spi_start_xfer(&write_buf[NPI_SPI_HEADER_LEN],
-            &read_buf[NPI_SPI_HEADER_LEN], xfer_size);
-        PROCESS_WAIT_UNTIL(spi_xfer_complete);
+        // HACK: sometimes there is no data to read or write (SRDY was set but
+        // data read is all 0s). Need to verify that this is really happening
+        // with logic analyser, otherwise we could be dropping packets here.
+        if (xfer_size) {
+            // read the remaining message
+            spi_xfer_complete = false;
+            pdata->spi_start_xfer(&write_buf[NPI_SPI_HEADER_LEN],
+                &read_buf[NPI_SPI_HEADER_LEN], xfer_size);
+            PROCESS_WAIT_UNTIL(spi_xfer_complete);
+        }
 
         // HACK: SRDY can transition from low and back to high in the time
         // between we set MRDY and when we read SRDY again. So we use a timer
```
---



> The function `get_npi_rx_size(&read_xfer_size)` returns `false` since `read_buf[0] = 00` instead of `FE`,

Did you look at `read_buf[1]` and `read_buf[2]` to see if they are also 0?

dlech:
Another thing to try. This should let us see if the loop is starting again without being triggered by an interrupt.

```diff
diff --git a/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c b/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c
index 639905ae..39c47fca 100644
--- a/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c
+++ b/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c
@@ -733,7 +733,14 @@ void pbdrv_bluetooth_disconnect_remote(void) {
 
 // Driver interrupt callbacks
 
+static uint32_t srdy_count, n_srdy_count;
+
 void pbdrv_bluetooth_stm32_cc2640_srdy_irq(bool srdy) {
+    if (srdy) {
+        srdy_count++;
+    } else {
+        n_srdy_count++;
+    }
     spi_srdy = srdy;
     process_poll(&pbdrv_bluetooth_spi_process);
 }
@@ -1664,11 +1671,17 @@ start:
 
     while (true) {
         static uint8_t real_write_xfer_size;
+        static uint32_t srdy_loop_count;
 
         PROCESS_WAIT_UNTIL(spi_srdy || write_xfer_size);
 
         spi_set_mrdy(true);
 
+        if (spi_srdy) {
+            srdy_loop_count++;
+            DBG("%u %u %u\n", srdy_count, n_srdy_count, srdy_loop_count);
+        }
+
         // Need to make a copy of the write transfer size for the case where we
         // are reading only. In this case we still have to set write_xfer_size
         // to a non-zero value to indicate that write_buf is being used. But we
```

NStrijbosch:
> As you probably say in the comments in the code, this is expected.

Ofcourse. This was just a to trigger this undesired behaviour and check if this matched the behaviour I observed. Which on first sight turns out to be different. 



> Here is another hack to try:

This unfortunately did not help..

> Did you look at `read_buf[1]` and `read_buf[2]` to see if they are also 0?

I just had a look. It depends on where I evaluate them: 

Evaluating them just after `PROCES_WAIT_UNTIL` at the start of the while loop, i.e., line 1669:
 https://github.com/pybricks/pybricks-micropython/blob/da8fd73420ffa0950be92d1ce069e36e6c75d28e/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c#L1664-L1669

`read_buf[0]=00`, `read_buf[1]=FE`, `read_buf[2]=09`

Evaluating them just before reading the remaining message, i.e., line 1703:
 https://github.com/pybricks/pybricks-micropython/blob/da8fd73420ffa0950be92d1ce069e36e6c75d28e/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c#L1703-L1708

`read_buf[0]=00`, `read_buf[1]=00`, `read_buf[2]=00`

> Another thing to try. This should let us see if the loop is starting again without being triggered by an interrupt.

This seems to lead to interesting results. At every print each of the values is strictly increasing, except when it blocks then the values of `srdy_count` and `n_srdy_count` are identical to the previous print. This indicates a loop being triggered by an interrupt, right?

Just to be sure: the values of `srdy_count` and `n_srdy_count` are not necessarily increased by 1 at each print. Just before the blocking state it is increased by 2 (the few times I checked). Is this just an indication of a simultaneous read write? Or could this also be something to investigate further?

dlech:
> This seems to lead to interesting results. At every print each of the values is strictly increasing, except when it blocks then the values of `srdy_count` and `n_srdy_count` are identical to the previous print. This indicates a loop being triggered without an interrupt, right?

Interesting indeed. FYI, I just pushed a new commit yesterday that gets rid of the timer hack and relies on the `n_srdy_count` to wait before continuing with the loop. We should see if this makes a difference. If it doesn't, then it seems like the only other possibility is that `write_xfer_size` is being reset to 0 when it should not be.

> Just before the blocking state it is increased by 2 (the few times I checked).

That is weird. It is like it thinks we already read did a read when we didn't. The SRDY signal should not change until we release the MRDY. There is a timing diagram [here](https://dev.ti.com/tirex/explore/content/simplelink_cc13x2_26x2_sdk_3_10_00_53/docs/ble5stack/ble_user_guide/html/ble-stack-common/npi-index.html#npi-handshake) that explains how the protocol works (note, it shows actual voltage levels on the pins, which are active low, so "high" and "low" will look inverted compared to our code).

Or maybe there is something blocking the event loop for running for too long? We recently noticed https://github.com/pybricks/pybricks-micropython/issues/77 with motors. We could log timestamps for each rising and falling edge of SRDY to see how long it is between when the signal changes and when we actually start the SPI read. I wonder if we take too long to read, if the chip gives up and turns off SRDY.

> Evaluating them just after `PROCES_WAIT_UNTIL` at the start of the while loop, i.e., line 1669:

This would be leftover from the previous read.


> Evaluating them just before reading the remaining message, i.e., line 1703:

Makes sense that these are all 0s here if `spi_srdy` is false. Just wanted to be sure there wasn't an off by one error or something.

NStrijbosch:
@dlech thank you very much for your suggestions. I wasn't yet aware of those docs, thank you for point me to those. 

I found your recent commit, it solved the `xfer_size=0` state!
However there is a catch: once in a blue moon the spi process is blocked by the `PROCESS_WAIT_UNTIL(spi_n_srdy_count != prev_n_srdy_count); ` 
https://github.com/pybricks/pybricks-micropython/blob/e9a20efbc3d786d2b4436d87f7be861f68eae7f1/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c#L1725-L1730
I have checked the srdy interrupt callback, and it is not called anymore after the `spi_set_mrdy(false);`

This blocking state is less likely then the `xfer_size=0` state that it solved. To get into this state within about the same time interval I increased the advertisement updates from 20 times per second to 50 times per second.

laurensvalk:
I've added a suggested commit to move the import to the `pybricks.experimental` module for now. Feel free to cherry-pick the relevant pieces if you like.

We won't be finalizing the BLE roadmap for the next release, so this increases the chances of merging it a bit.

Assuming this gets robust enough (e.g. the aforementioned issues are resolved) and it doesn't impact other areas of Pybricks when users don't use it, you have my vote to get it merged as an experimental feature. :+1:

winklerp:
I think this will be an awesome feature. Especially making it compatible with the standard word blocks. Thank you very much for your effort! Is support for EV3 also planned? I am heavily invested in EV3 and 3rd party sensors and would like to continue to use this hardware in combination with the newer stuff.

dlech:
> I increased the advertisement updates from 20 times per second to 50 times per second.

This seems quite fast. I have a feeling we might be overloading the Bluetooth chip and causing it to get in a bad state. We saw something similar to this in https://github.com/pybricks/support/issues/397 (it's a long issue, so start at the end).

How long does it run if we only send 10 times per second?

dlech:
> Is support for EV3 also planned?

No. EV3's built-in Bluetooth doesn't support Bluetooth Low Energy, so it is not compatible.

dlech:
> How long does it run if we only send 10 times per second?

And a related question: what rate does the official LEGO firmware use?

NStrijbosch:
> And a related question: what rate does the official LEGO firmware use?

As in advertising interval or the maximum rate you can send updates?

dlech:
> As in advertising interval or the maximum rate you can send updates?

The advertising interval.

dlech:
Here is an interesting discussion: https://e2e.ti.com/support/wireless-connectivity/bluetooth-group/bluetooth/f/bluetooth-forum/463083/cc2640-as-spi-slave-npi-tl-protocol

It suggests that SRDY may sometimes change before MRDY at the end of a transaction. If this is what is happening, I think this should help.

```diff
diff --git a/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c b/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c
index 7624ef79..c5555159 100644
--- a/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c
+++ b/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c
@@ -1699,6 +1699,14 @@ start:
             write_xfer_size = PBIO_ARRAY_SIZE(write_buf);
         }
 
+        // After the transfer is complete, we release the MRDY signal to the
+        // Bluetooth chip and it acknowledges by releasing the SRDY signal.
+        // Sometimes SRDY will change before we change MRDY? so we need to arm
+        // the trigger here.
+        // https://e2e.ti.com/support/wireless-connectivity/bluetooth-group/bluetooth/f/bluetooth-forum/463083/cc2640-as-spi-slave-npi-tl-protocol
+        static uint8_t prev_n_srdy_count;
+        prev_n_srdy_count = spi_n_srdy_count;
+
         // send the write header
         spi_xfer_complete = false;
         pdata->spi_start_xfer(write_buf, read_buf, NPI_SPI_HEADER_LEN);
@@ -1718,10 +1726,6 @@ start:
             &read_buf[NPI_SPI_HEADER_LEN], xfer_size);
         PROCESS_WAIT_UNTIL(spi_xfer_complete);
 
-        // After the transfer is complete, we release the MRDY signal to the
-        // Bluetooth chip and it acknowledges by releasing the SRDY signal.
-        static uint8_t prev_n_srdy_count;
-        prev_n_srdy_count = spi_n_srdy_count;
         spi_set_mrdy(false);
         // Multiple interrupts can happen between line above and line below,
         // so we can't use PROCESS_WAIT_UNTIL(!spi_srdy) as the SRDY signal may
```

NStrijbosch:
> How long does it run if we only send 10 times per second?

Just did two bidirectional test (with advertising interval 100ms), both crashed after around 5-6 minutes.
Two test with just updating advertising data (with advertising interval 100ms), no bad state for 20 minutes straight.  

With 50 times per second it is typically within 1 minute (bidirectional). So that makes debugging a bit more time efficient. 
But I get your point that if 10 times per second leads to never ending up in bad state that would already be great and something we could work with, unfortunately this is not the case... 

> I think this should help.
Unfortunately not... there is no interrupt after arming the trigger. Also when I arm the trigger even earlier.

I will try to play a bit further with the advertising/scan intervals. But from my observations until now it seems to only affect the likelihood and not eliminate the issue. 

Just some thoughts: When only communication is blocked what would be the behaviour of the bluetooth chip? What i observe is that advertising of data is stopped immediately. I would expect it to continue if it is only the communication that fails, i.e., the bluetooth chip does not receive commands to stop advertising and just continues. Is the immediate stop of advertising data a sign of the BTchip crashing?

dlech:
> Is the immediate stop of advertising data a sign of the BTchip crashing?

Yes, I think so. I suspect data is coming in faster than it can be offloaded back to the MCU and causes a memory allocation failure or buffer overflow or something like that.

Have you also tried running with debug printing disabled? This could be blocking too long and make the problem worse. 

Maybe there are some settings we could adjust to reduce RAM usage on the BT chip too? For example, maybe we could reduce `maxScanResponses` in `GAP_deviceInit()`?

NStrijbosch:
> Have you also tried running with debug printing disabled?

Yes, the problem persists. 

> maybe we could reduce `maxScanResponses` in `GAP_deviceInit()`?

I have tried this before, even with `maxScanResponses=1` het problem persists. 

I have spend some time experimenting with the order of the messages send to BTchip. I think I can conclude at this time is that scanning and advertising at the same time will lead to this bad state at some point. The rate of updating the advertising data might increase or decrease the likelihood. 

However, I cooked up an example of we can get close to bidirectional communication and still send 10 messages per second while preventing the BTchip crashing. 

The following example sends messages about 10 times per second and continuously scans for advertising data. The BTchip crashes on average after 5-6 minutes. 
``` python
from pybricks.ble import Broadcast


hub = ThisHub()
hub.light.on(Color.WHITE)

# Initialize broadcast with two topics.
radio = Broadcast(["counter", "hue"])

for i in range(10000):
    radio.send("counter", i%1000)
    wait(100)

    # Receive  hue signal.
    hue = radio.receive("hue")

    # If we received something, set the light.
    if hue:
        hub.light.on(Color(hue))

```

Just as a test I set the advertising time to 50 ms and ran the following example which turns scanning off while advertising for 50ms. Basically this is cheating on true bidirectional communication. However, this turned out to work flawless. 
``` python
from pybricks.ble import Broadcast


hub = ThisHub()
hub.light.on(Color.WHITE)

# Initialize broadcast with two topics.
radio = Broadcast(["counter", "hue"])

for i in range(10000):
    # turn off scanning for 50 ms and start advertising (advertising stops after 50ms)
    radio.scan(False)
    radio.send("counter", i%1000)
    wait(50)

    # advertising has stopped now restart scanning for 50 ms
    radio.scan(True)
    wait(50)

    # Receive  hue signal.
    hue = radio.receive("hue")

    # If we received something, set the light.
    if hue:
        hub.light.on(Color(hue))

```
In my home environment there is not really a difference in the number of messages that are lost. In other words the performance of the two examples is comparable with the benefit of the bottom one not leading to a crash of the BTchip. 

Just as an interesting note: the bottom example sends 5*10 = 50 messages per second to the BTchip. The top example sends only 10 messages per second. However, the number of messages with received advertising data from the BTchip probably halved compared to the top example. 

So I am quite positive about this observation and will explore the possibility to implement this behaviour in a process in the driver such that in the end the top example runs identical to the bottom one.

NStrijbosch:
I have had a look at this PR again with a fresh perspective, and I have made a good step to get bidirectional communication working reliably on the technic/city hubs! There is still some work to do when starting the program from code.pybricks.com but when running a script from `main.py` the experience is similar to the primehub now, and does not crash the bt-chip as discussed before. 

The main change is the addition of a broadcast process in the stm32_cc2640 driver. This process periodically scans (30 ms) and advertises (20 ms). This prevents the bt-chip form scanning/advertising at the same time (leading to undesired behaviour) . On the user end this allows to receive broadcast messages and send messages 'simultaneously' without having to consider the limitations of this BTchip.

dlech:
Glad to know you are still working on this. Is there still problems with the Bluetooth chips crashing when connected to Pybricks Code?

NStrijbosch:
> Is there still problems with the Bluetooth chips crashing when connected to Pybricks Code?

I need to investigate further. But I have not noticed crashes when connected to Pybricks Code. 

The main issues I have now:
-  The broadcast process is not stopped correctly, so a restart of the hub is necessary after stopping the program. If I understand it correctly the BTchip does not get restarted when it stays connected, while it does restart if you stop the program when running from `main.py`. I am working on this. 
-  The broadcast process seems to be blocked for 500ms periodically. I think this corresponds to the frequency of bt messages send to code.pybricks.com (when there is not `print` in the program). I have to do more thorough tests to get to the root cause. But at the moment this results in updates of the broadcasted messages every 500 ms instead of 50 ms.

NStrijbosch:
> * The broadcast process is not stopped correctly,

This is fixed now. It is not necessary to power cycle the technic hub anymore.

NStrijbosch:
The lag I observe is caused by the BT messages send to code.pybricks.com. To be more precise while the BT driver is executing the following thread

https://github.com/pybricks/pybricks-micropython/blob/64c5d6cea2abbf6ad1faa5c2ff3b20244a60628b/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c#L491-L544

The duration of this task is noticeable when the program aims to update the advertising data every 100ms....  
While this thread is active the following function, which updates the advertising data is not called. (Or can it not be called?)

https://github.com/pybricks/pybricks-micropython/blob/64c5d6cea2abbf6ad1faa5c2ff3b20244a60628b/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c#L405-L409

I am not sure yet if there could be a simple fix... 

One question I have which I did not really dig into: is this behaviour also observed with the PUP remote? Just as a note: I plan to test this updating the LED  color on the remote with a similar frequency. I do not recall ever observing this behaviour, so maybe this is not an issue for the remote and therefore could also be fixed for broadcasting data.

dlech:
It's been a while since I've been in the Bluetooth code, but IIRC, it is currently designed to basically do one thing at a time as a simple way to avoid overloading the Bluetooth chip. The chip should be able to do certain things in parallel, but we would need to update all of the code to check the value of status events to see if the command was actually accepted or not and keep retrying until it goes through.

I suspect the powered up remote has light messages delayed for the same reason, but it probably isn't long enough to visually noticeable.

NStrijbosch:
This is the lag I observe with the current implementation of broadcast when connected to code.pybricks.com:

https://user-images.githubusercontent.com/67324303/181808770-8cc7a039-0f28-4c3f-a070-9cb7bf5b2428.mp4

The values on the Inventor hub display are the values received from the Technic hub. 

A quick test with the remote while connected to code.pybricks.com:

https://user-images.githubusercontent.com/67324303/181808721-64c6fb23-1244-4c41-baeb-f30f448fd4f8.mp4

Clearly the updates (with the same frequency) of the remote led are less laggy. 
So I think even with the current "one thing at a time" approach for the BTchip there should be room for improvement for the broadcast messages. 

I will dig deeper into what the fundamental difference is between the remote and broadcast that causes this.

dlech:
> I will dig deeper into what the fundamental difference is between the remote and broadcast that causes this.

I think the difference is that the remote is using a connection with reliable writes and broadcasts are connectionless. The former will ensure the write completes as long as the connection is not lost while the latter depends on random timing of one hub transmitting and the other hub receiving at the same time.

NStrijbosch:
Of course that is a fundamental difference between the two. But if that is the root cause of this issue it would also be laggy when starting the same program from  `main.py`. This is not the case:

https://user-images.githubusercontent.com/67324303/181811701-fee13f80-4f22-4420-a014-7b16c1706221.mp4

I am quite happy with this responsiveness when starting from `main.py`. Hence, my quest to get similar results when starting from code.pybricks.com. From my point of view the only difference between the two is the extra bt interaction with code.pybricks.com. Clearly this interaction is done well with the remote, but in the current broadcast implementation there is room for improvement

NStrijbosch:
> The lag I observe is caused by the BT messages send to code.pybricks.com. To be more precise while the BT driver is executing the following thread

I think this statement is incorrect. I timed the thread below in the following settings:
- idle connection with code.pybricks.com
- updating the LED on the remote 10 times per second
- broadcasting with 10 message updates per seconds

In each of these situations the duration of this thread is ~2ms. While the observed lag is ~500 ms. 

In other words the bt chip being busy with the messages to code.pybricks.com is not the rootcause. I have to find the solution somewhere else...

https://github.com/pybricks/pybricks-micropython/blob/64c5d6cea2abbf6ad1faa5c2ff3b20244a60628b/lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c#L491-L544

NStrijbosch:
I think I have found something that could explain the difference in behaviour between running from `code.pybricks.com` and running directly on the hub from `main.py`:

When only receiving on the Technic hub I notice that when I run the program from `code.pybricks.com` a lot of `HCI_COMMAND_STATUS` messages with the status `ble_pending`. When running the same script from `main.py` the `HCI_COMMAND_STATUS` is always `ble_success`.  At the moment many interactions with the BTchip in the broadcast process wait for a `HCI_COMMAND_STATUS` message and not check the actual status. 

The extra communication could explain the `ble_pending`. Not handling this `ble_pending` correctly at the moment could explain the difference in behaviour. 

At the moment I am not sure yet what the exact meaning of `ble_pending` is, so I will try to figure this out in order to correctly handle this situation. I am thinking along the lines of just waiting for `ble_success` or retry the action that triggers the response `ble_pending` again.

dlech:
`blePending` means that the chip is busy doing something else so the command was ignored and needs to be sent again.  

There are several instances where we already retry when this status is received. For example:

```c

    HCI_StatusCodes_t status = read_buf[8];

    if (status != bleSUCCESS) {
        if (task->cancel) {
            goto cancel;
        }

        if (status == blePending) {
            goto retry;
        }

        goto exit;
    }
```

laurensvalk:
You can now try this out without building the firmware yourself: https://pybricks.com/projects/tutorials/wireless/hub-to-hub/broadcast/

lib/pbio/src/broadcast.c
```diff
@@ -0,0 +1,236 @@
...
+    if (transmit_signal.hash == hash) {
+        transmit_signal.index++;
+    } else {
+        // Otherwise reset the index.
+        transmit_signal.index = 0;
+        transmit_signal.hash = hash;
+    }
+
+    // Copy the payload.
+    transmit_signal.length = size - 1;
```
pbochynski, line 156: Why length does not include PBIO_BROADCAST_META_SIZE?
In my opinion, it should be added to the length. When I receive the advertising data I miss exactly 9 bytes.
```suggestion
    transmit_signal.length = size + PBIO_BROADCAST_META_SIZE - 1;
```

laurensvalk: @NStrijbosch: Feel free to comment on this. It may be a while before we can get back to this.

NStrijbosch: @pbochynski You are correct! I just pushed an update with your fix. 

In the current implementation in pybricks this value is never checked, so from pybricks hub to pybricks hub communication this value didn't really matter, so I never noticed the mistake.

NStrijbosch:
With [https://github.com/pybricks/pybricks-micropython/pull/80/commits/a9e5b410f8a6da896744397cb3135214b44f3787](https://github.com/pybricks/pybricks-micropython/pull/80/commits/a9e5b410f8a6da896744397cb3135214b44f3787) it is now possible to access the RSSI value, counter, and timestamp corresponding to the last received data for a topic, as discussed in [https://github.com/pybricks/support/issues/801](https://github.com/pybricks/support/issues/801). 

A short example program:
```python
from pybricks.experimental import Broadcast

# Initialze hub
hub = PrimeHub()

# Initialize broadcast with one topic.
radio = Broadcast(["hue"])

while True:
    index, timestamp, rssi = radio.info("hue")
    print(rssi)

    if rssi > -60:
        hub.light.on(Color.GREEN)
    else:
        hub.light.on(Color.RED)

    wait(100)
```

massimiliano-mantione:
Hi!
I have been experimenting with this feature using the build provided in the "hub to hub communication" project page ("https://pybricks.com/projects/tutorials/wireless/hub-to-hub/broadcast/").
Thanks for your work on this, it is amazing!

What I need is bidirectional communication between two technic hubs, ideally with "minimal" lag.

To test responsiveness (a bit empirically) I connected two motors to each hub to ports A and B and used the position of the motor on port A (the `angle{}` reading) to move motor B on the other hub with `track_target` (broadcasting positions bidirectionally with two topics).

I noticed that the communications works for a few seconds, then stops for a few seconds, then resumes again...

I load the programs on both hubs from `beta.pybricks.com` but then I disconnect the browsers to avoid console-related buetooth traffic.

Is there something I can do to troubleshoot this?
Anf BTW, what does "start from main.py" mean?

NStrijbosch:
> I noticed that the communications works for a few seconds, then stops for a few seconds, then resumes again...

Thank you for the feedback. I do not recognize the drop in communication that takes seconds. Can you share a small code example such that I can reproduce this?

> Anf BTW, what does "start from main.py" mean?

With this implemention it is important to distinguish between the two possible scenario's to run a program: 
- Start from beta.pybricks.com and press the start button on the screen (this way the hub stays connected to your PC)
- Start by pressing the button on the hub (while being disconnected from your PC). The program that is stored on the hub is used in that case. When I build the experimental firmware I can already store the program in a file name main.py. Hence, I refer to this situation as starting from main.py. But I think this is practically the same as starting the program once from beta.pybricks.com, disconnect from beta.pybricks.com, and then start the program by pressing the button on the hub (This is what you already are doing, which is the most optimal situation for broadcasting)

massimiliano-mantione:
Well, the two programs I use are short enough.

Here is the "commander":
```
from pybricks.hubs import TechnicHub
from pybricks.parameters import Color, Port
from pybricks.robotics import DriveBase
from pybricks.tools import wait, StopWatch
from pybricks.experimental import Broadcast

hub = TechnicHub()
ma = Motor(Port.A)
mb = Motor(Port.B)

radio = Broadcast(topics=["data", "cmd"])

data_angle = 0
cmd_angle = 0

while True:
    cmd_angle = ma.angle()
    mb.track_target(data_angle)

    radio.send("cmd", (cmd_angle))
    data = radio.receive("data")
    if data:
        data_angle = data
        hub.light.on(Color.GREEN)
    else:
        hub.light.on(Color.RED)

    wait(10)
```

And here is the "reporter":
```
from pybricks.hubs import TechnicHub
from pybricks.parameters import Color, Port
from pybricks.robotics import DriveBase
from pybricks.tools import wait, StopWatch
from pybricks.experimental import Broadcast

hub = TechnicHub()
ma = Motor(Port.A)
mb = Motor(Port.B)

radio = Broadcast(topics=["data", "cmd"])

data_angle = 0
cmd_angle = 0

while True:
    data_angle = ma.angle()
    mb.track_target(cmd_angle)

    radio.send("data", (data_angle))
    cmd = radio.receive("cmd")
    if cmd:
        cmd_angle = cmd
        hub.light.on(Color.GREEN)
    else:
        hub.light.on(Color.RED)

    wait(10)
```

They are actually mostly identical, with just the topics reversed.
I refer to them as "commander" and "reporter" because in the actual project I plan to keep all the real "logic" on one brick (the "commander") and use the other one like an "extension" to report data and execute commands with the connected motors and sensors (hence the name "reporter").

massimiliano-mantione:
BTW, I'd like to know the actual rate limit for bidirectional communication :-)
My project is a robot for a competition, and if at all possible I'd like to push the msg exchange to 100Hz (and even that would be slow for what would be really needed).

It's a bit off-topic, but I wonder if a Technic Hub could be configured as a *master* for the Nordic UART Service (NUS), and connect to the other hub using the NUS, and if this would result in better performance than this amazing but tricky broadcast machinery (which, I admit, is a stroke of genius in its way of piggybacking on the BLE device advertising data frames...).

NStrijbosch:
> Well, the two programs I use are short enough.

Thanks I will try to reproduce your issue in the coming days. 

> I'd like to know the actual rate limit for bidirectional communication

In the current implementation this is 20 Hz for the Technic/City hubs (these switch between scanning (20ms) and advertising (30ms)). For the Inventor/Prime/Essential hubs this rate might be better, these hubs are able to scan and advertise at the same time. However you have to consider that it is a lossy communication protocol, i.e., increasing the communication rate might lead to a lot of lost data.

> It's a bit off-topic, but I wonder if a Technic Hub could be configured as a _master_ for the Nordic UART Service (NUS), and connect to the other hub using the NUS,

I think this should be possible, and it could definitely give you better performance than the current broadcast. At the time this broadcast implementation seemed to be an easier way to get hub-to-hub communication working for a lot of usecases.

massimiliano-mantione:
I think I have a reasonable explanation for the behavior I am observing.

> In the current implementation this is 20 Hz for the Technic/City hubs (these switch between scanning (20ms) and advertising (30ms)).

The two hubs are not synchronizing.

What I fear is happening is that the 50ms period (20ms scan + 30ms send) is not "exact", and the two hub periods change "phase" periodically (relatively slowly).
So, there are time intervals when the two periods are phased "properly" (one sends and the other receives), and there are time intervals when they are either talking past each other or listening to each other's silence.

I think that this could be verified with a BLE radio scanner on a third device (like a PC or a radio-enabled CPU like the ESP32), and I have seen instructions on how to make it in other related issues (like [here](https://github.com/pybricks/support/issues/800)) but I still did not have the time to do it.

Now, supposing that I am right, and given the fact that technic hubs are not able to advertise and receive at the same time, to make this work in the bidirectional case we'd need to sync their periods in some way.

I already have a few ideas, but before explaining them I'd like to check a few things about hardware capabilities.
I am reading the code for this PR in full, but there are a few details I am not getting.
Specifically, I have these questions:

- when the BLE chip is in advertising mode, do we have control over how frequently it sends advertising packets over time?
- do we have control ever every sent packet, in the sense that we can change its contents? (is seems we do, because the 1 byte "counter" is incremented at each retransmission of the same packet, but maybe I am misunderstanding
- could we attach a microsecond-resolution timestamp to every received packet?
- do we have access to the sender BLE device ID for every received packet?

massimiliano-mantione:
My general idea is to give more information to the broadcast system initialization about the "network topology".
Every brick participating in the network must already "share" the topic names with the other bricks.
In every brick, the list of topics should be identical (and *ordered*), and each topic should have the following info attached:

- whether this brick is going to publish on it or receive from it
- a transmission time period (this could have a sensible default, again, shared between brick builds)

Each brick is supposed to follow the topic list cyclically, and for each topic either receive or transmit for the transmission time period of that topic.

Each packet should contain a number that tells how many microseconds the transmitter is inside its transmission time window.
Each brick, during each "reception" time window, should use the timestamps in the received packets to "sync its period" with the transmitter.

Note two more things:
- also hardware that can receive and transmit at the same time should participate in this dance, otherwise it risks transmitting when others are not listening (and, maybe, produce ratio traffic that can lead to lost packets)
- in the simple unidirectional case, with a set of topics written by a single brick and received by all others, the transmitter always sends and the readers always receive, and the latency would be the minimal BLE physical latency

I know this complicates things, but I fear that it is the only way to make it work with the technic hub BLE chip.

The technic hub is dear to me because it is likely the best hardware we can have for Lego robotic competitions.
Mindstorms have been discontinued.
Spike Prime Hubs are expensive, and they have built-in 2 cells LiPo batteries that severely limit the performance of the attached motors.
Technic hubs are amazing but they have only four ports, no display, and a single button.
Two technic hubs, properly paired, and talking to an external device for initial robot configuration, would be a perfect replacement for a single EV3 brick :-) (8 ports total)

I care about latency because our robots move at about 1m/s (line followers are even faster).
At those speeds waiting 50ms means reacting 50mm later, which means losing the competition :-/

massimiliano-mantione:
One final comment, with one more detail on where we could put the timestamp inside the packet payload.

We are already using every available byte, but I think we could repurpose the topic CRC32 four bytes.
First of all, a few observations/questions:

- does it really make sense for a topic to be readable *and* writable by the same device? (I guess no)
- does it really make sense for a single topic to be writable by more than one device? (I hope not, not for our use cases)
- do we *really* want to support more than one published topic per device? (I guess yes :-/ )
- could we accept filtering messages by sender BLE ID before dealing with topics?
- could we accept *mandating* that all the bricks joining a "network" share the full "network topology configuration", including the BLE IDs of all the nodes?

My general idea is to use "sender IDs" to understand "what is being transmitted", and let each sender write ideally to a single topic, and at most a handful of topics (we already have a static limit of four topics for the technic hub...).
This way the 32 bits of the CRC32 would be mostly useless, and they could be repurposed to contain:

- likely 24 bits of "microseconds inside the time window" timestamp information
- likely 8 bits to identify the sender topic, if the sending device has more than one

This would leave the whole radio framework unchanged but would support the "sync dance" described in the previous message.

If you think all of this makes sense, I can *really* help with the implementation on top of this PR!!!

pbochynski:
@massimiliano-mantione I am rather skeptical about using broadcast for real-time synchronization between hubs. There is no guarantee that the message is received and many other factors can break the communication. A reliable synchronization protocol would require acknowledging messages and would introduce a substantial delay in the transmission. I see broadcast as more like an asynchronous communication channel. The main benefit of using it is that you don't need to know the topology. New hubs just join the network and you don't need to pair them. So my suggestion would be to support sharing topics by multiple hubs and collect responses from all devices (now responses are collected per topic hash and the last response wins).
@laurensvalk @dlech I am not sure the discussion should happen in the PR itself - maybe it should be moved to a separate issue. Anyway, when version 3.2 is out you promised to come back to the Bluetooth topic again. What is your plan for merging this PR?

laurensvalk:
Thanks for all the great discussions and ideas everyone! Feel free to open new discussions if you'd like to explore certain topics further.

NStrijbosch:
> The two hubs are not synchronizing.

That is correct, and I think the issue you notice is indeed related to this. 

Ideally the advertising and scanning are completely opposite, i.e., the broadcast process is asynchronous as possible. 
With a constant period length and two hubs this could be possible. The solution for three or more hubs is not as trivial.  

In the current version available [here](https://pybricks.com/projects/tutorials/wireless/hub-to-hub/broadcast/) the constant period length is not exactly constant, leading to the process getting slowly in sync and slowly out of sync again. This can be observed as data drop outs for multiple seconds. 

One solution I could think of to minimize the chance of the process being in sync is to make the period random, which is included by [https://github.com/pybricks/pybricks-micropython/pull/80/commits/f74733521597b2d882a15489747fff2230eb6589](https://github.com/pybricks/pybricks-micropython/pull/80/commits/f74733521597b2d882a15489747fff2230eb6589). This solution should work for any number of hubs.

I did some experiments and this greatly improved the bidirectional communication between two technic hub that transfer data at a high rate. Of course there are the expected data dropouts, but they don't take seconds anymore.

laurensvalk:
I rebased this branch here: https://github.com/pybricks/pybricks-micropython/tree/broadcast-rebase.

The basics still seem to work, but for Prime Hub it now only works if it is not connected to the computer. So I did not force push this branch yet.

laurensvalk:
A pull request with similar functionality inspired by this one has been merged in https://github.com/pybricks/pybricks-micropython/pull/158. Thanks everyone for all the contributions, especially @NStrijbosch !

PR #81, opened 2021-12-23
micropython: Rebase to upstream master branch.

laurensvalk:
Several of our upstream pull requests have been merged so we can drop the respective patches from our fork:

- py/gc: Add hook for events during garbage collect.  https://github.com/pybricks/micropython/commit/ba58e957343fab0cd9dc62c95eb380f9e09de7ee

- pybricks: STM32 UART fixes for BTStack https://github.com/pybricks/micropython/commit/ebf92a34a13e1511dec8e86092b597d2903b88af

- py/runtime: Init dest[1] in convert_member_lookup. https://github.com/pybricks/micropython/commit/c8a0ccb08619f1aa2e21863bfb9ee9692b036d81

- py/runtime: Optional lookup if type->attr fails. https://github.com/pybricks/micropython/commit/16a2a96e928046ea3dac4a9870731407fba355b3

- py/runtime: Return obj->attr for obj.attr if attr has offset type. https://github.com/pybricks/micropython/commit/a98265f6a4108b6fbe5145f0d7532293ea05ca70

Additionally, there were breaking changes to py/builtinimport and mpy-tool, so it is better to upgrade sooner than later.

- py/builtinimport.c: Support builtin-package as nested modules. https://github.com/pybricks/micropython/commit/2a02929521419495227dac0f4b3728bd2c8ef932 broke via https://github.com/micropython/micropython/commit/a7fa18c203a241f670f12ab507aa8b349fcd45a1. It is now fixed via https://github.com/pybricks/micropython/commit/a0e10dde02336031c5544d5f1d4761f845d6d01c

Once we fix all issues, rebasing later should be a bit easier. We can settle on a particular MicroPython version (e.g. v1.18)
when we get closer to Pybricks v3.2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
   settings will not take effect until a new motor command is given.
 - Disabled `Motor.control` and `Motor.log` on Move Hub to save space.
 - Changed LED color calibration on Prime hub to make yellow less green.
+- Updated to upstream MicroPython v1.18.
 
 ### Fixed
 - Fixed color calibration on Powered Up remote control ([support#424]).
--- a/bricks/ev3dev/Makefile
+++ b/bricks/ev3dev/Makefile
@@ -39,10 +39,8 @@ USER_C_MODULES = ../..
 -include mpconfigport.mk
 include ../../micropython/py/mkenv.mk
 
-# use FROZEN_MANIFEST for new projects, others are legacy
+# Frozen Python code
 FROZEN_MANIFEST ?= manifest.py
-FROZEN_DIR =
-FROZEN_MPY_DIR =
 
 # define main target
 PROG = pybricks-micropython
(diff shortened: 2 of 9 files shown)


laurensvalk:
There were also several changes to mpy-cross and mpy-tool that cause the ev3 builds and tests to fail. I'm not sure where to start on these.

Would you want to have a look @dlech?

dlech:
MicroPython has been doing approx. quarterly releases so I would rather just wait for v1.18 to do this.

laurensvalk:
> MicroPython has been doing approx. quarterly releases so I would rather just wait for v1.18 to do this.

Updated to v1.18 now, and fixed the breakage due to https://github.com/micropython/micropython/commit/a7fa18c203a241f670f12ab507aa8b349fcd45a1.

> There were also several changes to mpy-cross and mpy-tool that cause the ev3 builds and tests to fail. I'm not sure where to start on these.

This is still to be done.

PR #82, opened 2022-01-06
pybricks.common: Introduced capability to set volume on Prime/Inventor Hubs.

lobodpav:


CONTRIBUTING.md
```diff
@@ -301,3 +305,28 @@ Then in the `pybricks-micropython` directory:
...
+       
+### Deploy the firmware to a hub
+
+First, follow the guide to turn your Hub into DFU mode:
+1. https://pybricks.com/install, or 
+2. https://dfu.pybricks.com for Spike Prime or Mindstorms Inventor
+                
+Execute the deployment:
+
+    make -C bricks/primehub -j8 deploy-dfu
```
dlech, line 328: As Laurens pointed out, `deploy` should now work for all hubs. 
```suggestion
    make -C bricks/primehub -j8 deploy
```

CONTRIBUTING.md
```diff
@@ -301,3 +305,28 @@ Then in the `pybricks-micropython` directory:
...
+
+For Inventor Hub, use the following command:
```
dlech, line 330: This should not be needed for the reason above.

CONTRIBUTING.md
```diff
@@ -301,3 +305,28 @@ Then in the `pybricks-micropython` directory:
...
+
+Pick your Hub from the `bricks` sub-directory you want to compile.
+
+    poetry shell
+    make mpy-cross -j8
+    make -C bricks/primehub -j8
+       
+### Deploy the firmware to a hub
+
+First, follow the guide to turn your Hub into DFU mode:
```
dlech, line 322: For non-USB hubs, they also need to be in bootloader mode, but technically it is not "DFU" mode since they don't have USB.

pybricks/common/pb_type_speaker.c
```diff
@@ -22,34 +23,90 @@
 #include <pybricks/util_mp/pb_obj_helper.h>
 #include <pybricks/util_pb/pb_error.h>
 
+#define MP_PI MICROPY_FLOAT_CONST(3.14159265358979323846)
```
dlech, line 26: We should change the `MP_` prefix here. Unless we want to suggest making this macro public in upstream MicroPython.

pybricks/common/pb_type_speaker.c
```diff
@@ -22,34 +23,90 @@
...
+// The original sample amplitude must be in the -1..1 range.
+STATIC uint16_t sample_amplifier = INT16_MAX;
+
 STATIC uint16_t waveform_data[128];
 
+// The amount of radians each PCM sample represents to completely fill whe waveform buffer for 2*PI
+//STATIC mp_float_t pcm_sample_size_rad = MP_PI * 2 / MP_ARRAY_SIZE(waveform_data);
+
+STATIC mp_obj_t pb_type_Speaker_get_volume(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
+    return mp_obj_new_int(volume);
```
dlech, line 47: Micro-optimization (works since we know volume can never exceed `MP_SMALL_INT_MAX`
```suggestion
    return MP_OBJ_NEW_SMALL_INT(volume);
```

pybricks/common/pb_type_speaker.c
```diff
@@ -22,34 +23,90 @@
...
+}
+STATIC MP_DEFINE_CONST_FUN_OBJ_KW(pb_type_Speaker_get_volume_obj, 1, pb_type_Speaker_get_volume);
```
dlech, line 49: The pattern we have been using so far in Pybricks is to have a single method for getting and setting.

It would probably be best to start a pull request in the `pybricks-api` repo with the API and docs to work out those kinds of details first, then update the implementation here to match.

pybricks/common/pb_type_speaker.c
```diff
@@ -22,34 +23,90 @@
...
     bool intialized;
 } pb_type_Speaker_obj_t;
 
 STATIC pb_type_Speaker_obj_t pb_type_Speaker_singleton;
 
+STATIC uint8_t volume = 100; // volume in 0..100 range
+
+// The number to multiply the sample amplitude by, to amplify the signal by the defined volume.
+// The original sample amplitude must be in the -1..1 range.
+STATIC uint16_t sample_amplifier = INT16_MAX;
```
dlech, line 39: We should probably make these members of `pb_type_Speaker_obj_t` instead of global variables.

pybricks/common/pb_type_speaker.c
```diff
@@ -22,34 +23,90 @@
...
+        PB_ARG_DEFAULT_INT(volume, 100));
+
+    (void)self; // unused
+    mp_int_t new_volume = pb_obj_get_int(volume_in);
+
+    if (volume <= 100) {
+        volume = new_volume;
+    } else {
+        volume = 100;
+    }
```
dlech, line 63: This can be replaced by the helper function `pb_obj_get_pct()`

pybricks/common/pb_type_speaker.c
dlech, line 39: `attenuation` might be a more technically precise name for this. We can only reduce the amplitude, not increase it.

pybricks/common/pb_type_speaker.c
```diff
@@ -22,34 +23,90 @@
...
+}
+
+//STATIC void pb_type_Speaker_generate_sine_wave(void) {
+//    for (int i = 0; i < MP_ARRAY_SIZE(waveform_data); i++) {
+//        waveform_data[i] = sinf(i * pcm_sample_size_rad) * sample_amplifier + INT16_MAX;
+//    }
+//}
+
+// For 0 frequencies that are just flat lines.
+STATIC void pb_type_Speaker_generate_line_wave(void) {
```
dlech, line 95: If the frequency is 0, we should probably just turn off the DAC and audio amplifier instead of spending energy playing a silent sound. Not sure why I did it this way in the first place.

dlech (commented):
Looks like a great start!

CONTRIBUTING.md
lobodpav: Will try it out, thanks.

CONTRIBUTING.md
lobodpav: OK, will re-phrase that.

pybricks/common/pb_type_speaker.c
lobodpav: OK, will rename it to `PI` only. Pity such a constant is not defined somewhere already.

pybricks/common/pb_type_speaker.c
lobodpav: I have the get/set methods ready on my local already. But not sure at all how to have a single method for getting and setting? It does not make sense to me yet, would you have an example?

pybricks/common/pb_type_speaker.c
lobodpav: How would the `STATIC` methods have access to instance variables? In the Java world, this would not work for static methods.

pybricks/common/pb_type_speaker.c
lobodpav: Yes, will re-factor it.

pybricks/common/pb_type_speaker.c
dlech: example: https://github.com/pybricks/pybricks-micropython/blob/36b53df4d05fe6eddff7d74bf2b79b9a15f9447f/pybricks/pupdevices/pb_type_pupdevices_remote.c#L264-L307

It has a variable number of arguments, 1 or 2. If it is 1, then return the value like the getter. If it is 2, then set the value and return None.

pybricks/common/pb_type_speaker.c
lobodpav: Actually, will delete it, since the we use square wave only at the moment.

CONTRIBUTING.md
lobodpav: Tested and works. Great! Shall we remove the `deploy-dfu` target?

CONTRIBUTING.md
lobodpav: Done.

CONTRIBUTING.md
lobodpav: Done.

pybricks/common/pb_type_speaker.c
lobodpav: Done.

pybricks/common/pb_type_speaker.c
lobodpav: Is this a common pattern in Python? Because this would be a massive anti-pattern in the JVM world 😆

pybricks/common/pb_type_speaker.c
dlech: For the C functions that are MicroPython methods, through the `self` variable. For other functions, it should be passed as a parameter.

CONTRIBUTING.md
dlech: Yes, it would probably be better to only have one way to do a thing.

pybricks/common/pb_type_speaker.c
dlech: It is "old-school" python from before there were properties. It was once common, but not anymore. We are using is because we have very limited flash memory on some hubs and this makes the code as small as possible.

pybricks/common/pb_type_speaker.c
lobodpav: Done.

pybricks/common/pb_type_speaker.c
lobodpav: Done.

pybricks/common/pb_type_speaker.c
lobodpav: The flat line is used when playing `R` note. Is there a way to do that `pause` differently?

dlech: Ah, that is probably why I did it that way. I think we can leave it for now and open an issue to remind us to look at this again later.

CONTRIBUTING.md
lobodpav: Done.

DietrichChristopeit: Just a minuscule side-note: it could maybe confusing to write "turn into..." 
May I suggest to write "switch to.. "
It is clear from the instructions but if something fails for whatever reason, maybe people think that they have to do something else to "turn" the device into something else...

But thank you for that great addition. This whole project is making more and more fun by the day.

-Dietrich-

lobodpav: Yep. Since the DFU mode does not apply to all hubs, I've re-worded the whole sentence to
> 1. Follow the guide to prepare your Hub for Pybricks firmware installation:

😆

pybricks/common/pb_type_speaker.c
lobodpav: Done.

pybricks/common/pb_type_speaker.c
lobodpav: This is interesting. Would you mind elaborating on:
1. why do we have smaller code when using a single method only? Asking because, with a single method, we have to have additional if/else logic which takes quite some space too.
2. how much space-saving does this bring? Asking because I'm curious how the stuff works. Most of the time, developers care about code readability. Here we optimise for memory/space which is a whole different and interesting world.
3. Could we utilise properties somehow to avoid having a single method dealing with two concerns?

laurensvalk: > I have the get/set methods ready on my local already. But not sure at all how to have a single method for getting and setting? It does not make sense to me yet, would you have an example?

For additional context, we have a few like these as well:

image

We could make 6 individual getters and setters for this, or just have this one method. Besides space-savings on the firmware, the same applies to the end-user program. 

That said, it's worth adding that the direction of the project and space saving concerns evolved a bit over time. Some design patterns were established early on, and we do try to keep things somewhat consistent unless changing it is considerably better.

In this particular case, it's also worth considering how the speaker might work in the future. Would there be other speaker settings (waveform, ....? ), and if so, how would those be configured? Separately as done here or perhaps as `speaker.settings(volume, waveform)`?

This is why we usually work out API changes in `pybricks-api` first, and implement them later as David suggested above.

----

> How much space-saving does this bring? 

The build size is printed when you run the make command. We also track the [build size](https://stats.pybricks.com/primehub.html)

lobodpav: API Change PR: https://github.com/pybricks/pybricks-api/pull/92

lobodpav: The image size when using:
1. two get/set methods: `239485`
2. single merged get/set method: `239453`

I.e. saved 32 bytes.

lobodpav: Refactored now to a single `volume` method.

dlech: > I.e. saved 56 bytes.

This is exactly what I was going to guess. :grin: That is the overhead of declaring an extra MicroPython function object. The C functions are compareable is size, I'm sure, but it is the second `MP_DEFINE_CONST_FUN_OBJ_XXX()` that doubles the overhead.

pybricks/common/pb_type_speaker.c
dlech: I imagine that we will eventually have OS sounds and that the OS will take care of resetting the volume on each program run. 

> Wondering what would happen if a Python program creates two Speaker instances though.

For edge cases like this, I usually add a `// REVISIT: ...` comment that explains what I was thinking at the time. Then we can look at it again when we have better infrastructure to handle such things. Currently our cleanup/reset when a user program ends is a bit of a hack.

lobodpav: OK. Will add `REVISIT` comment let the volume be reset upon every instance creation.

lobodpav: Done.

dlech:
If you can split out the unrelated changes (DFU and README) in separate pull requests, we can go ahead and get those merged.

lobodpav:
> If you can split out the unrelated changes (DFU and README) in separate pull requests, we can go ahead and get those merged.

Done.

lobodpav:
Is there anything else I can do to get this PR merged?

dlech:
This is still marked as a draft, so I thought it meant that you were still working on it. Also I would like to get approval from @laurensvalk on https://github.com/pybricks/pybricks-api/pull/92 to make sure we don't have to change it later.

lobodpav:
> This is still marked as a draft, so I thought it meant that you were still working on it. Also I would like to get approval from @laurensvalk on [pybricks/pybricks-api#92](https://github.com/pybricks/pybricks-api/pull/92) to make sure we don't have to change it later.

Oh, silly me. Fixed.

lobodpav:
> Also, I just ordered a sound meter to test it. 😄

I did try that on my iPhone using some decibel app 😆

laurensvalk:
> Also I would like to get approval from @laurensvalk on pybricks/pybricks-api#92 to make sure we don't have to change it later.

I hope to make some time for this early next week. Feel free to tag me again on Monday :)

pybricks/common/pb_type_speaker.c
```diff
@@ -293,7 +333,7 @@ STATIC mp_obj_t pb_type_Speaker_play_notes(size_t n_args, const mp_obj_t *pos_ar
     mp_obj_t iterable = mp_getiter(notes_in, NULL);
     if (nlr_push(&nlr) == 0) {
         while ((item = mp_iternext(iterable)) != MP_OBJ_STOP_ITERATION) {
-            pb_type_Speaker_play_note(self, item, duration);
+            pb_type_Speaker_play_note(self, item, duration, self->sample_attenuator);
```
dlech, line 336: We already pass `self` as an argument, so shouldn't need to pass `self->sample_attenuator` separately.

pybricks/common/pb_type_speaker.c
```diff
@@ -25,31 +26,65 @@
...
+        PB_ARG_DEFAULT_NONE(volume));
+
+    if (volume_in == mp_const_none) {
+        return MP_OBJ_NEW_SMALL_INT(self->volume);
+    }
+
+    self->volume = pb_obj_get_pct(volume_in);
+
+    // exponential amplification (human ear perceives sample amplitude in a logarithmic way)
+    self->sample_attenuator = (powf(10, self->volume / 100.0) - 1) / 9 * INT16_MAX;
```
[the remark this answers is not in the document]
lobodpav: I see. Did not realise that without `F` it defaults to double. Thanks!

pybricks/common/pb_type_speaker.c
lobodpav: Very good catch, thanks! Fixed.

dlech (approved):

laurensvalk:
The following is not critical (so feel free to leave this one as is), but shorter commit titles allow you to read them without unfolding them.

image

Something like `pybricks.common.Speaker: Add volume setter.` probably suffices.

laurensvalk:
Since David has already approved this, I'll merge this as-is.

Thanks for your contribution!

lobodpav:
> The following is not critical (so feel free to leave this one as is), but shorter commit titles allow you to read them without unfolding them.
> 
> image
> 
> Something like `pybricks.common.Speaker: Add volume setter.` probably suffices.

Ahh, 7 extra characters. I'm always trying to fit into that length. Failed this time 😊

lobodpav:
> Here are the results of my testing: image
> 
> I'm not sure why it gets flat on both ends (rounding issues?), but the loundness looks quite linear in spite of that. Nice work!

Thanks for your effort with the measurements!! Even though it seemed linear measuring it by my subjective ears, I'm quite surprised the graph is so close to linear loudness 😊

As per the flat ends, it might be the rounding when too close to 0 and 100 per cent. I simply let the compiler truncate the fractional part by assigning float to the integer variable. Even though rounding might be better, I thought the rounding might be slower.

dlech:
It could just be an error in the measurement too. :wink:

PR #83, opened 2022-01-07
multiple: Fixed typos (`intialized` -> `initialized`)

lobodpav:


dlech:
For some reason I can never spell that right. Thanks!

PR #85, opened 2022-01-10
Improve charger state detection

laurensvalk:
This is an (untested) work in progress to address the charger state detection, which doesn't seem to be quite right at the moment. This is explained in the commit messages.

This is mainly for discussion and needs some work (and testing) before merging.

This is the relevant piece of documentation:

image


dlech:
Nice work. After some testing, it looks like some of the changes are correct and some are not. I will force push my corrections here when I am finished testing.

dlech:
I have a hub that thinks the battery temperature is 95 deg C so I was even able to test the fault detection logic.

PR #86, opened 2022-01-12
bricks/stm32: Removed obsolete `deploy-dfu` targets.

lobodpav:
The `deploy` target can deploy to any Hub as opposed to `deploy-dfu` that did not work for Inventor Hub without explicitly specifying `DFU_PID=0x0011`.


dlech:
Thanks!

PR #92, opened 2022-01-26
update CONTRIB with rust and libjpg install to

SpudGunMan:
new build environment and found some requirements for successful build not listed.

ERROR: Could not build wheels for cryptography requires Rust >=1.41.0
libjpeg, a required dependency when compiling Pillow from source.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,6 +15,7 @@ This is the recommended development environment used by the Pybricks maintainers
 - [GNU ARM Embedded Toolchain][arm-gcc] v10-2020-q4
 - [GNU GCC][gcc] for host operating system
 - [GNU Make][make]
+- [rust][rust] virtual enviroment build requires Rust >=1.41.0
 
 Optional:
 - [Docker][docker] (only needed if building for ev3dev)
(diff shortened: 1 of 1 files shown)


dlech:
What environment are you using where `cryptography` and `Pillow` have to be built from source?

SpudGunMan:
Bullseye/pi

dlech:
OK, can we make a special section for building on Raspberry Pi instead?

SpudGunMan:
If you feel it's helpful may not be that important your documentation was just so complete :)

dlech:
> If you feel it's helpful may not be that important your documentation was just so complete :)

I'm not sure I understand what you mean.

SpudGunMan:
It's a very minor detail easily remedied I just flashed a new pi and noticed it was missing. Thought to toss it up as a suggested change for your very nice contribution/setup guide. 

I didn't want to make it confusing by adding a non standard build how-to or un needed files for others. 

The ARM is handy and it might be worth a detailed section for PI? 

Anyway I wont be sad/hurt if you want to deny the commit :)

Scope creep and keeping up with documentation for other platforms can be a mess as well!

dlech:
Yeah, we don't build on RPi ourselves, so we can't really support this. So thanks, but probably better if we just leave this out.

PR #97, opened 2022-03-07
Fix typos in device naming

ysard:
Hi, this PR is about typo fixs due to wrong copy/paste.

Comments are updated according to https://github.com/pybricks/technical-info/blob/master/assigned-numbers.md

### DIFF
--- a/lib/pbio/test/src/uartdev.c
+++ b/lib/pbio/test/src/uartdev.c
@@ -522,7 +522,7 @@ static PT_THREAD(test_boost_interactive_motor(struct pt *pt)) {
         test_uart_dev.baud == 2400;
     }));
 
-    // send BOOST Color and Distance sensor info
+    // send BOOST Interactive Motor info
     SIMULATE_RX_MSG(msg0);
     SIMULATE_RX_MSG(msg1);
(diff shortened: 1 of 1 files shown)


lib/pbio/test/src/uartdev.c
```diff
@@ -626,7 +626,7 @@ static PT_THREAD(test_boost_interactive_motor(struct pt *pt)) {
 }
 
 static PT_THREAD(test_technic_large_motor(struct pt *pt)) {
-    // info messages captured from Technic Large Linear Motor with logic analyzer
+    // info messages captured from Technic Control+ Large Linear Motor with logic analyzer
```
dlech, line 629: I think Control+ is just the name of the app, not the motors or hubs. e.g. https://www.lego.com/en-us/product/technic-large-motor-88013 doesn't say Control+.

ysard: Ok, I edited the commit accordingly.

laurensvalk:
It would be nice to use the squash button in the GitHub UI which lets us conveniently rename the commit message as needed without local adjustments.

PR #104, opened 2022-07-07
HCL bicone color distance function (without libfixmath)

Novakasa:
This replaces #93, opting not to use libfixmath.

Uses the euclidean distance between colors when mapped into this HCL-bicone:
image

I've tested this quite a bit and found it to be robust. It can distinguish between black and dark bluish gray and NONE very reliably.

I also added calibrated default colors with the `BRICK_` prefix for a few standard lego colors and put them as new defaults. For now, I've chosen to use colors measured at a distance of 2 LEGO plates (=6.4 mm), but I have collected a lot of data so we can use other distance as well. If we chose colors at a larger distance, only their value will be lower, which might make the distance window larger to detect these colors rather than a darker one.

~~The `sensor.detectable_colors()` function now has a new kwarg `chroma_weight` which controls the relationship of relative height to width of the HCL-bicone used to estimate the color distance. Setting it to 0 would be equivalent seeing only black & white.~~

I consider this ready to review, with the following notes:
- I am not experienced with low level optimization. I've assumed that the compiler inlines my integer trigonometry functions and optimizes away some variables used mostly for readability.
- I personally have not seen the need for customizing the `chroma_weight` parameter so it might not be strictly necessary. Though to most users, it will have a useful default (50) and is entirely optional. EDIT: I removed it.
- I have only tested it with the ColorDistanceSensor, since I don't own other sensors and only PUP hubs. However the dependence of different sensors should only come in through the raw-to-hsv conversion which is out of scope of this PR, though I plan on looking at it further after this (https://github.com/pybricks/support/issues/116). If the raw-to-hsv calibration is done, most likely only the calibrated `BRICK_` colors need to be changed. As of now, if the hsv output of the other sensors is at least somewhat consistent with the ColorDistanceSensor, the results should be quite robust for the other sensors as well.
- There is an issue when measuring a white lego brick at the optimal distance of 2 plates, however that seems to be a bug somewhere in the conversion from raw to hsv, as I see the raw RGB values at correct values whereas the hsv values have some suspicious artifacts (saturation jumps from ~0 to ~100 with little change of the raw values). This should probably be addressed when looking at https://github.com/pybricks/support/issues/116

### DIFF
--- a/lib/pbio/src/color/util.c
+++ b/lib/pbio/src/color/util.c
@@ -2,64 +2,37 @@
 // Copyright (c) 2018-2022 The Pybricks Authors
 
 #include <pbio/color.h>
-
-// parabola approximating the first 90 degrees of sine. (0,90) to (0, 10000)
-static int32_t sin_deg_branch0(int32_t x) {
-    return (201 - x) * x;
-}
-
-// integer sine approximation from degrees to (-10000, 10000)
-static int32_t sin_deg(int32_t x) {
-    x = x % 360;
-    if (x < 90) {
-        return sin_deg_branch0(x);
-    }
-    if (x < 180) {
-        return sin_deg_branch0(180 - x);
-    }
-    if (x < 270) {
-        return -sin_deg_branch0(x - 180);
-    }
-    return -sin_deg_branch0(360 - x);
-}
-
-static int32_t cos_deg(int32_t x) {
-    return sin_deg(x + 90);
-}
+#include <pbio/int_math.h>
 
 /**
- * Gets squared Euclidean distance between HSV colors mapped into a chroma-lightness-bicone.
- * The bicone is 20000 units tall and 20000 units in diameter.
+ * Gets squared Euclidean distance between HSV colors mapped into a
+ * chroma-lightness-bicone. The bicone is 20000 units tall and 20000 units in
+ * diameter.
+ *
  * @param [in]  hsv_a    The first HSV color.
  * @param [in]  hsv_b    The second HSV color.
  * @returns              Squared distance (0 to 400000000).
  */
 int32_t pbio_color_get_bicone_squared_distance(const pbio_color_hsv_t *hsv_a, const pbio_color_hsv_t *hsv_b) {
 
-    int32_t a_h = hsv_a->h;
-    int32_t a_s = hsv_a->s;
-    int32_t a_v = hsv_a->v;
+    // Chroma (= radial coordinate in bicone) of a and b (0-10000).
+    int32_t radius_a = pbio_color_hsv_get_v(hsv_a) * hsv_a->s;
+    int32_t radius_b = pbio_color_hsv_get_v(hsv_b) * hsv_b->s;
 
-    int32_t b_h = hsv_b->h;
-    int32_t b_s = hsv_b->s;
-    int32_t b_v = hsv_b->v;
+    // Lightness (= z-coordinate in bicone) of a and b (0-20000).
+    // v is allowed to be negative, resulting in negative lightness.
+    // This can be used to create a higher contrast between "none-color" and
+    // normal colors.
+    int32_t lightness_a = (200 - hsv_a->s) * hsv_a->v;
+    int32_t lightness_b = (200 - hsv_b->s) * hsv_b->v;
 
-    // chroma (= radial coordinate in bicone) of a and b (0-10000)
-    int32_t radius_a = a_v * a_s;
-    int32_t radius_b = b_v * b_s;
-
-    // lightness (= z-coordinate in bicone) of a and b (0-20000)
-    int32_t lightness_a = (200 * a_v - a_s * a_v);
-    int32_t lightness_b = (200 * b_v - b_s * b_v);
+    // z delta of a and b in HSV bicone (-20000, 20000).
+    int32_t delta_z = (lightness_b - lightness_a);
 
     // x and y deltas of a and b in HSV bicone (-20000, 20000)
-    int32_t delx = (radius_b * cos_deg(b_h) - radius_a * cos_deg(a_h)) / 10000;
-    int32_t dely = (radius_b * sin_deg(b_h) - radius_a * sin_deg(a_h)) / 10000;
-    // z delta of a and b in HSV bicone (-20000, 20000)
-    int32_t delz = (lightness_b - lightness_a);
+    int32_t delta_x = (radius_b * pbio_int_math_cos_deg(hsv_b->h) - radius_a * pbio_int_math_cos_deg(hsv_a->h)) / 10000;
+    int32_t delta_y = (radius_b * pbio_int_math_sin_deg(hsv_b->h) - radius_a * pbio_int_math_sin_deg(hsv_a->h)) / 10000;
 
     // Squared Euclidean distance (0, 400000000)
-    int32_t cdist = delx * delx + dely * dely + delz * delz;
-
-    return cdist;
+    return delta_x * delta_x + delta_y * delta_y + delta_z * delta_z;
 }
--- a/lib/pbio/src/color/util.c
+++ b/lib/pbio/src/color/util.c
@@ -1,0 +1,65 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2018-2022 The Pybricks Authors
+
+#include <pbio/color.h>
+
+// parabola approximating the first 90 degrees of sine. (0,90) to (0, 10000)
+static int32_t sin_deg_branch0(int32_t x) {
+    return (201 - x) * x;
+}
+
+// integer sine approximation from degrees to (-10000, 10000)
+static int32_t sin_deg(int32_t x) {
+    x = x % 360;
+    if (x < 90) {
+        return sin_deg_branch0(x);
+    }
+    if (x < 180) {
+        return sin_deg_branch0(180 - x);
+    }
+    if (x < 270) {
+        return -sin_deg_branch0(x - 180);
+    }
+    return -sin_deg_branch0(360 - x);
+}
+
+static int32_t cos_deg(int32_t x) {
+    return sin_deg(x + 90);
+}
+
+/**
+ * Gets squared Euclidean distance between HSV colors mapped into a chroma-lightness-bicone.
+ * The bicone is 20000 units tall and 20000 units in diameter.
+ * @param [in]  hsv_a    The first HSV color.
+ * @param [in]  hsv_b    The second HSV color.
+ * @returns              Squared distance (0 to 400000000).
+ */
+int32_t pbio_color_get_bicone_squared_distance(const pbio_color_hsv_t *hsv_a, const pbio_color_hsv_t *hsv_b) {
+
+    int32_t a_h = hsv_a->h;
+    int32_t a_s = hsv_a->s;
+    int32_t a_v = hsv_a->v;
+
+    int32_t b_h = hsv_b->h;
+    int32_t b_s = hsv_b->s;
+    int32_t b_v = hsv_b->v;
+
+    // chroma (= radial coordinate in bicone) of a and b (0-10000)
+    int32_t radius_a = a_v * a_s;
+    int32_t radius_b = b_v * b_s;
+
+    // lightness (= z-coordinate in bicone) of a and b (0-20000)
+    int32_t lightness_a = (200 * a_v - a_s * a_v);
+    int32_t lightness_b = (200 * b_v - b_s * b_v);
+
+    // x and y deltas of a and b in HSV bicone (-20000, 20000)
+    int32_t delx = (radius_b * cos_deg(b_h) - radius_a * cos_deg(a_h)) / 10000;
+    int32_t dely = (radius_b * sin_deg(b_h) - radius_a * sin_deg(a_h)) / 10000;
+    // z delta of a and b in HSV bicone (-20000, 20000)
+    int32_t delz = (lightness_b - lightness_a);
+
+    // Squared Euclidean distance (0, 400000000)
+    int32_t cdist = delx * delx + dely * dely + delz * delz;
+
+    return cdist;
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,14 @@
 - The `use_gyro` method is added to the normal `DriveBase` class instead of
   having a separate `GyroDriveBase` class. Since the latter was only released
   in beta versions, this is not a breaking change ([support#1054]).
+- New color distance function used by the color sensors that is more
+  consistent when distinguishing user-provided colors ([pybricks-micropython#104]).
 
 ### Fixed
 - Improved external device detection speed ([support#1140]).
 - Fixed Powered Up Tilt Sensor not working  ([support#1189]).
 
+[pybricks-micropython#104]: https://github.com/pybricks/pybricks-micropython/pull/104
 [support#1054]: https://github.com/pybricks/support/issues/1054
 [support#1140]: https://github.com/pybricks/support/issues/1140
 [support#1189]: https://github.com/pybricks/support/issues/1189
--- a/bricks/_common/sources.mk
+++ b/bricks/_common/sources.mk
@@ -190,6 +190,7 @@ PBIO_SRC_C = $(addprefix lib/pbio/,\
 	src/angle.c \
 	src/battery.c \
 	src/color/conversion.c \
+	src/color/util.c \
 	src/control.c \
 	src/control_settings.c \
 	src/dcmotor.c \
--- a/lib/pbio/drv/led/led_pwm.c
+++ b/lib/pbio/drv/led/led_pwm.c
@@ -48,7 +48,7 @@ static pbio_error_t pbdrv_led_pwm_set_hsv(pbdrv_led_dev_t *dev, const pbio_color
     uint32_t Y = ((color->r_brightness * r + color->g_brightness * g + color->b_brightness * b) >> 12) + 1;
 
     // Reapply V from HSV for brightness. V is squared for gamma correction.
-    uint32_t scale_factor = hsv->v * hsv->v * pdata->scale_factor / 128;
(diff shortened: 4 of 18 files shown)


dlech:
FYI, this currently comes in at just under +500 bytes on the move hub firmware. About 1/2 of the size is the new color constants and 1/2 is the new implementation. I don't seen any obvious ways to save a few more bytes (other than leaving out the `chroma_weight` feature).

Novakasa:
This also helps with consistency in distinguishing bright colors from background colors. If we want to distinguish red from dark bluish gray, and add both colors to `detectable_colors` it works most of the time. However, due to the low saturation and value of DBG, small fluctuations in the raw RGB-values can produce a large variation in the hue ( See also some of my plots in https://github.com/pybricks/support/issues/627#issuecomment-1030190409).

Since the weight of the hue-delta in the original cost function largely value- and saturation- independent (save for a cut-off), dark bluish gray would still sometimes be classified as red. We could change the cutoff, but any value will be a bit arbitrary and have some blind spots, so the most robust solution would be to weigh the hue linearly in chroma, which by itself is effectively the same as mapping the colors to a "prism" shape. If we then make the hue influence periodic, we get the HSV-cone.

I have no intuition for the size cost on move hub, but I am guessing that 500 bytes is quite large. Do we have some tools to measure size cost or do I just have to look at the resulting binary? I can try to optimize it more. I could go back to a previous approach where I morphed the bicone to make it match the previous default colors. Will not be as robust with varying distance though and sacrifices some of the simplicity of the distance function.

Would it be feasible to lock this feature behind a `#ifdef NOT_ON_MOVE_HUB`?

dlech:
> Do we have some tools to measure size cost or do I just have to look at the resulting binary?

Yes, it is printed when building the firmware:

```
BIN creating firmware base file
META creating firmware metadata
100524 bytes
```

(there is some variance in the size due to other factors, so consider this +/- 20 bytes of the "real" number)



> Would it be feasible to put this feature behind a `#ifndef MOVE_HUB`?

We have recently freed up quite a bit of space on the move hub so 500 bytes isn't out of the question. If we leave out the `chroma_weight` option, then I think the code size for the improved color discrimination is really quite reasonable. And if we start running out of space again, we could possibly disable the `BRICK_XYX` color constants on move hub to free up some space.

Novakasa:
I removed the `chroma_weight` kwarg. This reduces the size by 152 bytes on move hub. Now this PR adds 322 bytes compared to master.
Thanks to @laurensvalk, I have a Spike sensor now and tested this PR with it. I have noticed no difference in reliability to the `ColorDistanceSensor` even though I am measuring against the same values (calibrated with `ColorDistanceSensor`)

For me this is ready for review or merge if you guys are fine with it.

lib/pbio/src/color/util.c
```diff
@@ -0,0 +1,57 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2018-2022 The Pybricks Authors
+
+#include <pbio/color.h>
+
+int32_t sin_deg_branch0(int32_t x) {
```
dlech, line 6:
```suggestion
static int32_t sin_deg_branch0(int32_t x) {
```

lib/pbio/src/color/util.c
```diff
@@ -0,0 +1,57 @@
...
+    return 201 * x - x * x;
+}
+
+// integer sine approximation from degrees to (-10000, 10000)
+int32_t sin_deg(int32_t x) {
```
dlech, line 11:
```suggestion
static int32_t sin_deg(int32_t x) {
```

lib/pbio/src/color/util.c
```diff
@@ -0,0 +1,57 @@
...
+    if (x < 180) {
+        return sin_deg_branch0(180 - x);
+    }
+    if (x < 270) {
+        return -sin_deg_branch0(x - 180);
+    }
+    return -sin_deg_branch0(360 - x);
+}
+
+int32_t cos_deg(int32_t x) {
```
dlech, line 25:
```suggestion
static int32_t cos_deg(int32_t x) {
```

lib/pbio/src/color/util.c
```diff
@@ -0,0 +1,57 @@
...
+    return sin_deg(x + 90);
+}
+
+// gets squared cartesian distance between hsv colors mapped into a chroma-lightness-bicone
+int32_t pbio_get_bicone_cost(const pbio_color_hsv_t *hsv_a, const pbio_color_hsv_t *hsv_b) {
```
dlech, line 30: Since this is measuring distance, can we call it `pbio_get_bicone_distance` instead of `pbio_get_bicone_cost`?

lib/pbio/src/color/util.c
```diff
@@ -0,0 +1,57 @@
...
+        return -sin_deg_branch0(x - 180);
+    }
+    return -sin_deg_branch0(360 - x);
+}
+
+int32_t cos_deg(int32_t x) {
+    return sin_deg(x + 90);
+}
+
+// gets squared cartesian distance between hsv colors mapped into a chroma-lightness-bicone
```
dlech, line 29: It would be nice to have doxygen style comments for public functions. And to clarify what the units of the return value are.

```suggestion
/**
 * Gets squared Cartesian distance between HSV colors mapped into a chroma-lightness-bicone.
 * @param [in]  hsv_a    The first HSV color.
 * @param [in]  hsv_b    The second HSV color.
 * @returns              This distance (0 to ???).
 */
```

lib/pbio/src/color/util.c
Novakasa: I probably stuck to cost since this actually returns the squared distance. So it probably should be either `pbio_get_bicone_cost` or `pbio_get_bicone_squared_distance`?

dlech: The meaning of "cost" in this context isn't intuitive to me, so it would be nice to either explain what it means in the doc comments or use the more explicit name.

Novakasa: OK, then I would opt for "squared distance". Cost only makes sense in the context of the color map.

Novakasa:
Should be ready now. Also changed
```C
return 201 * x - x * x;
```
to
```C
return (201 - x) * x;
```

laurensvalk:
Rebased for easier review on latest firmware. Review in progress :)

laurensvalk:
This looks nice, thanks for submitting this!

How do you feel about merging the new cost function, but leave it up to the user to add custom brick colors?

Since the `Color` type supports adding extra attributes, that's really easy to do in an end-user script like this.

```python

Color.BRICK_YELLOW = Color(140, 89, 48)
```

We can add an example in the docs with all available colors for users to pick from, including `Color.BRICK_MEDIUM_AZURE` etc.

This saves space and we wouldn't be limited to only a small set of brick colors --- users would have to add color objects anyway.

Novakasa:
If you mean that the default colormap would be empty and the user would always provide his own detectable colors, that sounds good to me.

However, if we went back to the default colormap with default Colors, as discussed in the previous PR (https://github.com/pybricks/pybricks-micropython/pull/93#issuecomment-1105769788), they are not reliably detected with realistic lego brick HSV values (Since e.g. realistic green is closer to `Color.None` than `Color.GREEN` in most metrics other than the previous specialized cost function). It is possible to modify this bicone metric so that it can detect the previous colors to an extent, but that would sacrifice some simplicity and would be a bit arbitrary (see my graphs in the previous PR)

laurensvalk:
A couple of other notes:
- It might be good make the height and width units set with a `#define`. The build size would stay the same but the code may be easier to follow. It's hard to see now how to modify the various hardcoded numbers.
- Is there a rationale behind the current dimensions of the bicone? Or similarly, the weight in the cost function?

laurensvalk:
> This [radial coordinate non-linear in s and v] has the most robust similarity between the measured color and Color.GREEN to my eyes, but also has the most complex computations.
>
>To my eyes, all of these approaches seem better suited than the previous implementation in case the user provides his own calibrated colors [...] Personally, I would probably go ahead with the cone or bicone approach with nonlinear radial coordinate.

_Originally posted by @Novakasa in https://github.com/pybricks/pybricks-micropython/issues/93#issuecomment-1105769788_

Do you still have the formulas for this approach somewhere? Complexity in the cost function shouldn't be a major contributor to build size, so I'm curious about this one.

laurensvalk:
Between the various problems addressed and tested here, is the following still your setup?

image

Novakasa:
> Do you still have the formulas for this approach somewhere? Complexity in the cost function shouldn't be a major contributor to build size, so I'm curious about this one.

For normalized S and V, the radial coordinate would be `V*(2-V)*S*(2-S)` instead of just `S*V`. In practice it would look like `v*(200-v)*s*(200-v)/10000` for values between 0 and 100.

> * Is there a rationale behind the current dimensions of the bicone? Or similarly, the weight in the cost function?

The Lightness (z-coordinate in the cone) can be calculated without division or additional multiplication only if it ranges from 0 to 10000.
Since the x-/y- coordinates have to be rescaled anyway to prevent int overflow for the squared distance, I chose to rescale to the same range as the z-coordinate.
Thus, all the magic numbers in the current implementation are due to the range of the hsv-values (0-100) and factors of 2.

If we added an `#define` for an arbitrary scale, that would introduce some additional multiplications/divisions, but I could do it if that's not that important.
We could probably do a custom scale for the ratio of width/height without additional operations.

> Between the various problems addressed and tested here, is the following still your setup?

Yes

laurensvalk:
Thanks for the additional info.

Don't worry about updating the PR right now though, I'm still reading through all the issues and trying to set up something to compare. Especially after realizing that

> One issue seems to be that the euclidean distance hard-codes the relative weight of chroma and hue.

I don't think scaling the cone would make much of a difference.

If I understand correctly, the original problem was mainly sensitivity to hue on low saturations/values. That's definitely something we can improve but based on some experiments I'm not sure that hard-coding the weight of hue relative to chroma is the best way to do it.

> Yes, though I have since moved on to processing the hsv values instead of using the sensor.color() method

What does your new approach look like? Would it be better to just make the color map configurable?

laurensvalk:
I'd also be curious to hear about specific problematic cases, so I can test them on my end too.

E.g. green bricks on a gray track with a black background.

Novakasa:
So my initial issue was like this: I had custom colors for the black background, the DBG tracks, and red and blue plates. I wanted to detect the red and blue plates and ignore the black and DBG colors. However, the color() method would sometimes return the red and blue color when it actually measured the DBG tracks. The reason for that is that the saturation fluctuates quite a bit for DBG, for the same reason that the saturation does not encode any real information for pure black.

That's the reason that to me it would make the most sense to rather take into account the chroma value than the saturation. However, that way we can't control the saturation/value weight individually anymore, which is a problem when trying to associate colors with their S=V=100 counterpart, but for any cost function that actually wants to measure the perceptual distance of colors, this is a good approach.

Once again, the core difficulty here is that we want a cost function that works well for custom calibrated colors as well as the default ones, which is why to me it seemed best to replace the default colors with realistic ones, but I am biased here since I only ever want to compare realistic colors.



> What does your new approach look like? Would it be better to just make the color map configurable?

My new approach is a quite simple filter based on the chroma value, and then checking the hue only for the filtered colors. It works very well so I doubt I would replace it with a colormap approach again, but if there would have been the possibility to provide custom color distance function I would have used that with success as well.

laurensvalk:
Thanks a lot for summarizing the issue and solutions so far, that really helps.

Just to clarify, what does DBG stand for? Is that a train specific term? Or is it perhaps LEGO Dark Bluish Gray?

laurensvalk:
Awesome. I'm going to build some trains and [connect](https://pybricks.com/projects/tutorials/wireless/hub-to-device/pc-communication/) some Jupyter notebooks.

By the way, we have two new project sections [on our site](https://pybricks.com/projects/). If you want to, you're welcome to create e.g. a remix project to share your project, this time with no restrictions (:smile:), or link to your own page as well.

Novakasa:
> However, the color() method would sometimes return the red and blue color when it actually measured the DBG tracks. The reason for that is that the saturation fluctuates quite a bit for DBG, for the same reason that the saturation does not encode any real information for pure black.

Correction: I think the reason for the wrong color is that the value/saturation of DBG is high enough to not ignore the hue in the cost function (as you quoted), but the hue should still be quite irrelevant when comparing darker colors.

In the cone approach, the hue (angle coordinage) contribution is always scaled by the chroma (radial coordinate), so this issue is eliminated.

Novakasa:
> That sounds a little bit like what we do here, but perhaps with a better filter.

Indeed, it is a similar thought, but the choice of the filter was very specific to my problem (s*v > 3500), and a more aggressive filter there could hinder the performance when trying to distinguish darker colors from each other (rather than comparing darker colors to bright colors)

Novakasa:
https://github.com/pybricks/support/issues/627#issuecomment-1030842007 Here I showed some plots to compare the various properties of the color cost functions.

laurensvalk:
> Awesome. I'm going to build some trains.

It does look better on DBG train tracks for sure :wink:

Novakasa:
Thanks for giving this another chance! The timing is quite neat, as I have found some situations in my project where my previous approach for detecting markers needs some more sophistication. To have full control though, I might need access to the hsv cost function in a user program. Would this be too costly to implement? In either case, if this PR gets merged I can just do a draft PR to expose the cost function to see how expensive it is.

Interesting idea to make None have negative value! If this makes the color detection robust with all default colors (Including green and white, which was problematic for me), this seems to be the best approach if we want to avoid adding calibrated colors.

Unfortunately, I might have difficulty testing this currently, as I don't have all the colors available to me currently (my collection has been moved). Maybe I can find some LEGO around the house in a few hours.

laurensvalk:
That could certainly be an interesting addition. I can have a look. 

Also if you have found a better cost function since your original contribution, it would be easy enough to swap it in.

Novakasa:
My previous approach was to record all hues above a certain chroma threshold and then average over the hues to match against the "marker hues" after chroma is below the threshold again. This worked flawlessly in my experience. Recently, I had to relax this to match already the first hue after the chroma threshold is met (to reject certain colors),  and now I seem to get color detection issues from time to time. I did not have time to investigate this in full detail yet, but it seems like the detection worked better when I averaged over multiple hue samples. The first sample might capture the "transition" between two colors, which is a possible reason for the degraded reliability in this approach.

Writing this, I am honestly not sure why I now only take the first sample now, I might have had a good reason for this but maybe not, I need to check. (maybe writing this just solved my issues)

Novakasa:
I think the HCL bicone is the best function that is a good compromise between simplicity and perceptual consistency, the only alternative to me is the regular HSV cone, but I find it unintuitive that white is exactly between two colors with opposite hues and max chroma. I think that in the HSV cone, Green would also be closer to white than None,  so then the `None.v=-30` trick would not work. (https://github.com/pybricks/pybricks-micropython/pull/93#issuecomment-1105769788)

laurensvalk:
Thanks for the input! Perhaps to achieve a similar effect, there could be some form of hysteresis. Here are two partially overlapping variants/ideas:

- When a new color is detected, don't switch until it has reached, say, 55% of the way there or after a timeout was reached. This may help avoid frequent jumps between adjacent hyperspaces in the map. The 55% may be defined as the cost difference between the last detected color and the new best match.
- Or, keep a count and update the color on some number of consecutive matches. For best performance, the map would need to be tied to the incoming sensor data (I think 100 Hz), but this would not be safe with a user-defined cost function.

But since this PR changes (almost) nothing for the user, we could always start by getting this merged and building on the more advanced features later.

Since you're positive about the suggestions, I'll update this PR with those changes so we can review and test it in more detail.

laurensvalk:
When the pull request finishes building in half an hour or so, you can find the ready-made firmware files here: https://github.com/pybricks/pybricks-micropython/pull/104#issuecomment-1620253919

Novakasa:
OK, I just did some testing using this build. I was using this script:

```python
from pybricks.pupdevices import ColorDistanceSensor
from pybricks.parameters import Color, Port, Button

sensor = ColorDistanceSensor(Port.B)

last_color = None
counter = 0

hub = ThisHub()

LBG = Color(h=217, s=46, v=60)
DBG = Color(h=207, s=48, v=39)

Color.DBG = DBG
Color.LBG = LBG

sensor.detectable_colors([Color.RED, Color.GREEN, Color.BLACK, Color.BLUE, Color.YELLOW, Color.DBG, Color.WHITE, Color.LBG])

hub.system.set_stop_button(None)

while True:
    color = sensor.color()
    counter += 1
    if color != last_color:
        print(counter, color)
        last_color = color
        counter = 0
    if hub.button.pressed():
        print(sensor.hsv())
```

And everything seems to work as expected. I had to calibrate DBG and LBG carefully (measure at the same distance), because they mostly differ in the brightness, and the brightness is very sensitive to distance. (seems to be inverse square law, see also https://github.com/pybricks/support/issues/116#issuecomment-1114057472)

(uncalibrated) Green seems to be detected reliably, and None still works when pointing into the air. So, to me, this looks good!

laurensvalk:
Thanks for the quick test! Glad to hear the basics are working for you.

In the tutorials, when people change the detectable colors, I think we’ll want to recommend that they measure all their colors instead of combining a few measurements with some of the builtin colors, for optimal performance.

With the updated map, it was able to distinguish between the dbg sleepers, my desk underneath, a red tile, and a blue tile, which is pretty neat. Thanks for this, and sorry it took so long!

Novakasa:
Btw, I was testing with city hub + ColorDistanceSensor

laurensvalk:
Added a commit to move some of the math, and clean up some of the steps.

The only algorithmic change is that I clamped the lower v bound at 0 for radius computations. It probably makes sense for the negative V to only affect the delta z. (Ran out of time for testing today.)

laurensvalk:
It would be nice to also do a `1:1` test with the default settings on this firmware and master on a different hub side by side.

If we can get ballpark the same performance, that would be nice. We can adjust the `-30` as needed to get a similar threshold as before.

laurensvalk:
It's merged! :tada:

Novakasa:
Thank you for seeing this through to the end!

PR #112, opened 2022-08-23
pbdrv/block_device_flash_stm32: Implement data storage.

laurensvalk:
This follows up on https://github.com/pybricks/pybricks-micropython/pull/110 to implement persistent storage on Powered Up hubs to make the user experience consistent with the SPIKE Hubs.

This changes the user experience since the built in program would no longer be needed or used. However, Pybricks Code should continue to work since it caps the firmware+mpy size to never reach the user flash region.

Some things to consider:
- [ ] Yield the (otherwise blocking) block device store protothread between erases/writes. Although writing the maximum size takes only about 50--100 ms, so we could just leave it. ---> Deferred to https://github.com/pybricks/support/issues/708
- [ ] Not use HAL to save some space as suggested by @dlech. This could possibly be done later if we make a dedicated issue for it. ---> Deferred to https://github.com/pybricks/support/issues/708
- [ ] Select the right sizes for user flash on each hub. The current values are a bit too small to be useful for this use case. ---> Deferred to https://github.com/pybricks/support/issues/708
- [x] Update commit message since this is not as much of a breaking change as initially thought.
- [x] Test on City Hub. It doesn't seem to work yet. Test generic shutdown behavior with button disabled.


laurensvalk:
@bricklife :steam_locomotive: :train: :train:

lib/pbio/drv/block_device/block_device_flash_stm32.c
```diff
@@ -28,34 +30,112 @@ uint32_t pbdrv_block_device_get_size(void) {
...
+
+    // Erase only as much as we need.
+    FLASH_EraseInitTypeDef erase_init = {
+        #if defined(STM32F0)
+        .PageAddress = base_address,
+        #elif defined(STM32L4)
+        .Banks = FLASH_BANK_1,
+        .Page = (FLASH_SIZE - (pbdrv_block_device_get_size())) / FLASH_PAGE_SIZE,
+        #else
+        #warning "Unsupported target."
```
dlech, line 85: Should probably be `error` instead of `warning`.

lib/pbio/drv/block_device/block_device_flash_stm32.c
```diff
@@ -28,34 +30,112 @@ uint32_t pbdrv_block_device_get_size(void) {
...
+    if (hal_err != HAL_OK) {
+        return PBIO_ERROR_IO;
+    }
+
+    // Erase only as much as we need.
+    FLASH_EraseInitTypeDef erase_init = {
+        #if defined(STM32F0)
+        .PageAddress = base_address,
+        #elif defined(STM32L4)
+        .Banks = FLASH_BANK_1,
```
dlech, line 82: Let's add a comment here explaining why bank 1 is selected. I.e. hard-coded for Technic Hub MCU and may need to be looked at again if support for another MCU is added.

lib/pbio/drv/block_device/block_device_flash_stm32.c
```diff
@@ -28,34 +30,112 @@ uint32_t pbdrv_block_device_get_size(void) {
...
+        .Page = (FLASH_SIZE - (pbdrv_block_device_get_size())) / FLASH_PAGE_SIZE,
+        #else
+        #warning "Unsupported target."
+        #endif
+        .NbPages = size / FLASH_PAGE_SIZE + (size % FLASH_PAGE_SIZE != 0),
```
dlech, line 87:
```suggestion
        .NbPages = (size + (FLASH_PAGE_SIZE - 1) / FLASH_PAGE_SIZE,
```

This is a more idiomatic way to round up while dividing in C.

lib/pbio/drv/block_device/block_device_flash_stm32.c
```diff
@@ -28,34 +30,112 @@ uint32_t pbdrv_block_device_get_size(void) {
...
+        .TypeErase = FLASH_TYPEERASE_PAGES
+    };
+
+    // Disable interrupts while erasing.
```
dlech, line 91: A more useful comment would explain why interrupts need to be disabled.

lib/pbio/drv/block_device/block_device_flash_stm32.c
```diff
@@ -28,34 +30,112 @@ uint32_t pbdrv_block_device_get_size(void) {
...
+        HAL_FLASH_Lock();
+        return PBIO_ERROR_IO;
+    }
 
+    // Write data chunk by chunk.
+    uint32_t done = 0;
+    while (done < size) {
+
+        // Adjust write type for last page.
+        uint32_t type = FLASH_TYPEPROGRAM_DOUBLEWORD;
```
dlech, line 109: Why do we need a variable for this if it never changes?

lib/pbio/drv/block_device/block_device_flash_stm32.c
```diff
@@ -28,34 +30,112 @@ uint32_t pbdrv_block_device_get_size(void) {
...
+
+        // Cast data to a doubleword.
+        double_word_t *data = (double_word_t *)(buffer + done);
```
dlech, line 112: why do we need the union type when only the uint64_t field is used? Couldn't we just cast to `uint64_t *`?

lib/pbio/drv/block_device/block_device_flash_stm32.c
laurensvalk: Thanks. This was when I first tried the fast (32-dword) write, but I ended up choosing something that was available on all hubs. Removed.

lib/pbio/drv/block_device/block_device_flash_stm32.c
laurensvalk: It was to ensure correctness (set as bytes, read uint64_t), but we can drop this since it is a platform specific driver indeed.

PR #115, opened 2022-08-24
Enable multi-file projects

laurensvalk:
DESCRIPTION TODO

### DIFF
--- a/bricks/stm32/main.c
+++ b/bricks/stm32/main.c
@@ -95,6 +95,42 @@ static const pbsys_user_program_callbacks_t user_program_callbacks = {
     .stdin_event = user_program_stdin_event_func,
 };
 
+typedef struct _pb_reader_user_data_t {
+    const byte *beg;
+    const byte *cur;
+    const byte *end;
+} pb_reader_user_data_t;
+
+mp_uint_t pb_reader_user_data_readbyte(void *data) {
+    pb_reader_user_data_t *blob = (pb_reader_user_data_t *)data;
+    if (blob->cur < blob->end) {
+        return *blob->cur++;
+    } else {
+        return MP_READER_EOF;
+    }
+}
+
+STATIC void pb_reader_user_data_close(void *data) {
+    (void)data;
+}
+
+const uint8_t *pb_reader_user_data_readchunk(void *data, size_t len) {
+    pb_reader_user_data_t *blob = (pb_reader_user_data_t *)data;
+    const uint8_t *ptr = blob->cur;
+    blob->cur += len;
+    return ptr;
+}
+
+STATIC void pb_reader_user_data_new(mp_reader_t *reader, pb_reader_user_data_t *data, const byte *buf, size_t len) {
+    data->beg = buf;
+    data->cur = buf;
+    data->end = buf + len;
+    reader->data = data;
+    reader->readbyte = pb_reader_user_data_readbyte;
+    reader->readchunk = pb_reader_user_data_readchunk;
+    reader->close = pb_reader_user_data_close;
+}
+
 static void run_user_program(uint32_t len, uint8_t *buf, bool run_repl) {
 
     #if MICROPY_ENABLE_COMPILER
@@ -125,7 +161,8 @@ restart:
         } else {
             // run user .mpy file without erasing it
             mp_reader_t reader;
-            mp_reader_new_mem(&reader, buf, len, 0);
+            pb_reader_user_data_t data;
+            pb_reader_user_data_new(&reader, &data, buf, len);
             mp_module_context_t *context = m_new_obj(mp_module_context_t);
             context->module.globals = mp_globals_get();
             mp_compiled_module_t compiled_module = mp_raw_code_load(&reader, context);
--- a/bricks/stm32/main.c
+++ b/bricks/stm32/main.c
@@ -25,6 +25,7 @@
 #include "py/gc.h"
 #include "py/mperrno.h"
 #include "py/mphal.h"
+#include "py/objmodule.h"
 #include "py/persistentcode.h"
 #include "py/repl.h"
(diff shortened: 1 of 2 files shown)


bricks/stm32/main.c
```diff
@@ -189,6 +332,64 @@ mp_obj_t mp_builtin_open(size_t n_args, const mp_obj_t *args, mp_map_t *kwargs)
...
+
+    // Pybricks extension: Search for modules in downloaded user data.
+    module_obj = pb_import_loaded_module(n_args, args);
+    if (module_obj != MP_OBJ_NULL) {
+        return module_obj;
+    }
+
+    // Couldn't find the module, so fail
+    #if MICROPY_ERROR_REPORTING <= MICROPY_ERROR_REPORTING_TERSE
+    mp_raise_msg(&mp_type_ImportError, MP_ERROR_TEXT("module not found"));
```
dlech, line 387: If we really want this to be terse, we could omit the message altogether. As it is, this message is only a few bytes shorter than the other one, so its not really saving much.

bricks/stm32/main.c
```diff
@@ -95,6 +96,72 @@ static const pbsys_user_program_callbacks_t user_program_callbacks = {
     .stdin_event = user_program_stdin_event_func,
 };
 
+typedef struct _pb_reader_user_data_t {
+    const byte *beg;
```
dlech, line 100: It looks like `beg` is assigned but never ready, so possibly could be removed.

bricks/stm32/main.c
```diff
@@ -189,6 +332,64 @@ mp_obj_t mp_builtin_open(size_t n_args, const mp_obj_t *args, mp_map_t *kwargs)
...
+
+    // Return the module we found.
+    return module_obj;
+}
+
+// Overrides MicroPython's mp_builtin___import__
+mp_obj_t pb_builtin_import(size_t n_args, const mp_obj_t *args) {
+    // Check that it's not a relative import
+    if (n_args >= 5 && MP_OBJ_SMALL_INT_VALUE(args[4]) != 0) {
+        mp_raise_NotImplementedError(MP_ERROR_TEXT("relative import"));
```
dlech, line 369: How hard would it be to implement relative imports?

bricks/stm32/main.c
```diff
@@ -95,6 +96,72 @@ static const pbsys_user_program_callbacks_t user_program_callbacks = {
...
+    data->beg = buf;
+    data->cur = buf;
+    data->end = buf + len;
+    reader->data = data;
+    reader->readbyte = pb_reader_user_data_readbyte;
+    reader->readchunk = pb_reader_user_data_readchunk;
+    reader->close = pb_reader_user_data_close;
+}
+
+static void do_execute_raw_code(mp_module_context_t *context, const mp_raw_code_t *rc, const mp_module_context_t *mc) {
```
dlech, line 135: We should mention that this is copied from MicroPython in case it is made public in the future and there is a possibility of sharing the code.

bricks/stm32/main.c
```diff
@@ -157,6 +225,66 @@ static void run_user_program(uint32_t len, uint8_t *buf, bool run_repl) {
...
+     * Data follows thereafter.
+     */
+}  __attribute__((scalar_storage_order("little-endian"))) mpy_info_t;
+
+// Rounds up a given size to an integer multiple of 4.
+static uint32_t padded(uint32_t len) {
+    if (len % 4 == 0) {
+        return len;
+    }
+    return len + 4 - (len % 4);
```
dlech, line 249: A slightly more concise way to do this is `(len + 3) / 4 *  4`.

bricks/stm32/main.c
```diff
@@ -157,6 +225,66 @@ static void run_user_program(uint32_t len, uint8_t *buf, bool run_repl) {
...
+    uint32_t mpy_size;
+    /**
+     * Null-terminated name of the script, padded with additional zeros to
+     * make the size an integer multiple of 4.
+     */
+    char mpy_name[];
+    /**
+     * Data follows thereafter.
+     */
+}  __attribute__((scalar_storage_order("little-endian"))) mpy_info_t;
```
dlech, line 242: Let's leave out the attribute here too.

bricks/stm32/main.c
```diff
@@ -157,6 +225,66 @@ static void run_user_program(uint32_t len, uint8_t *buf, bool run_repl) {
...
+typedef struct {
+    /**
+     * Size of the mpy program.
+     */
+    uint32_t mpy_size;
+    /**
+     * Null-terminated name of the script, padded with additional zeros to
+     * make the size an integer multiple of 4.
+     */
+    char mpy_name[];
```
dlech, line 238: Should there be an upper bound on file name size? Or maybe would be simpler to just use a fixed size? Or use one byte to store the string length?

dlech: Is this stored the name `module.mpy` or just `module`?

bricks/stm32/main.c
laurensvalk: If Pybricks Code only has one folder, what would we use it for? If it is mainly to encourage users to use relative imports, then yes we could just ignore the leading `.` to make it work, but I suppose I'm not sure that's helpful.

If we need more functionality, we could just enable external imports.

bricks/stm32/main.c
laurensvalk: Removed, thanks.

bricks/stm32/main.c
laurensvalk: This was coming from the default import, but I agree we can drop it in the terse case.

laurensvalk:
I'm merging this so we have a working point to start from, but the functionality and implementation/format are subject change and open to user feedback.

PR #117, opened 2022-08-26
bricks/stm32: Force exit when shutdown requested.

laurensvalk:
If a user catches the `SystemExit` exception indefinitely, there is currently not a way to power off the hub.

This PR makes the following program exit normally so we can still de-init pbsys and do a clean power off.

```python

while True:
    try:
        wait(10)
    except SystemExit:
        print("nope!")
```

I'm not merging this yet because there may be a cleaner way to exit MicroPython using `PYEXEC_FORCED_EXIT` and / or various available hooks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,17 @@
 ### Added
 - Added support for `PBIO_PYBRICKS_COMMAND_REBOOT_TO_UPDATE_MODE` Pybricks Profile BLE command.
 
+### Changed
+- The Pybricks Code stop button will force the program to exit even if the user
+  catches the `SystemExit` exception ([pybricks-micropython#117]).
+
 ### Fixed
 - Fixed connecting `Remote` on BOOST move hub ([support#793]).
 
 ### Removed
 - Removed `hub.system.reset()` method.
 
+[pybricks-micropython#117]: https://github.com/pybricks/pybricks-micropython/pull/117
 [support#793]: https://github.com/pybricks/support/issues/793
 
 ## [3.2.0b5] - 2022-11-11
--- a/bricks/_common/micropython.c
+++ b/bricks/_common/micropython.c
@@ -32,6 +32,17 @@
 #include "py/stackctrl.h"
 #include "py/stream.h"
 
+// Implementation for MICROPY_EVENT_POLL_HOOK
+void pb_event_poll_hook(void) {
+    while (pbio_do_one_event()) {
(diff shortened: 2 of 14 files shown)


dlech:
Does this work if you replace the `wait()` and `print()` with `pass` so the the event poll hook never runs?

laurensvalk:
> Does this work if you replace the wait() and print() with pass so the the event poll hook never runs?

```python
while True:
    try:
        pass
    except:
        pass
```

Yes and no - this already cleanly exits on a single press because most of the time is not spent in the try part.

laurensvalk:
This, however, does still get stuck:

```python

while True:
    try:
        while True:
            try:
                wait(10)
            except SystemExit:
                print("nope!")
    except SystemExit:
        pass
```

laurensvalk:
> This, however, does still get stuck:

We can detect the situation just fine, nothing truly gets stuck, it's a matter of telling MicroPython to stop.

Raising a specific exception could work, but the user may still use a bare except.

laurensvalk:
Another possibility: With the recent code refactoring, reintroducing the `longjmp` can be done now by confining it to just the main application:

image

Along with this to call it:

image

This costs about 200 bytes instead of the 500 bytes from before (?).

This freezes the hub though (watch dog), so this isn't quite right.

laurensvalk:
> I had a thought as I was falling asleep last night that maybe we could do something like the long jump, but put it in the MicroPython code so that we could make use of the MicroPython "nlr" code to hopefully keep the code size down.

That could be nice too.

`pbsys_main_stop_program` always gets called correctly, so we could jump from there.

For generality, it would be nice to change:

```diff
-void pbsys_main_stop_program(void)
+void pbsys_main_stop_program(bool force_stop)
```

That way the micropython application code doesn't have to poke at pbsys, just exit when it it told to.

laurensvalk:
I think I've figured out how to make this work. When a forced exit is triggered we can pop the `nlr` to the top level that was used to start the main program or REPL, and then raise the `SystemExit` to make it jump to the end. Does that make sense? See updated PR.

Separately, I've added a commit to also use this when pressing the stop button in Pybricks Code. Thoughts? Until now the stop button could theoretically be used as an "input", but we have the I/O terminal for that as well (on most hubs). It seems nice if the stop button always makes it stop.

laurensvalk:
Feel free to merge this in if it looks right, I may be offline for a bit.

PR #121, opened 2022-10-21
pbsys/program_load: Add read/write access to user data.

laurensvalk:
This feature has been requested many times (e.g. [here](https://www.eurobricks.com/forum/index.php?/forums/topic/186132-pybricks-qa/), [here](https://github.com/orgs/pybricks/discussions/623), or [here](https://github.com/pybricks/support/issues/85)). Now that we have enabled storage to save programs, we might as well add some storage for user data.

Proposed API:

```python
# To write some data:
hub.system.storage(offset=5, write=b"Hello, world!")

# To read some data:
data = hub.system.storage(offset=12, read=6)
print(data) # This will print b"world!"
```
Appropriate errors are be raised so read/write access is always safe.

We could also add a `storage_view` later on for hubs that can support it, but this should cover the basics.

### DIFF
--- a/pybricks/common/pb_type_system.c
+++ b/pybricks/common/pb_type_system.c
@@ -8,12 +8,16 @@
 #include <string.h>
 
 #include <pbdrv/bluetooth.h>
+#include <pbsys/program_load.h>
 
 #include "py/obj.h"
+#include "py/objstr.h"
 #include "py/runtime.h"
 
 #include <pybricks/common.h>
 #include <pybricks/util_pb/pb_error.h>
+#include <pybricks/util_mp/pb_kwarg_helper.h>
+#include <pybricks/util_mp/pb_obj_helper.h>
 
 STATIC mp_obj_t pb_type_System_name(void) {
     const char *hub_name = pbdrv_bluetooth_get_hub_name();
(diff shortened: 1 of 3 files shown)


laurensvalk:
It would be great to have this part of the next beta so it will get some proper testing / feedback.

laurensvalk:
Demo time!

https://user-images.githubusercontent.com/12326241/197208154-02750e82-edf4-4a33-9670-0bfa36374123.mp4

laurensvalk:
We are expediting merging this so users will be able to try it out in the beta.

This API is still open for changes after the beta release; we won't lock it until the full V3.2 release.

pybricks/common/pb_type_system.c
```diff
@@ -101,6 +105,34 @@ STATIC mp_obj_t pb_type_System_shutdown(void) {
...
+    }
+
+    // Handle write.
+    if (read_in == mp_const_none && !mp_obj_is_str(write_in) && mp_obj_is_str_or_bytes(write_in)) {
+        mp_obj_str_t *obj = ((mp_obj_str_t *)MP_OBJ_TO_PTR(write_in));
+        pbsys_program_load_set_user_data(offset, obj->data, obj->len);
+        return mp_const_none;
+    }
+
+    mp_raise_ValueError(MP_ERROR_TEXT("Must set either read (int) or write (bytes)."));
```
dlech, line 132: Many Pythonistas would expect `TypeError` rather than a `ValueError` (we get here based on type checks after all).

pybricks/common/pb_type_system.c
```diff
@@ -101,6 +105,34 @@ STATIC mp_obj_t pb_type_System_shutdown(void) {
...
+        pb_assert(pbsys_program_load_get_user_data(offset, &data, size));
+        return mp_obj_new_bytes(data, size);
+    }
+
+    // Handle write.
+    if (read_in == mp_const_none && !mp_obj_is_str(write_in) && mp_obj_is_str_or_bytes(write_in)) {
+        mp_obj_str_t *obj = ((mp_obj_str_t *)MP_OBJ_TO_PTR(write_in));
+        pbsys_program_load_set_user_data(offset, obj->data, obj->len);
+        return mp_const_none;
+    }
```
dlech, line 130: How about allowing any object that uses the buffer protocol instead of strictly `bytes` to avoid casting/copying?

```suggestion
   if (read_in == mp_const_none) {
        mp_buffer_info_t bufinfo;
        mp_get_buffer_raise(read_in, &bufinfo, MP_BUFFER_READ);
        
        pb_assert(pbsys_program_load_set_user_data(offset, bufinfo.buf, bufinfo.len));
        return mp_const_none;
    }
```

pybricks/common/pb_type_system.c
```diff
@@ -101,6 +105,34 @@ STATIC mp_obj_t pb_type_System_shutdown(void) {
...
+        byte *data;
+        mp_uint_t size = mp_obj_get_int(read_in);
+        pb_assert(pbsys_program_load_get_user_data(offset, &data, size));
+        return mp_obj_new_bytes(data, size);
+    }
+
+    // Handle write.
+    if (read_in == mp_const_none && !mp_obj_is_str(write_in) && mp_obj_is_str_or_bytes(write_in)) {
+        mp_obj_str_t *obj = ((mp_obj_str_t *)MP_OBJ_TO_PTR(write_in));
+        pbsys_program_load_set_user_data(offset, obj->data, obj->len);
```
dlech, line 128: missing `pb_assert()`

pybricks/common/pb_type_system.c
laurensvalk: Yep, this is the one I mentioned I was hoping to fix.

laurensvalk:
Feel free to merge in all three of your suggestions above.

PR #130, opened 2022-12-07
bricks/_common/micropython: Fix program stop message.

laurensvalk:
This avoids mentioning the stop button so that it is correct for all SystemExit or SystemAbort exceptions.

Fixes https://github.com/pybricks/support/issues/836


dlech:
Might want to sleep on it to see if there is any other considerations we didn't think of, but this looks good to me.

laurensvalk:
Processing PRs is the new counting sheep :smile:

laurensvalk:
It shouldn't be worse than the existing variant, so I'll merge it.

PR #133, opened 2022-12-26
Fixes gyro read using the wrong scaling and data indices for IMU data (angular_rate)

CasperClarke:
The code found here:
https://github.com/pybricks/pybricks-micropython/blob/af76eecd6bfaff97c9e6435ed16d4bc97b9a946e/lib/pbio/drv/imu/imu_lsm6ds3tr_c_stm32.c#L283-L298
is sourcing both the accel and gyro values from the same source.

The original code was using the accel scaling and the data from the accelerometer, indices 0-2, while the gyro data is stored in indices 3-5 as notated here:
[pybricks-micropython/lib/pbi](https://github.com/pybricks/pybricks-micropython/blob/af76eecd6bfaff97c9e6435ed16d4bc97b9a946e/lib/pbio/drv/imu/imu_lsm6ds3tr_c_stm32.c#L240-L246)

This fork fixes the angular_rate function for all hubs to return the correct value in deg/s

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 ## [Unreleased]
 
+### Fixed
+- Fixed `imu.angular_velocity` returning the values of `imu.acceleration`.
+
+[support#885]: https://github.com/pybricks/support/issues/885
+
 ## [3.2.0] - 2022-12-20
 
 ### Changed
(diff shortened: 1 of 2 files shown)


laurensvalk:
Thanks! This was accidentally changed recently in https://github.com/pybricks/pybricks-micropython/commit/af76eecd6bfaff97c9e6435ed16d4bc97b9a946e.

PR #135, opened 2022-12-28
Adds Mahony filter based IMU orientation estimation.

CasperClarke:
This branch adds a Mahony filter that processes the gyro and accelerometer data to create a quaternion estimation of the hub's rotation.
 
This branch adds a few python bindings in order to read the heading of the vehicle, and its rotation quaternion, calibrate the gyro and modify the Mahony filter gains. 

This branch also modifies the existing function that made use of the accelerometer to determine the pitch and roll of the hub. 
Now the IMU estimation is used for the tilt function. 

Over the span of 30 minutes, a drift of 0.5 degrees was observed with a good calibration of the gyro bias on a technichub. 

The axis of roll and pitch and heading need to be validated, I expect they may be inverted from the intended direction (CCW Positive). 

Another slight problem: If the start gyro calibration function is started and the program is then ended, it will continue running until the stop gyro calibration function is called. This could result in unintended behavior if a program has a start calibration and does not have a matching stop gyro calibration. 

This is a side effect of the code being run beneath the python interpreter, the code running the gyro estimation does not stop between runs of python programs. This also has the beneficial side effect of gyro calibration being stored until a power cycle occurs or it is overwritten.


laurensvalk:
Thanks for this contribution! It may take us a while to get back with feedback, but we will try to do so in the new year.

dlech:
Thanks again for the work on this. In the mean time, we've made some progress on implementing orientation estimation in a different way.

PR #138, opened 2023-01-10
pybricks/util_mp/pb_obj_helper: fix crash caused by PB_ATTRIBUTE_TABLE

dlech:
Using the key of the locals dict to store non-qstr info causes the builtin help() method to crash. Instead we can use the protocol slot to avoid the crash. The attr entry arrays are now zero terminated instead of storing the size (since the entries are < 32-bits we are not using any more space than we would if we stored the size somewhere else where it had to be 32-bit aligned).

Fixes: https://github.com/pybricks/support/issues/909


dlech:
bonus: also saves 100 bytes on move hub

laurensvalk:
Thanks. I just finished dinner and you've already fixed it :smile:

I'm not entirely sure the size savings will still hold for MicroPython 1.20 since we are now using an extra slot (protocol), but otherwise it looks good to me.

I still think that the table reference in the globals dict is fine, but hacking the size onto the MP_QSTRnull was probably optimizing it a bit too much indeed.

dlech:
> I'm not entirely sure the size savings will still hold for MicroPython 1.20 since we are now using an extra slot (protocol), but otherwise it looks good to me.

I think a slot is only one pointer vs. two pointers for a locals dict entry.

PR #140, opened 2023-02-01
pbio/drv/imu_lsm6ds3tr_c_stm32: Read gyro and acceleration in one pass.

laurensvalk:
This should further reduce overhead.

### DIFF
--- a/lib/pbio/drv/imu/imu_lsm6ds3tr_c_stm32.c
+++ b/lib/pbio/drv/imu/imu_lsm6ds3tr_c_stm32.c
@@ -208,10 +208,13 @@ static PT_THREAD(pbdrv_imu_lsm6ds3tr_c_stm32_init(struct pt *pt)) {
     /* Gyroscope - filtering chain */
     // PT_SPAWN(pt, &child, lsm6ds3tr_c_gy_band_pass_set(&child, ctx, LSM6DS3TR_C_HP_16mHz_LP1_LIGHT));
 
-    // configure interrupts
+    // Configure INT1 to trigger when new gyro data is ready.
     PT_SPAWN(pt, &child, lsm6ds3tr_c_pin_int1_route_set(&child, ctx, (lsm6ds3tr_c_int1_route_t) {
         .int1_drdy_g = 1,
     }));
+
+    // If we leave the default latched mode, sometimes we don't get the INT1 interrupt.
+    PT_SPAWN(pt, &child, lsm6ds3tr_c_data_ready_mode_set(&child, ctx, LSM6DS3TR_C_DRDY_PULSED));
 
     // Enable rounding mode so we can get gyro + accel in one read.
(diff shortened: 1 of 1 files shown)


laurensvalk:
When we do this, we technically shouldn't even need a process anymore.

We could just start reading on the pin1 interrupt, and do the memcpy on getting the read_complete interrupt.

I suppose error handing is easier with the process, let's keep an eye on whether the reset is still triggered a lot.

lib/pbio/drv/imu/imu_lsm6ds3tr_c_stm32.c
```diff
@@ -213,6 +213,9 @@ static PT_THREAD(pbdrv_imu_lsm6ds3tr_c_stm32_init(struct pt *pt)) {
         .int1_drdy_g = 1,
     }));
 
+    // Enable rounding mode so we can get gyro + accel in one read.
+    PT_SPAWN(pt, &child, lsm6ds3tr_c_rounding_mode_set(&child, ctx, LSM6DS3TR_C_ROUND_GY_XL));
```
dlech, line 217: This is not needed. We only read 12 bytes so we never hit the rounding point.

lib/pbio/drv/imu/imu_lsm6ds3tr_c_stm32.c
```diff
@@ -228,7 +231,7 @@ PROCESS_THREAD(pbdrv_imu_lsm6ds3tr_c_stm32_process, ev, data) {
     I2C_HandleTypeDef *hi2c = &imu_dev->hi2c;
 
     static struct pt child;
-    static uint8_t buf[6];
+    static uint8_t buf[12];
```
dlech, line 234: We are using 12 enough times we should probably introduce a `#define` for it.

lib/pbio/drv/imu/imu_lsm6ds3tr_c_stm32.c
dlech: Alternately to actually take advantage of this, we would need to use different I2C functions to write the register once outside of the loop and then read inside of the loop (`HAL_I2C_Master_Transmit_IT` and `HAL_I2C_Master_Receive_IT` or possibly `HAL_I2C_Master_Sequential_Receive_IT`?) and a goto after error recovery to write the register again.

laurensvalk: Or something like the note above with no loop at all?

dlech: Doing more in the interrupt handlers sounds dangerous to me. For example, if we call HAL functions from there, they may have (short) blocking busy loop waits that we don't know about. The less we do in interrupt handlers the better.

And I don't see how that would make using the rounding function work either.

dlech:
I went ahead and made the suggested changed and fixed some other unrelated issues.

PR #145, opened 2023-02-14
pbio/drivebase: change to dynamic allocation

dlech:
This changes the allocation of `pbio_drivebase_t` "objects" to be dynamically allocated in MicroPython instead of being statically allocated in the PBIO library. (Technically, other consumers of the PBIO library could still statically allocate them in their application code).

This requires enabling the finalizer in MicroPython on all ports to ensure that the the drive base objects are removed from PBIO before the memory is freed by MicroPython.


laurensvalk:
One potential issue for library users could be the introduced Contiki dependency.

laurensvalk:
Thanks! The dynamic allocation part looks good, but I think we can do without the linked list in `pbio` because we already keep track of the servo/parent objects from which we can get the info needed to run the drive base update.

dlech:
> I think we can do without the linked list

What alternative do you propose for finding all drive base instances? Iterating all motors and testing if they are part of a drive base?

laurensvalk:
Something like that. I'll try working it out in a bit.

laurensvalk:
Archiving this since we won't be working on this in the near future.

PR #146, opened 2023-02-18
Improve EV3 loop time and motor performance

laurensvalk:
Before the change, it looked like this. Speed estimates are way off:

image

After this change, it looks like this. The motor makes a perfect speed trapazoid:

image

### DIFF
--- a/bricks/ev3dev/pbinit.c
+++ b/bricks/ev3dev/pbinit.c
@@ -39,7 +39,7 @@ static pthread_t task_caller_thread;
 static void *task_caller(void *arg) {
     struct timespec ts;
     ts.tv_sec = 0;
-    ts.tv_nsec = PBIO_CONFIG_CONTROL_LOOP_TIME_MS * 1000000;
+    ts.tv_nsec = 2000000;
 
     while (!stopping_thread) {
         MP_THREAD_GIL_ENTER();


laurensvalk:
Are there any downsides to doing this? Is 1ms polling too much?

laurensvalk:
I don't quite understand these first couple of samples:

image

Only the first sample is allowed to be way off, because the user may give their command any time within the 5ms window. But after that, I would expect it to hit the average sooner than later. This is not a one off, it happens every time.

laurensvalk:
Also need to test this with 4 motors running.

dlech:
> Only the first sample is allowed to be way off, because the user may give their command any time within the 5ms window. But after that, I would expect it to hit the average sooner than later. This is not a one off, it happens every time.

If the timer in `pbio_motor_process` gets behind, it will run every 1 ms until it is caught up. So it is probably getting behind when the `pybricks-micropython` runtime is starting (e.g. we probably don't release GIL during all of the blocking sysfs reads/writes) and it is still catching up when the motor command is called. I suspect if you add a `wait(100)` before the motor call, you will see the expected results.

dlech:
> I'm not sure why the CI is failing yet; I can't seem to reproduce it locally.

I suspect `sysfs_write_str` is failing and the error is bubbling up. We could add some debugging there. All of these syscalls should probably have a retry if they are interrupted.

laurensvalk:
Maybe that explains the long beep too. I was also seeing more delays throughout the program.

laurensvalk:
Merged so we can make further progress on motors.

We can open a new issue if timing becomes an issue elsewhere. (like the beep mentioned above)

PR #149, opened 2023-02-25
Allow loading binary .mpy files

pepijndevos:
As is, trying to load a native .mpy file will result in an error: `ValueError: incompatible .mpy arch` which can be traced down to a check that depends on this define.

Setting it to one seems to cause an error, which I tried to fix here: https://github.com/micropython/micropython/pull/10855


dlech:
After fixing the MicroPython bug, it looks like this feature requires 15KB of flash memory. So if we do decide to enabled this by default, we could only do so on the SPIKE hubs (they are the only ones that have 1MB of flash memory).

Have you been successful in building and running a native module after enabling this?

pepijndevos:
Yes I have successfully ran the factorial example from [here](https://docs.micropython.org/en/latest/develop/natmod.html) as well as statically compiled Julia code.

So I guess we'd need to overwrite the define in the SPIKE makefile instead. Is the user code also stored in this 1MB? It may not be that common to want to run native code on these things.

dlech:
> Is the user code also stored in this 1MB?

No, it is stored on the external flash.

> So I guess we'd need to overwrite the define in the SPIKE makefile instead.

We should add a new flag like `PYBRICKS_OPT_EXTRA_MOD` for this.

pepijndevos:
Well in that case there seem not much harm in having more features on the spike hub.

I also assume that EV3, using an SD card has space for this.

(though the instructions on the website mention only official lego firmware for the ev3?)

bricks/_common/mpconfigport.h
```diff
@@ -40,7 +40,7 @@
 #define MICROPY_ALLOC_PATH_MAX                  (256)
 #define MICROPY_ALLOC_PARSE_CHUNK_INIT          (16)
 #define MICROPY_EMIT_X64                        (0)
-#define MICROPY_EMIT_THUMB                      (0)
+#define MICROPY_EMIT_THUMB                      (PYBRICKS_OPT_EXTERNAL_MOD)
```
dlech, line 43: I think `EXTERNAL` has a different meaning already in MicroPython (e.g. see `MICROPY_ENABLE_EXTERNAL_IMPORT`). Perhaps a better name is `PYBRICKS_OPT_NATIVE_MOD`.

dlech: This should also depend on `__thumb2__` in case this config file is used on a different or older arch.
```suggestion
#define MICROPY_EMIT_THUMB                      (PYBRICKS_OPT_NATIVE_MOD && __thumb2__)
```

dlech (changes requested):
FYI, I added a few commits to pull in the required fix to upstream MicroPython and the pybricksdev changes.

bricks/_common/mpconfigport.h
pepijndevos: Would you not simply set it to false on any device with an older CPU?

dlech: For older Arm archs, sure, makes sense. But it still makes sense to do this for compatibility with non-Arm: 

```c
#define MICROPY_EMIT_X64                        (PYBRICKS_OPT_NATIVE_MOD && __x86_64__)
#define MICROPY_EMIT_THUMB                      (PYBRICKS_OPT_NATIVE_MOD && __thumb2__)
```

pepijndevos: Alright

dlech:
Thanks!

PR #150, opened 2023-02-25
bricks/virtualhub: Build without CPython.

laurensvalk:
This uses the test drivers instead of the CPython implementation. This is easier to debug and maintain, and it runs a lot faster.

No specialized configuration is needed to run it anymore. You can just run the virtualhub-micropython executable as if it is normal MicroPython. It just so happens to have a working `pybricks` package.

The robot setup is defined in `platform.c`. It currently uses 5 motors, but this can be changed later to simulate a few sensors. Lights and buttons are not done yet, which is a regression for now.

The goal is to get to a virtualhub that we'll use more often (as opposed to using DFU on a real Prime Hub all the time). The existing implementation is very extensible and potentially very powerful, but sometimes a bit cumbersome to work with (which could be me).

This is what the updated implementation can currently do (although I haven't pushed the *optional* React app that makes this visualization yet):

https://user-images.githubusercontent.com/12326241/221375380-f7f759c5-763b-45b2-a3ea-6d09629c1556.mp4

---------

Some next steps (not in this PR):
- Remove the CPython drivers from `lib/pbio`
- Clean up and publish the React app used in the video. Since this is aimed at just developers for now, I suppose we could include it here instead of making yet another repo (thoughts?).
- Implement lights, light matrix, buttons.
- Implement sensors and incorporate them into the 2D visualization, e.g. a motor that presses the force sensor, and a motor that spins colored bricks in front of the color sensor.
- Revise how the data piping works so we can include lights and handle on-click button events.

### DIFF
--- a/bricks/virtualhub/mp_port.c
+++ b/bricks/virtualhub/mp_port.c
@@ -11,15 +11,13 @@
 #include <sys/select.h>
 
 #include <contiki.h>
-#include <Python.h>
 
 #include <pbio/main.h>
 #include <pbsys/core.h>
 #include <pbsys/program_stop.h>
 #include <pbsys/status.h>
 
-#include "../../lib/pbio/drv/virtual.h"
-
+#include "py/mphal.h"
 #include "py/mpconfig.h"
 #include "py/obj.h"
 #include "py/objexcept.h"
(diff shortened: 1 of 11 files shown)


dlech:
> * Since this is aimed at just developers for now, I suppose we could include it here instead of making yet another repo (thoughts?).

My preference is to keep it in this repo.

dlech:
One one hand, it would be nice to use a high level library like [glib](https://docs.gtk.org/glib/) to replace the CPython layer for handling communications, on the other hand, it makes the code less portable (i.e. it would be nice to be able to compile for the web using emscripten).

bricks/virtualhub/mp_port.c
```diff
@@ -181,7 +119,7 @@ void pb_virtualhub_event_poll(void) {
 
     struct timespec timeout = {
         .tv_sec = 0,
-        .tv_nsec = 1000000,
```
dlech, line 184: Why does this need to be faster? It should just be the 1ms tick.

laurensvalk: This entire section can go in principle. The clock_linux driver drives the clock tick now.

laurensvalk: 100*1000:

image

1000*1000:

image

laurensvalk: CPU overhead isn't too bad at all with 100*1000.

dlech: Does 500*1000 work just as well?

laurensvalk:
> > * Since this is aimed at just developers for now, I suppose we could include it here instead of making yet another repo (thoughts?).
> 
> My preference is to keep it in this repo.

Will that lead to a bunch of dependabot pull requests or can we disable this? Since this will only be used locally by the developers, we can probably do without constant updates.

dlech:
> Will that lead to a bunch of dependabot pull requests or can we disable this?

Dependabot is not enabled. We will just get security updates.

PR #151, opened 2023-03-02
Virtual Hub live animation

laurensvalk:
This is a working proof of concept that shows how we can optionally show a live animation of the virtual hub state:

https://user-images.githubusercontent.com/12326241/221375380-f7f759c5-763b-45b2-a3ea-6d09629c1556.mp4

To try it:
```
cd tools/virtual-hub-animator
yarn install
yarn start
```
Also run:
```
# You can just start the virtual hub debug launch task, or alternatively do this:

make virtualhub -j
PBIO_TEST_DATA_PARSER=./tools/virtual-hub-animator/data_server.py ./bricks/virtualhub/build/virtualhub-micropython
```

This starts the REPL, where you can interact with the motors as shown in the video.

Note that use of this is optional. Running `virtualhub-micropython` standalone is fine and motors will still work. There is just no visualization that way.


laurensvalk:
Quite a lot of clean up is probably needed to make this a proper app. I've not done much with JS/TS or React before.

It does prove quite useful for everyday motor debugging, so maybe we can fix the worst of this for now, and defer the larger cleanups to a separate issue:

- Only motors are supported. No lights, sensors, etc.
- The data comes out of the virtual motor driver process, but the generic implementation should probably be placed in something like `mp_port.c`.
- Only state output is supported (one way communication, so no buttons/events)

PR #158, opened 2023-04-14
BLE Broadcast v2

dlech:
This is an alternate attempt at hub to hub communication using BLE advertising data similar #80. (Thanks again @NStrijbosch and @laurensvalk  for all of the work done there.)

### Differences from #80

- Advertising and scanning and being connected to Pybricks Code and being connected to a remote on the Technic hub all at the same time seems to be working, so there is no background process to switch between advertising and scanning. The previous observed problems could have been caused by bugs in the existing Bluetooth driver code that have (hopefully) been fixed since then.
- Support for Move hub is added.
- Since LEGO has announced the discontinuation of MINDSTORMS, compatibility with the official firmware seems less important, so this has a different data encoding scheme and uses non-connectable advertisements.
  - The new encoding scheme allows us to squeeze in more data.
  - Using non-connectable advertisements is the "right" way to do this according to BLE author recommendations and prevents other devices from connecting while broadcasting. It also sends data less frequently (every 100ms) which should reduce radio interference.
- Different Python API that better fits new encoding scheme.

### Limitations

- Broadcast (sending non-connectable advertisements in any form) does not work on City hub. This appears to be a firmware bug (it has the same chip as Technic hub but different firmware version).
- Broadcast on Move hub only works when not connected to Pybricks Code, otherwise it raises `OSError` with `EPERM`.
- Using broadcast and observe on the Move hub at the same time has very inconsistent (and usually very high) latency.
- Using observe while connected to Pybricks Code results in inconsistent latency on smaller hubs. SPIKE hubs seem to work well even when connected to Pybricks Code and city/technic hubs work better when not connected to Pybricks Code.

### Python API

This adds a new `ble` attribute to the `Hub` object with the following methods and extends the `Hub` constructor with additional arguments.

```python
class BLE:
    def broadcast(*args: None | bool | int | float | str | bytes) -> None:
        """
        Starts broadcasting advertising data containing *args*.

        Params:
            *args: Zero or more values to be broadcast.

        .. note:: Advertising data size is quite limited. For example, you
            can send one string or byte array that is 25 bytes or 12
            ``True``/``False`` values or 5 floating point values or any
            combination of values as long as the total packed size doesn't
            exceed the available space. The exact technical specification
            can be found at ...

        Example::

            # Broadcast 3 values.
            hub.ble.broadcast(100, "hi", False)

            # Broadcast no values. Useful if you only care about RSSI.
            hub.ble.broadcast()

        Bad example::

            # Dont' do this! The first value will never be broadcast.
            hub.ble.broadcast("first")
            hub.ble.broadcast("second")
        """

    def observe(channel: int) ->  tuple[None | bool | int | float | str | bytes, ...] | None:
        """
        Retrieves the last observed data for a given channel.

        Args:
            channel: The channel to observe (0 to 255).

        Returns:
            A tuple of the most recent RSSI and the advertising data or ``None`` if no data
            has been received yet or it has been more than one second since the last data
            was received.

        Example::

            data = hub.ble.observe(channel=0)

            if data is not None:
                a, b, c = data
                # if the other hub called hub.ble.broadcast(100, "hi", False)
                # then a == 100 and b == "hi" and c is False
                ...

        Bad example::

            # Don't do this! This will fail the first time it is called because no
            # data has been received yet causing unpacking to fail.
            a, b, c = hub.ble.observe(channel=0)
        """

    def signal_strength(channel: int) -> int:
    """
    Gets the average signal strength in dBm for the given channel.

    This is useful for detecting how near the broadcasting device is. A close
    device may have a signal strength around -40 dBm while a far away device
    might have a signal strength around -70 dBm.

    Args:
        channel: The channel number (0 to 255).

    Returns:
        The signal strength or ``-128`` if nothing has been received yet or the
        broadcasts are no longer being received.
    """

    def version() -> str:
        """
        Gets the firmware version from the Bluetooth chip.

        Example::

            print(hub.ble.version())
        """


class Hub(..., broadcast_channel=1, observe_channels=[]):
    """
    Params:
        broadcast_channel:
            A value from 0 to 255 indicating which channel ``hub.ble.broadcast()``
            will use. Default is channel 0.
        observe_channels:
            A list of channels channel to listen to when
            ``hub.ble.observe()`` is called. Listening to more channels
            requires more memory. Default is an empty list (no channels).

    Example::

        # Broadcast on channel 1. Observe channels 2 and 3.
        hub = PrimeHub(broadcast_channel=1, observe_channel=[2, 3])
    """
    ...
    ble: BLE
```

### Ideas for changes to current implementation

- (fixed) ~~Known issue: RSSI is initially 0 instead of -127.~~
- (done) ~~RSSI should change to -127 after some timeout of not receiving new advertisement data. How long should this be? 1 second, 10 seconds?~~
- (done) ~~`last_observe_channel` should be changed `observe_channels` and take an iterable of channels. The range of allowable channel values for both broadcast and observe can then be expanded to 0 to 255 instead of 0 to 15. This would be nice to make it easier to avoid using the same channels as your neighbor.~~
- (done) ~~Furthermore, BLE scanning could be enabled at construction time rather than the first time `hub.ble.observer()` is called.~~
- State in documentation that broadcasting is not supported on Move hub and City hub and recommend using nRF UART instead if needed on those hubs. Broadcast would be modified to to simply always raise an error on those hubs (would probably save quite a bit of firmware size on Move hub which is always a good thing).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Added
 - Enabled the `gc` module (except on BOOST Move hub).
+- Added `hub.ble` attribute for accessing Bluetooth capabilities.
 
 ### Changed
 - Updated MicroPython to v1.20.0.
--- a/bricks/_common/sources.mk
+++ b/bricks/_common/sources.mk
@@ -22,6 +22,7 @@ CONTIKI_SRC_C = $(addprefix lib/contiki-core/,\
 # Pybricks modules
 
 PYBRICKS_PYBRICKS_SRC_C = $(addprefix pybricks/,\
+	common/pb_type_ble.c \
 	common/pb_type_battery.c \
 	common/pb_type_charger.c \
 	common/pb_type_colorlight_external.c \
--- a/bricks/cityhub/mpconfigport.h
+++ b/bricks/cityhub/mpconfigport.h
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-// Copyright (c) 2018-2021 The Pybricks Authors
+// Copyright (c) 2018-2023 The Pybricks Authors
 
 #include "stm32f030xc.h"
 
@@ -12,6 +12,7 @@
 
 // Pybricks modules
 #define PYBRICKS_PY_COMMON                      (1)
+#define PYBRICKS_PY_COMMON_BLE                  (1)
 #define PYBRICKS_PY_COMMON_CHARGER              (0)
 #define PYBRICKS_PY_COMMON_CONTROL              (1)
 #define PYBRICKS_PY_COMMON_IMU                  (0)
--- a/bricks/debug/mpconfigport.h
+++ b/bricks/debug/mpconfigport.h
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-// Copyright (c) 2019-2021 The Pybricks Authors
+// Copyright (c) 2019-2023 The Pybricks Authors
 
 #include "stm32f446xx.h"
 
@@ -11,6 +11,7 @@
 
 // Pybricks modules
 #define PYBRICKS_PY_COMMON                      (1)
+#define PYBRICKS_PY_COMMON_BLE                  (0)
 #define PYBRICKS_PY_COMMON_CHARGER              (0)
 #define PYBRICKS_PY_COMMON_CONTROL              (0)
 #define PYBRICKS_PY_COMMON_IMU                  (0)
--- a/bricks/essentialhub/mpconfigport.h
+++ b/bricks/essentialhub/mpconfigport.h
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-// Copyright (c) 2019-2021 The Pybricks Authors
+// Copyright (c) 2019-2023 The Pybricks Authors
 
 #include "stm32f413xx.h"
 
@@ -13,6 +13,7 @@
 
 // Pybricks modules
 #define PYBRICKS_PY_COMMON                      (1)
+#define PYBRICKS_PY_COMMON_BLE                  (1)
 #define PYBRICKS_PY_COMMON_CHARGER              (1)
 #define PYBRICKS_PY_COMMON_CONTROL              (1)
 #define PYBRICKS_PY_COMMON_IMU                  (1)
--- a/bricks/ev3dev/brickconfig.h
+++ b/bricks/ev3dev/brickconfig.h
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-// Copyright (c) 2018-2021 The Pybricks Authors
+// Copyright (c) 2018-2023 The Pybricks Authors
 
 #include <pbdrv/config.h>
 #include "pbinit.h"
@@ -16,6 +16,7 @@
 
 // Pybricks modules
 #define PYBRICKS_PY_COMMON              (1)
+#define PYBRICKS_PY_COMMON_BLE          (0)
 #define PYBRICKS_PY_COMMON_CHARGER      (0)
 #define PYBRICKS_PY_COMMON_CONTROL      (1)
 #define PYBRICKS_PY_COMMON_IMU          (0)
--- a/bricks/ev3rt/mpconfigport.h
+++ b/bricks/ev3rt/mpconfigport.h
@@ -16,6 +16,7 @@
 
 // Pybricks modules
 #define PYBRICKS_PY_COMMON              (1)
+#define PYBRICKS_PY_COMMON_BLE          (0)
 #define PYBRICKS_PY_COMMON_CHARGER      (0)
 #define PYBRICKS_PY_COMMON_CONTROL      (0)
 #define PYBRICKS_PY_COMMON_IMU          (0)
--- a/bricks/movehub/mpconfigport.h
+++ b/bricks/movehub/mpconfigport.h
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-// Copyright (c) 2018-2021 The Pybricks Authors
+// Copyright (c) 2018-2023 The Pybricks Authors
 
 #include "stm32f070xb.h"
 
@@ -12,6 +12,7 @@
 
 // Pybricks modules
 #define PYBRICKS_PY_COMMON                      (1)
+#define PYBRICKS_PY_COMMON_BLE                  (1)
 #define PYBRICKS_PY_COMMON_CHARGER              (0)
 #define PYBRICKS_PY_COMMON_CONTROL              (0)
 #define PYBRICKS_PY_COMMON_IMU                  (0)
--- a/bricks/nxt/mpconfigport.h
+++ b/bricks/nxt/mpconfigport.h
@@ -14,6 +14,7 @@
 
 // Pybricks modules
 #define PYBRICKS_PY_COMMON                      (1)
+#define PYBRICKS_PY_COMMON_BLE                  (0)
 #define PYBRICKS_PY_COMMON_CHARGER              (0)
 #define PYBRICKS_PY_COMMON_CONTROL              (1)
 #define PYBRICKS_PY_COMMON_IMU                  (0)
--- a/bricks/primehub/mpconfigport.h
+++ b/bricks/primehub/mpconfigport.h
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-// Copyright (c) 2019-2021 The Pybricks Authors
+// Copyright (c) 2019-2023 The Pybricks Authors
 
 #include "stm32f413xx.h"
 
@@ -14,6 +14,7 @@
 
 // Pybricks modules
 #define PYBRICKS_PY_COMMON                      (1)
+#define PYBRICKS_PY_COMMON_BLE                  (1)
 #define PYBRICKS_PY_COMMON_CHARGER              (1)
 #define PYBRICKS_PY_COMMON_CONTROL              (1)
 #define PYBRICKS_PY_COMMON_IMU                  (1)
--- a/bricks/technichub/mpconfigport.h
+++ b/bricks/technichub/mpconfigport.h
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-// Copyright (c) 2018-2021 The Pybricks Authors
+// Copyright (c) 2018-2023 The Pybricks Authors
 
 #include "stm32l431xx.h"
 
@@ -12,6 +12,7 @@
 
 // Pybricks modules
 #define PYBRICKS_PY_COMMON                      (1)
+#define PYBRICKS_PY_COMMON_BLE                  (1)
 #define PYBRICKS_PY_COMMON_CHARGER              (0)
 #define PYBRICKS_PY_COMMON_CONTROL              (1)
 #define PYBRICKS_PY_COMMON_IMU                  (1)
--- a/bricks/virtualhub/mpconfigvariant.h
+++ b/bricks/virtualhub/mpconfigvariant.h
@@ -8,6 +8,7 @@
 
 // Pybricks modules
 #define PYBRICKS_PY_COMMON              (1)
+#define PYBRICKS_PY_COMMON_BLE          (0)
 #define PYBRICKS_PY_COMMON_CHARGER      (1)
 #define PYBRICKS_PY_COMMON_CONTROL      (1)
 #define PYBRICKS_PY_COMMON_IMU          (0)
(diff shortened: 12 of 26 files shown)


dlech:
Here are the programs I was using during testing:

```python
from pybricks.parameters import Port, Color

hub = ThisHub()
motor = Motor(Port.A)

print("radio fw", hub.ble.version())

hub.light.on(Color.GREEN)

prev = None

while True:
    rssi, data = hub.ble.observe(0)
    
    if data:
        target = data[0]
        motor.track_target(target)

        if target != prev:
            print(rssi, data)
            prev = target
    else:
        motor.track_target(0)

    wait(10)
```

```
from pybricks.pupdevices import Remote, Motor
from pybricks.parameters import Color, Button, Port

hub = ThisHub(broadcast_channel=1)
motor = Motor(Port.A)

print("radio fw", hub.ble.version())

hub.light.on(Color.YELLOW)
remote = Remote()

hub.light.on(Color.GREEN)
remote.light.on(Color.GREEN)


def button_oneshot(button: Button):
    prev = False
    oneshot = None

    while True:
        new = yield oneshot
        
        if new is None:
            continue

        now = button in new

        if now == prev:
            oneshot = None
        else:
            oneshot = prev = now

def make_button_oneshot(button):
    gen = button_oneshot(button)
    gen.send(None)
    return gen


left_plus_oneshot = make_button_oneshot(Button.LEFT_PLUS)
left_minus_oneshot = make_button_oneshot(Button.LEFT_MINUS)

right_plus_oneshot = make_button_oneshot(Button.RIGHT_PLUS)
right_minus_oneshot = make_button_oneshot(Button.RIGHT_MINUS)

local_target = 0
remote_target = 0

while True:
    motor.track_target(local_target)

    pressed = remote.buttons.pressed()

    if left_plus_oneshot.send(pressed):
        local_target += 45

    if left_minus_oneshot.send(pressed):
        local_target -= 45

    if right_plus_oneshot.send(pressed):
        remote_target += 45
        hub.ble.broadcast(remote_target)

    if right_minus_oneshot.send(pressed):
        remote_target -= 45
        hub.ble.broadcast(remote_target)

    wait(10)
```

This uses the LEGO Powered Up remote to control 2 motors, one connected to each of two different hubs.

The left side of the controller controls the motor on the hub the remote is connected to and the right side of the controller controls the motor connected to the secondary hub.

dlech:
For higher-level documentation, we should also explain proper usage of this. The main point to get across is that this type of communication is best suited to sending _output state_ rather than commands/events or input state. (The technical reason being this is unreliable (delivery is not guaranteed), low-bandwidth communication.) This is why in my test program, it is sending the target motor angle rather than button press events or button state.

laurensvalk:
Nice work!

Would it be better if the received data is packed the same as the sent data? In this example the sender has:

```python
hub.ble.broadcast(target)
```

but the receiver has:

```python
target = data[0]   # Why is unpacking needed here, but packing not needed on sender?
```

It would be nice if packing and unpacking was the same, or even absent altogether for single-valued data.

_Originally posted by @laurensvalk in https://github.com/pybricks/pybricks-micropython/commit/a02c18c02b7c7c373d2ee7e43e0bfa4d1da166cb#r108901780_

> last_observe_channel should be changed observe_channels and take an iterable of channels

:+1:

dlech:
> Would it be better if the received data is packed the same as the sent data?

According to intellesense, it is the same. The packing in a tuple is just implicit because of the *args.

image

dlech:
updated

laurensvalk:
> > Would it be better if the received data is packed the same as the sent data?
>
> According to intellesense, it is the same. 

If it's the same in a technical sense, it would be nice to pick the simplest for users (which might not be the `*` operator).

I've added a proposal to get it up the same level as #80. So allowing int/float/bool/str or a tuple thereof. It is quite a small change, without adding any extra encoding fields.

I've also added a 1 second timeout so that you can easily detect that the sender has stopped.

Do we want a way to stop broadcasting? Since `None` is already used as valid data, we can't use that. Perhaps using no args (the empty tuple) is the most intuitive way to achieve that? This matches what you see on the receiving end when there was no data to begin with. If you really want to send "nothing" just to provide RSSI, you could always just send `None`.

Personally, I would perhaps rather drop `None` as an allowed data type, and use it to indicate no data, but either will work.

laurensvalk:
> Thank you for the suggestion. However, I am really very happy with the simplicity of the the way I had implemented it. 

Having implemented the MicroPython module side of #80 originally, so was I :smile:

I've moved the explorations to a separate branch.

It still doesn't seem that intuitive to me to treat data on the receiving end different (two unpacking) from the sending end (no packings). Even if intellisense says it is the same, I usually look at it in terms of the resulting code.  I'll explorer a bit further on this one.

laurensvalk:
This is good by the way --- heated API debates remind me of 2018, and we've made some pretty good stuff since then :smile:

dlech:
Updated with some suggestions from Laurens.

Known issue:

On Technic hub, if observing is started before creating a `Remote()` object, creating the `Remote()` object will fail with `RuntimeError`. Probably because the Bluetooth stack is already scanning and needs to do a different scan to connect to the remote.  To fix this we could make the Bluetooth driver "smart" and stop and restore the observer scanning when connecting to the remote.

laurensvalk:
> [pybricks.common.BLE: return None if no observed data](https://github.com/pybricks/pybricks-micropython/pull/158/commits/8f7ba06ad05a48b5995fe8ae3a58e31b82dd9063)

Nitpick, this gets somewhat confusing:

```python
hub.ble.broadcast(None)
```

```python
data = hub.ble.observe(0)

if data is None:
    # It's not none? 🤔
```

I think we could perhaps do without 8f7ba06.

Since the empty set is used for no data, users can still do `if data` to check for presence of data.

laurensvalk:
Idea: How about merging it up to [20ba223](https://github.com/pybricks/pybricks-micropython/pull/158/commits/20ba22368191bdb4123b63a2fbc48ea2a596fe8e) already and use a follow up PR for the next round of improvements like:
- Detect/handle no data as in 8f7ba06ad05a48b5995fe8ae3a58e31b82dd9063
- Decide what to do with City Hub
- Decide what to do with Move Hub
- Known issue about technic hub combined with Remote

This would be a convenient way to bring it into day-to-day testing and the next beta round.

On the topic of City/Move Hub, even if we decide to have only scanning, we could make it work intuitive enough in the docs and autocomplete similar to the IMU being a superset (inherited) of the simpler Accelerometer of the Move Hub: Only the available methods are be documented/autocompleted.

dlech:
Simpler example where motor on Hub 1 mirrors motor that is manually turned on Hub 2.

Hub 1:

```python
from pybricks.parameters import Port, Color

hub = ThisHub(observe_channels=[2])
motor = Motor(Port.A)

print("radio fw", hub.ble.version())

prev = None

while True:
    data = hub.ble.observe(2)
    
    if data is None:
        hub.light.on(Color.RED)
        motor.track_target(0)
    else:
        hub.light.on(Color.GREEN)

        target = data[0]

        motor.track_target(target)

        if target != prev:
            print(target)
            prev = target

    # broadcasts cannot be sent/received faster than once every 100ms
    wait(100)
```

Hub 2:

```python
from pybricks.parameters import Port, Color

hub = ThisHub(broadcast=2)
motor = Motor(Port.A)

print("radio fw", hub.ble.version())

prev = None

while True:
    target = motor.angle()

    hub.ble.broadcast(target)

    if target != prev:
        print(target)
        prev = target

    # broadcasts cannot be sent/received faster than once every 100ms
    wait(100)
```

laurensvalk:
> I think we could perhaps do without https://github.com/pybricks/pybricks-micropython/commit/8f7ba06ad05a48b5995fe8ae3a58e31b82dd9063.

For what it's worth, I'm fine keeping this in also after thinking about it some more.

So, if there are no technical hurdles that could make other Bluetooth developments more complicated, I vote to merge.

> Move Hub (...) City Hub (...)

Maybe we could start by having broadcast disabled on these?

michaelwareman:
I had a change to play with my Johnny 5 robot and using the broadcast feature. It all started when I wanted to test the robot without first connecting to the PC (Windows 10) The robot did nothing. I thought maybe it was due to the print statements in the code. So, I commented them out and the robot does not do anything. If I uncomment the statements the robot moves its arms. I then got a message that there was a new beta.pybricks.com so I restarted both browser tabs. No change. I then also thought there might be a new beta firmware version. I downloaded the pybricks-code-2.2.0-beta.3.zip file. When I add that to the beta.pybricks.com for firmware using the advanced feature I get this error message:
FirmwareReaderError: missing firmware-base.bin
    at new t (https://beta.pybricks.com/static/js/main.14781b98.js:2:794746)
    at Function.<anonymous> (https://beta.pybricks.com/static/js/main.14781b98.js:2:795612)
    at https://beta.pybricks.com/static/js/main.14781b98.js:2:793287
    at Object.next (https://beta.pybricks.com/static/js/main.14781b98.js:2:793392)
    at a (https://beta.pybricks.com/static/js/main.14781b98.js:2:792105
The code for the two hubs is listed below. Why does the code work when the print statements are active but not when they are commented out? Also, it is real annoying to have to restart the computer every couple of runs because I cannot connect the bluetooth to the one or more hubs. Sometimes the one of the hubs will turn off or not stop the program.
Mike

'''
#Upper hub with arms
from pybricks.hubs import InventorHub
from pybricks.pupdevices import Remote, Motor
from pybricks.parameters import Color, Button, Port, Direction

hub = InventorHub(broadcast_channel=1, observe_channels=[1, 2])
left_arm_motor = Motor(Port.E, Direction.COUNTERCLOCKWISE, [20, 60])
right_arm_motor = Motor(Port.F, Direction.CLOCKWISE, [20, 60])

hub.light.on(Color.YELLOW)

while True:
    left_arm_target = left_arm_motor.angle()
    right_arm_target = right_arm_motor.angle()
    hub.ble.broadcast(left_arm_target, right_arm_target)
    print(left_arm_target, right_arm_target)

    data = hub.ble.observe(2)
    

    if data is not None:
        left_arm_target = data[0]
        right_arm_target = data[1]
        #arms move one at a time
        #left_arm_motor.run_target(75, left_arm_target)
        #right_arm_motor.run_target(75, right_arm_target)
        #arms move together
        left_arm_motor.run_target(75, left_arm_target, wait=False)
        right_arm_motor.run_target(75, right_arm_target, wait=False)

        #print(left_arm_target, right_arm_target)
'''
'''
# Drive Hub (lower hub)
from pybricks.hubs import InventorHub
from pybricks.parameters import Port, Color

hub = InventorHub(broadcast_channel=2, observe_channels=[1])

right_arm_target_angle = 0
right_arm_angle = 0
left_arm_angle = 0
left_arm_target_angle = 0

hub.light.on(Color.GREEN)

while True:
    data = hub.ble.observe(1)

    if data is not None:
        left_arm_angle = data[0]
        right_arm_angle = data[1]

        #move arms to down ititialization position
        if left_arm_angle < 10:
            left_arm_target_angle = 300

        if right_arm_angle < 10:
            right_arm_target_angle = 300

    print(left_arm_angle, right_arm_angle)
    hub.ble.broadcast(left_arm_target_angle, right_arm_target_angle)
'''

dlech:
>  I downloaded the pybricks-code-2.2.0-beta.3.zip file

This is not a firmware file which is why it fails to flash to the hub. The current beta has support for BLE broadcasting so you no longer need a special firmware.zip file, just flash using https://beta.pybricks.com. 

In the future, if you need support the best place to ask is https://github.com/orgs/pybricks/discussions.

vardayzsolt:
I tested the new example programs included in the documentation of hub.ble.broadcast / observe  functions included in latest 2.2.0-beta.4 release. I'm really happy to see it finally implemented, thank you for that!
While connected to Pybricks API, everything goes well, just as expected. But on Technic Hub there is a consistent problem of not broadcasting any data when the code is run directly from the hub (after disconnecting from pc). The other Techic Hub acting as receiver keeps receiving ble signal when disconnected from pc and code run from directly hub. When code is adapted to Inventor Hub and started directly from Hub everything goes fine. So the problem occurs only on Tecyhnic Hub and only in broadcasting mode. I tested on several Technic Hubs as broadcaster and various hubs as receiver, and the problem is 100% persistent.
Another observation: there is much less latency in receiver Technic Hub motor folowing the broadcaster Hub's motor movements when the broadcaster is Inventor Hub (it is close to 100ms as observed). It seems as Technic Hub has much more than 100ms intervals (rather 800-1000ms as observed) between data transmission session.

laurensvalk:
Thanks @vardayzsolt! Let's follow up in https://github.com/pybricks/support/issues/1086

PR #159, opened 2023-04-18
pbio/lego_port: Introduce port mode specification.

laurensvalk:
**What this is for**
This is a start of a collection of ideas and notes to work towards a unified lego port model across the platforms.

Perhaps slightly ambitious (but it appears possible), but if it works it could help us simplify support across platforms, And also as:
- Support custom UART devices https://github.com/pybricks/support/issues/220
- Support custom I2C devices https://github.com/orgs/pybricks/discussions/1047
- Support custom GPIO devices https://github.com/pybricks/support/issues/1027
- All of the above on even on embedded EV3 without linux to provide those handles

It should also get us closer to sharing the nxtdevices MicroPython module across NXT/EV3, and also get rid of the intermediate layers like `pb_device_stm32`, `pb_device_ev3`, etc.

**Differences from ev3dev model**
There are two main differences:
- Distinguish between functionality modes and (a pair of) physical modes on P1P2 and P5P6
- The port mode is *not* auto-selected. Instead, the mode is set to a default on boot and can be changed by the user (usually by a MicroPython class init). Within the selected mode (usually the default LEGO mode), the device type *is* auto-identified, and the underlying P1P2 and P5P6 modes *are* automatically changed accordingly.

### DIFF
--- a/lib/pbio/include/pbio/lego_port.h
+++ b/lib/pbio/include/pbio/lego_port.h
@@ -1,0 +1,296 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2023 The Pybricks Authors
+
+/**
+ * @addtogroup LegoPort pbio/lego_port: I/O port control for LEGO 6-wire ports.
+ * @{
+ */
+
+#ifndef _PBIO_LEGO_PORT_H_
+#define _PBIO_LEGO_PORT_H_
+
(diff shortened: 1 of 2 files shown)


laurensvalk:
This project will be on hold for a while. Closing now, and reopening when we pick this up.

PR #161, opened 2023-04-28
pbio/motor_process: catch up timer if behind

dlech:
This adds an extra check to make sure the motor poll timer hasn't fallen too far behind.

Before this change, if calling the motor process was blocked for too long, it would cause the motor contiki process to never yield until the wall clock caught up. This would also cause motor updates to be called with a time delta of 0 which probably caused unexpected conditions in some of the calculations.

We can avoid this by checking to see if we have a condition that would trigger the unwanted behavior and reset the timer to a larger interval. This could cause a hiccup in the observer calculations but is a significant improvement over the alternative.

Fixes: https://github.com/pybricks/support/issues/1035

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,9 @@
 
 ### Fixed
 - Fixed stdin containing `0x06` command byte ([support#1052]).
+- Fixed motor process causing delays on ev3dev ([support#1035]).
 
+[support#1035]: https://github.com/pybricks/support/issues/1035
 [support#1052]: https://github.com/pybricks/support/issues/1052
 
 ## [3.3.0b4] - 2023-04-21
--- a/lib/pbio/src/motor_process.c
+++ b/lib/pbio/src/motor_process.c
@@ -37,8 +37,17 @@ PROCESS_THREAD(pbio_motor_process, ev, data) {
         // Update servos
         pbio_servo_update_all();
 
-        // Reset timer to wait for next update
+        // Reset timer to wait for next update. Using etimer_reset() instead
+        // of etimer_restart() makes average update period closer to the expected
+        // PBIO_CONFIG_CONTROL_LOOP_TIME_MS when occasional delays occur.
         etimer_reset(&timer);
+
+        // If we have fallen too far behind though, jump ahead to catch up.
+        // Otherwise, this process will run in a tight loop and not yield
+        // until the timer catches up.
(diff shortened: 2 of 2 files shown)


laurensvalk:
Thanks for looking into this!

I'll try running some data logs next week.

laurensvalk:
Before PR:

image

After:

image

So this probably still needs a closer look.

To reproduce: 

Open `single_motor.py` and launch the `Motor Test EV3` config. Then just wait and you'll get a graph:

image

dlech:
Tried a different approach which makes better graphs.

image

Still occasionally getting some really ugly ones though (maybe a separate issue due to not realtime Linux though?).

image

laurensvalk:
Thanks for the update. I built something with EV3 recently, so I'll be having a look at this as I program it.

> maybe a separate issue due to not realtime Linux though?

The loop time (bottom graph) appears to suggest that the delayed loops are still part of it.

image

It would be possible to make it less dependent on loop time, but that would be a fairly big rewrite. Two easier options:
- Discretize observer model at 10 ms instead of 5 ms (and run control loop at 10 ms). Maybe we have more breathing space that way?
- Disable use of observer on EV3. The speed measurements here are quite good.

laurensvalk:
> Discretize observer model at 10 ms instead of 5 ms (and run control loop at 10 ms). Maybe we have more breathing space that way?

Took matters a bit further and tried 20 ms, with updated model and set the loop thread to 3.3 ms.

image

Longer test, `single_motor_4moves.py`:

image

It is a bit coarse in the graph but it looks reasonable on the EV3. Also, `pybricks-micropython` starts and stops way faster and the beep issue is fixed.

I'll convert it to an alternate PR for comparison.

laurensvalk:
As a minor optimization we could also cache the `motor_driver` state so we don't coast when already coasting, which should help with startup.

laurensvalk:
Alternative in https://github.com/pybricks/pybricks-micropython/compare/master...motor-poll-fix-2

dlech:
Nice work.

I think we still might want to include the change from this PR just for safety to be certain we avoid the bad state in the motor poll loop.

laurensvalk:
Yes, sounds good. If you want to merge yours first, I'll update and merge mine after.

Based on what you are seeing in the other PR, feel free to pick your first or second approach?

PR #162, opened 2023-05-02
Filter RSSI, add signal_strength method

laurensvalk:
Proposed improvement to #158.

The RSSI value is quite jittery. This filters the signal to be smoother. It takes the time between samples into account, so that there is no filter delay to register the first valid RSSI value.

The `rssi` value gets its own `signal_strength` method. Rationale:
- Unpacking two tuples to get at the data is a bit much, this reduces it one level
- The rssi seems like a fairly advanced feature that most applications won't need. Detecting data is just as good to determine presence of broadcasted signals.
- Due to filtering, there is no direct correspondence to a particular data set, so coupling them with the data is not very useful in this case.
- On naming: `Received Signal Strength Indicator` is a bit long so I went with `signal_strength`. However, since it is a separate method for more advanced users, I'm not opposed to calling it just `rssi`.
- Could be used for signal strength of other BLE devices if we ever have those; is not restricted to pybricks broadcasting


dlech:
Sounds good, thanks!

PR #163, opened 2023-05-02
bricks: Checkout micropython-lib.

laurensvalk:
This appears to be necessary to build frozen modules now.

Is this the right way to do it?

Fixes:
```
CC ../../micropython/py/showbc.c
CC ../../micropython/py/repl.c
CC ../../micropython/py/smallint.c
CC ../../micropython/py/frozenmod.c
-e Error: micropython-lib submodule is not initialized. Run 'make submodules'  <-------
CC ../../micropython/extmod/machine_bitstream.c
CC ../../micropython/extmod/machine_i2c.c
```

### DIFF
--- a/bricks/_common/arm_none_eabi.mk
+++ b/bricks/_common/arm_none_eabi.mk
@@ -36,6 +36,13 @@ ifeq ("$(wildcard $(PBTOP)/micropython/README.md)","")
 $(error failed)
 endif
 endif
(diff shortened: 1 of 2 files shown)


dlech:
yes, looks right.

We should also modify `.github/workflows/build.yaml` to `git submodule update --init --depth 0 micropython-lib` to speed up CI. We already have this for other submodules on the virtual hub but will need to add it to other hubs too now.

Although if the timestamps in the logs can be believed, maybe this is not even necessary any more?

PR #164, opened 2023-05-02
pbio/platform/ev3dev_stretch: Set control loop time to 10 ms.

dlech:
This reduces the load on the rest of the program.

See: https://github.com/pybricks/support/issues/1035


laurensvalk:
I'm merging in 10 ms as a starting point. We can scale this down further to 20 ms if needed.

PR #174, opened 2023-06-02
pbdrv/legodev: Refactor LEGO device abstraction.

laurensvalk:
This is a draft to revise how we deal with lego devices at the low and high level.

This large change is based on findings from the past years. Instead of trying to have an all-encompassing `iodev` abstraction (which ended up being hard-coded for LEGO devices), we have an interface specifically for LEGO devices, separate from port I/O.

In the future, this may play the role of LEGO-mode on a port (alongside raw UART/I2C modes, etc.).

The interface is platform agnostic, but there will be platform-specific implementations for:
- Powered Up (done)
- Virtual Hub (done)
- ev3dev (in progress)
- tests (todo)
- nxt (todo)

The following sums it up quite well -- this pull request goes a long way to get rid of technical debt like this:

```c
// Delete me: refactor getter only still used for passive external light.
pbio_iodev_t *pb_pup_device_get_device(pbio_port_id_t port, pbio_iodev_type_id_t valid_id) {

    pbio_iodev_t *iodev;
    pbio_error_t err;
    while ((err = pbdrv_ioport_get_iodev(port, &iodev)) == PBIO_ERROR_AGAIN) {
        mp_hal_delay_ms(50);
    }
    pb_assert(err);

    // Verify the ID or always allow generic LUMP device
    if (iodev->info->type_id != valid_id && valid_id != PBIO_IODEV_TYPE_ID_LUMP_UART) {
        pb_assert(PBIO_ERROR_NO_DEV);
    }

    return iodev;
}

// Revisit: get rid of this abstraction
void pb_pup_device_setup_motor(pbio_port_id_t port, bool is_servo) {
    // HACK: Built-in motors on BOOST Move hub do not have I/O ports associated
    // with them.
    #if PYBRICKS_HUB_MOVEHUB
    if (port == PBIO_PORT_ID_A || port == PBIO_PORT_ID_B) {
        return;
    }
    #endif

    pb_module_tools_assert_blocking();

    // Get the iodevice
    pbio_iodev_t *iodev;
    pbio_error_t err;

    // Set up device
    while ((err = pbdrv_ioport_get_iodev(port, &iodev)) == PBIO_ERROR_AGAIN) {
        mp_hal_delay_ms(50);
    }
    pb_assert(err);

    // Only motors are allowed.
    if (!PBIO_IODEV_IS_DC_OUTPUT(iodev)) {
        pb_assert(PBIO_ERROR_NO_DEV);
    }

    // If it's a DC motor, no further setup is needed.
    if (!PBIO_IODEV_IS_FEEDBACK_MOTOR(iodev)) {
        return;
    }

    // FIXME: there is no guarantee that the iodev is a uartdev
    #if !PYBRICKS_HUB_VIRTUALHUB

    // Choose mode based on device capabilities.
    uint8_t mode_id = PBIO_IODEV_IS_ABS_MOTOR(iodev) ?
        PBIO_IODEV_MODE_PUP_ABS_MOTOR__CALIB:
        PBIO_IODEV_MODE_PUP_REL_MOTOR__POS;

    // Activate mode.
    pb_pupdevices_obj_base_t device = {
        .iodev = iodev,
    };
    pb_pupdevices_get_data_blocking(&device, mode_id);

    #endif // !PYBRICKS_HUB_VIRTUALHUB
}

// REVISIT: Drop pb_device abstraction layer
void pb_device_setup_motor(pbio_port_id_t port, bool is_servo) {
    pb_pup_device_setup_motor(port, is_servo);
}
```

It also adds an option to exclude some `pup_uart` functionality to reduce build size on Move Hub by about 500 bytes.


laurensvalk:
Getting somewhere on this one now. Remaining tasks:

- [ ] finish ev3dev sensors
- [ ] fix uartdev tests not giving expected results

laurensvalk:
> There was one spurious bug earlier today where a motor would not sync up or start, but I can't seem to reproduce it now.

I am occasionally able to reproduce it. It appears that setting the default mode (immediately after syncing up) does not always succeed, and we currently don't retry.

So maybe we should wait briefly before setting the default mode, or retry if a mode switch hasn't succeeded.

dlech:
We should put a logic analyzer on it to see what is actually going over the wire.

laurensvalk:
Nice.

I have an idea that would take care of this issue and also make the general mode setting and data setting code easier to follow and more robust.

Basically, we could modify this existing loop:

```c
while (data->status == PBDRV_LEGODEV_PUP_UART_STATUS_DATA) { 
     // await send keep alive on EV3_UART_IO_TIMEOUT
}
```

to:

```c
while (data->status == PBDRV_LEGODEV_PUP_UART_STATUS_DATA) { 

     // wait until any of following conditions, then

     // await send keep alive if EV3_UART_IO_TIMEOUT

     // await set mode if requested

     // await set data if requested
}

```

This avoids clashes between setting stuff and sending keep alive (theoretically possible since https://github.com/pybricks/pybricks-micropython/pull/169 although never happened) by keeping all sending in one place.

But it also lets us simplify the main process loop:

```c

    while (true) {
        for (i = 0; i < PBDRV_CONFIG_LEGODEV_PUP_UART_NUM_DEV; i++) {
            port_data = &dev_data[i];

            pbdrv_legodev_pup_uart_update(port_data);

            if (port_data->status == PBDRV_LEGODEV_PUP_UART_STATUS_DATA) {
                pbdrv_legodev_pup_uart_receive_data(port_data);
            }

            // CAN DROP: this was a hack to handle awaited mode/data sets
            pbdrv_legodev_pup_uart_handle_write_end(port_data);

            // CAN DROP: handled simpler above
            if (port_data->status == PBDRV_LEGODEV_PUP_UART_STATUS_DATA) {
                pbdrv_legodev_pup_uart_handle_data_set_start(port_data);
            }
        }
        PROCESS_WAIT_EVENT();
    }

```

laurensvalk:
Above issue is fixed, and did some more code consolidation to save on size. This PR saves us another 700 bytes  so we have a whole 1KB to spare again which should cover us on bug fixes for a while. :rocket:

PR #177, opened 2023-06-06
pybricks.tools: Add read_one() function.

dlech:
This adds a new non-blocking function to read one byte from stdin. This is needed since there is no other way to read from stdin on the Move hub. It is also useful on other hubs as it eliminates the need to use poll() to avoid blocking reads.

Fixes: https://github.com/pybricks/support/issues/1102

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Added
 - Added `'modes'` entry to the dictionary returned by `PUPDevice.info()`. It
   is a tuple of `(name, num_values, data_type)` tuples for each available mode.
+- Added `pybricks.tools.read_input_byte()` function ([support#1102]).
 
 ### Changed
 - Changed internal drivers for LEGO devices (motors and sensors) on all platforms.
@@ -17,6 +18,7 @@
 
 [support#1095]: https://github.com/pybricks/support/issues/1095
 [support#1098]: https://github.com/pybricks/support/issues/1098
+[support#1102]: https://github.com/pybricks/support/issues/1102
 
 ## [3.3.0b6] - 2023-06-02
 
--- a/bricks/ev3dev/Makefile
+++ b/bricks/ev3dev/Makefile
@@ -311,6 +311,9 @@ test: $(BUILD)/$(PROG) $(TOP)/tests/run-tests.py $(GRX_TEST_PLUGIN_LIB)
 		GRX_PLUGIN_PATH=$(realpath $(BUILD)) GRX_DRIVER=test \
 		./run-tests.py
 
+repl: $(BUILD)/$(PROG) $(GRX_TEST_PLUGIN_LIB) modules/core.py
+	GRX_PLUGIN_PATH=$(realpath $(BUILD)) GRX_DRIVER=test $< -i modules/core.py
+
 # Value of configure's --host= option (required for cross-compilation).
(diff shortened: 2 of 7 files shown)


laurensvalk:
Nice!

Looking at a glance, I wonder if there’s a more descriptive name. What is it reading one of? What is it reading from? Maybe get_input, get_byte, or get_input_char?

dlech:
> What is it reading one of? What is it reading from?

Then the name should be `read_one_byte_from_stdin_nonblock()`, but that seems rather long. 

It is *not* reading a character because UTF-8 characters/code points can be more than one byte.

dlech:
Here is an updated version of the sample from https://pybricks.com/projects/tutorials/wireless/hub-to-device/pc-keyboard/ that can handle control keys and non-ascii characters.



```python
from pybricks.tools import read_one, wait

print("activate the terminal and press a key...")

while True:
    b = read_one()

    if b is None:
        wait(10)
        continue

    if b == 9:
        key = "TAB"
    elif b == 13:
        key = "ENTER"
    elif b == 27:
        key = "ESC"

        # might be a terminal escape code
        code = read_one()

        if code is None:
            pass
        elif code == ord('['):
            # start of control sequence
            final = read_one()

            params = []
            inter = []

            while final >= 0x20 and final < 0x40:
                if final < 0x30:
                    params.append(final)
                else:
                    inter.append(final)

                final = read_one()

            if final == ord('A'):
                key = "UP"
            elif final == ord('B'):
                key = "DOWN"
            elif final == ord('C'):
                key = "RIGHT"
            elif final == ord('D'):
                key = "LEFT"
            elif final == ord('F'):
                key = "END"
            elif final == ord('H'):
                key = "HOME"
            elif final == ord('~'):
                vt = "".join(chr(x) for x in inter)

                if vt == "2":
                    key = "INSERT"
                elif vt == "3":
                    key = "DELETE"
                elif vt == "5":
                    key = "PAGE UP"
                elif vt == "6":
                    key = "PAGE DOWN"
                else:
                    key = "?VT {}".format(vt)
            else:
                key = "?CSI {} {} {}".format(params, inter, final)
    
    elif b == 32:
        key = "SPACE"
    elif b > 32 and b < 127:
        key = chr(b)
    elif b == 127:
        key = "BACKSPACE"
    elif (b & 0xC0) == 0xC0:
        # multi-byte UTF-8
        if (b & 0x20) == 0:
            n = 1
        elif (b & 0x10) == 0:
            n = 2
        elif (b & 0x08) == 0:
            n = 3

        cp = [b]

        for _ in range(n):
            cp.append(read_one())

        key = str(bytes(cp), "utf8")
    else:
        key = "unknown - {}".format(b)

    print("You pressed:", key)

```

dlech:
Bare minimum example:

```python
from pybricks.tools import read_one, wait

print("activate the terminal and press a key...")

while True:
    b = read_one()

    if b is None:
        wait(10)
        continue

    print("You pressed:", b)

```

laurensvalk:
> > What is it reading one of? What is it reading from?
> 
> Then the name should be `read_one_byte_from_stdin_nonblock()`, but that seems rather long. 
> 
> It is *not* reading a character because UTF-8 characters/code points can be more than one byte.

Right, so hence the suggestion for input or byte. Since byte is singular, one could be omitted.

Maybe `read_input_byte`.

PR #178, opened 2023-06-28
pybricks.tools: Add hub_menu function.

laurensvalk:
This implements the first approach discussed in https://github.com/pybricks/support/issues/1064.

```python
from pybricks.tools import hub_menu

# Choose a letter.
selected = hub_menu("F", "D", "Z")

# Based on the selection, run a program.
if selected == "F":
    import fly_mission
elif selected == "D":
    import drive_mission
elif selected == "Z":
    import zigzag
```

This is really only applicable to systems with the light matrix, so maybe it belongs there? (Only that doesn't have buttons, so maybe not).

### DIFF
--- a/pybricks/tools/pb_module_tools.c
+++ b/pybricks/tools/pb_module_tools.c
@@ -13,6 +13,8 @@
 #include "py/stream.h"
 
 #include <pbio/int_math.h>
+#include <pbsys/light.h>
+#include <pbsys/program_stop.h>
 
 #include <pybricks/parameters.h>
 #include <pybricks/common.h>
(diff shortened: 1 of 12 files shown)


pybricks/tools/pb_module_tools.c
```diff
@@ -143,10 +144,88 @@ STATIC mp_obj_t pb_module_tools_run_task(size_t n_args, const mp_obj_t *pos_args
...
+        pb_type_LightMatrix_display_char(pbsys_hub_light_matrix, symbol);
+    } else {
+        pb_type_LightMatrix_display_number(pbsys_hub_light_matrix, symbol);
+    }
+}
+
+STATIC pbio_button_flags_t pb_module_tools_hub_menu_wait_for_press(bool press) {
+    pbio_error_t err;
+    pbio_button_flags_t btn;
+    while ((err = pbio_button_is_pressed(&btn)) == PBIO_SUCCESS && ((bool)btn) == !press) {
```
dlech, line 162: Raise error if event loop is running since this is a blocking function?

pybricks/tools/pb_module_tools.c
```diff
@@ -143,10 +144,88 @@ STATIC mp_obj_t pb_module_tools_run_task(size_t n_args, const mp_obj_t *pos_args
 }
 STATIC MP_DEFINE_CONST_FUN_OBJ_KW(pb_module_tools_run_task_obj, 1, pb_module_tools_run_task);
 
+#if PYBRICKS_PY_COMMON && PYBRICKS_PY_COMMON_LIGHT_MATRIX
```
dlech, line 147:
```suggestion
#if PYBRICKS_PY_TOOLS_HUB_MENU
```

If given infinite time, I would like to use Kconfig to express complex relations of config options. But for now, I think it is best to just keep things simple and make unique config options for each feature use this sort of pattern.

```c
#if THIS_FEATURE
#if !DEPENDENT_FEATURE
#error "THIS_FEATURE requires DEPENDENT_FEATURE"
#endif

...

#endif // THIS_FEATURE
```

pybricks/tools/pb_module_tools.c
```diff
@@ -143,10 +144,88 @@ STATIC mp_obj_t pb_module_tools_run_task(size_t n_args, const mp_obj_t *pos_args
...
+    // Validate arguments by displaying all of them, ending with the first.
+    // This ensures we fail right away instead of midway through the menu. It
+    // happens so fast that there isn't a time penalty for this.
+    for (int i = n_args - 1; i >= 0; i--) {
+        pb_module_tools_hub_menu_display_symbol(args[i]);
+    }
+
+    // Disable stop button and cache original setting to restore later.
+    pbio_button_flags_t stop_button = pbsys_program_stop_get_buttons();
+    pbsys_program_stop_set_buttons(0);
```
dlech, line 187: The rest of the code should probably be wrapped in `nlr_push()` (eventually to be replaced by https://github.com/micropython/micropython/pull/11445) so that the stop button gets restored no matter what.

pybricks/tools/pb_module_tools.c
```diff
@@ -143,10 +145,107 @@ STATIC mp_obj_t pb_module_tools_run_task(size_t n_args, const mp_obj_t *pos_args
...
+        // Restore stop button setting prior to starting menu.
+        pbsys_program_stop_set_buttons(stop_button);
+
+        // Complete and return selected object.
+        nlr_pop();
+        return args[selection];
+    } else {
+        pbsys_program_stop_set_buttons(stop_button);
+        nlr_jump(nlr.ret_val);
+        return mp_const_none;
```
dlech, line 235: This is not reachable. We can use `MP_UNREACHABLE` instead if the compiler doesn't respect the `noreturn` attribute on `nlr_jump`.

PR #179, opened 2023-07-04
pbio/drv/legodev: Replace unplug reset by watchdog.

laurensvalk:
This keeps the device connection manager in control even if the UART device and PT threads get in a confused state.

Fixes https://github.com/pybricks/support/issues/1137

### DIFF
--- a/lib/pbio/drv/legodev/legodev_pup.c
+++ b/lib/pbio/drv/legodev/legodev_pup.c
@@ -34,6 +34,10 @@
 /** The number of consecutive repeated detections needed for an affirmative ID. */
 #define AFFIRMATIVE_MATCH_COUNT 20
 
+/** If the UART device did not call back to this driver after this time, assume
+    it is disconnected and restart device detection. */
+#define UART_WATCHDOG_TIMEOUT 1100
+
 typedef enum {
     DEV_ID_GROUP_GND,
     DEV_ID_GROUP_VCC,
@@ -49,6 +53,7 @@ typedef struct {
     uint8_t gpio_value;
     uint8_t prev_gpio_value;
     uint8_t dev_id_match_count;
+    struct timer watchdog;
 } dcm_data_t;
 
 typedef struct {
@@ -325,6 +330,13 @@ PROCESS_THREAD(pbio_legodev_pup_process, ev, data) {
             for (int i = 0; i < PBDRV_CONFIG_LEGODEV_PUP_NUM_EXT_DEV; i++) {
                 dev = &ext_devs[i];
 
+                // If UART device was connected but not reporting data, restart detection.
+                if (dev->connected_type_id == PBDRV_LEGODEV_TYPE_ID_LPF2_UNKNOWN_UART &&
+                    dev->connected_type_id == dev->prev_type_id && timer_expired(&dev->dcm.watchdog)) {
+                    dev->connected_type_id = PBDRV_LEGODEV_TYPE_ID_NONE;
+                    dev->dcm.dev_id_match_count = 0;
+                }
+
                 // Keep running device detection unless UART sensor is attached.
                 if (dev->connected_type_id != PBDRV_LEGODEV_TYPE_ID_LPF2_UNKNOWN_UART) {
                     poll_dcm(dev);
@@ -383,6 +395,7 @@ void pbdrv_legodev_init(void) {
         const pbdrv_ioport_pup_port_platform_data_t *port_data = &pbdrv_ioport_pup_platform_data.ports[legodev_data->ioport_index];
         legodev->ext_dev->pdata = legodev_data;
         legodev->ext_dev->pins = &port_data->pins;
+        timer_set(&legodev->ext_dev->dcm.watchdog, UART_WATCHDOG_TIMEOUT);
 
         // Initialize uart device manager.
         pbio_dcmotor_t *dcmotor;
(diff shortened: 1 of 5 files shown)


lib/pbio/drv/legodev/legodev_pup.h
```diff
@@ -33,11 +33,11 @@ extern const pbdrv_legodev_pup_int_platform_data_t
 extern const pbdrv_legodev_pup_ext_platform_data_t
     pbdrv_legodev_pup_ext_platform_data[PBDRV_CONFIG_LEGODEV_PUP_NUM_EXT_DEV];
 
-void pbdrv_legodev_pup_reset_device_detection(pbdrv_legodev_pup_uart_dev_t *dev);
+void pbdrv_legodev_pup_reset_watchdog(pbdrv_legodev_pup_uart_dev_t *uartdev);
 
 #else
 
-static inline void pbdrv_legodev_pup_reset_device_detection(pbdrv_legodev_pup_uart_dev_t *dev) {
+void pbdrv_legodev_pup_reset_watchdog(pbdrv_legodev_pup_uart_dev_t *uartdev) {
```
dlech, line 40: Should not remove `static inline`, otherwise we can end up with multiple definitions of the same symbol.

lib/pbio/drv/legodev/legodev_pup.c
```diff
@@ -394,25 +407,28 @@ void pbdrv_legodev_init(void) {
...
-        }
-        if (dev->ext_dev->uart_dev == uartdev) {
-            // On match reset device connection manager state.
-            dev->ext_dev->connected_type_id = PBDRV_LEGODEV_TYPE_ID_NONE;
-            dev->ext_dev->dcm.dev_id_match_count = 0;
-            return;
-        }
+void pbdrv_legodev_pup_reset_watchdog(pbdrv_legodev_pup_uart_dev_t *uartdev) {
+    pbdrv_legodev_dev_t *dev = pbdrv_legodev_get_parent_of(uartdev);
+    if (dev == NULL || dev->is_internal) {
```
dlech, line 428: Isn't the `is_internal` check redundant?

Is it even possible to have an "internal" uartdev?

lib/pbio/drv/legodev/legodev_pup_uart.c
```diff
@@ -726,6 +729,9 @@ static PT_THREAD(pbdrv_legodev_pup_uart_update(pbdrv_legodev_pup_uart_dev_t * da
     uint8_t speed_payload[4];
     pbio_set_uint32_le(speed_payload, EV3_UART_SPEED_LPF2);
     ev3_uart_prepare_tx_msg(data, LUMP_MSG_TYPE_CMD, LUMP_CMD_SPEED, speed_payload, sizeof(speed_payload));
+
+    pbdrv_uart_flush(data->uart);
```
dlech, line 733: Why do we need the flush here since we are already flushing on line 741?

lib/pbio/drv/legodev/legodev_pup.c
```diff
@@ -325,6 +330,13 @@ PROCESS_THREAD(pbio_legodev_pup_process, ev, data) {
             for (int i = 0; i < PBDRV_CONFIG_LEGODEV_PUP_NUM_EXT_DEV; i++) {
                 dev = &ext_devs[i];
 
+                // If UART device was connected but not reporting data, restart detection.
+                if (dev->connected_type_id == PBDRV_LEGODEV_TYPE_ID_LPF2_UNKNOWN_UART &&
+                    dev->connected_type_id == dev->prev_type_id && timer_expired(&dev->dcm.watchdog)) {
+                    dev->connected_type_id = PBDRV_LEGODEV_TYPE_ID_NONE;
+                    dev->dcm.dev_id_match_count = 0;
```
dlech, line 337: Modifying the DCM state outside of the DCM module feels like spaghetti code. IMHO it would be better if UART sensor code could be reused independent of a DCM and it would be easier to understand the DCM if all state is managed in a single file.

dlech (commented):
> This keeps the device connection manager in control even if the UART
device and PT threads get in a confused state.

It would be nice to know how it got in the confused state to begin with. Do you have logic analyzer data to show what is going on here?

lib/pbio/drv/legodev/legodev_pup_uart.c
laurensvalk: Without this, it always syncs up at 2400 baud.

lib/pbio/drv/legodev/legodev_pup.c
laurensvalk: That's what this commit is working towards. Note that this piece of code was already here, just moved.

laurensvalk: > Modifying the DCM state outside of the DCM module feels like spaghetti code.

So this reduces it. It used to be modified from outside the DCM, and not anymore after this change. This is why it could get stuck before and not anymore now.

laurensvalk: I've started gathering some ideas to further separate them in https://github.com/pybricks/support/issues/1140

laurensvalk:
Thanks for the review. It would be nice to get the sensors in a working state for now, while we can consider further separation as in e.g. https://github.com/pybricks/support/issues/1140 as a follow up.

lib/pbio/drv/legodev/legodev_pup_uart.c
dlech: I still don't understand how this could make a difference since nothing is read between this flush and the next flush.

laurensvalk: That makes two of us :smile:

laurensvalk: I did notice that on boot, it does always work on high baud the first time on Prime Hub, but on plugging in the second time, it usually goes at `2400` baud, as if something (else) isn't quite flushed out.

The easiest way to see is with the SPIKE Color Sensor, whose light will either turn on nearly instantaneously or otherwise take a few seconds to dump all its data.

PR #182, opened 2023-07-10
UART sensors cleanup

laurensvalk:
This implements a start to https://github.com/pybricks/support/issues/1140:

- This is a start to separating the device connection manager and the uart device code.
- Device synchronization is more likely to work in one go
- Disconnecting is handled more quickly because we can drop various timeouts
- Reduces code size on Move Hub by about 350 bytes

I have intentionally avoided much change to the `dcm` and `uartdev` code in this commit to make the change easier to process. We could do a few more cleanups though:

- Fix mixed use of `uartdev`, `data`, `port_data` name for the main data structure
- Fix names of the various protothreads: `data->pt` is not the same as `data_pt`, which can make it harder to follow.
- Now that the dcm and uartdev run separately from start to end, each can be a proper protothread that reads from start to finish, instead of needing an external loop with a function call that causes it to run to the next yield. This will work the same but it could make the code easier to follow, especially since we omit the `PT_SCHEDULE` macro in a few places when we don't need the return value.

### DIFF
--- a/lib/pbio/drv/legodev/legodev_pup.c
+++ b/lib/pbio/drv/legodev/legodev_pup.c
@@ -341,6 +341,12 @@ static PT_THREAD(pbdrv_legodev_pup_thread(ext_dev_t * dev)) {
         dev->dcm.dev_id_match_count = 0;
 
         // Run the device detection manager until a UART device is detected.
+        // This process continously checks which type of passive device is
+        // attached by setting and reading the ID1 and ID2 pins in a predefined
+        // sequence at 2ms intervals. Once a UART device is detected, this
+        // detection process stops because it would interfere with the UART
+        // data. Once the UART process completes (i.e. the device disconnects),
+        // device detection will resume here.
         etimer_set(&dev->timer, 2);
         PT_INIT(&dev->child);
         PT_WAIT_UNTIL(&dev->pt, ({
@@ -353,7 +359,8 @@ static PT_THREAD(pbdrv_legodev_pup_thread(ext_dev_t * dev)) {
             dev->dcm.connected_type_id == PBDRV_LEGODEV_TYPE_ID_LPF2_UNKNOWN_UART;
         }));
 
-        // UART device detected, so run that protocol until it disconnects.
+        // UART device detected, so hand off control to that protocol until it
+        // disconnects, as observed by the UART process not getting valid data.
         legodev_pup_enable_uart(dev->pins);
         PT_SPAWN(&dev->pt, &dev->child, pbdrv_legodev_pup_uart_thread(&dev->child, dev->uart_dev));
     }
--- a/lib/pbio/drv/legodev/legodev_pup.c
+++ b/lib/pbio/drv/legodev/legodev_pup.c
@@ -44,25 +44,37 @@ typedef enum {
 
 // Device connection manager state for each port
 typedef struct {
-    struct pt pt;
-    dev_id_group_t dev_id1_group;
+    /** Most recent one-off device ID candidate. */
     pbdrv_legodev_type_id_t type_id;
+    /** Previous one-off device ID candidate. */
     pbdrv_legodev_type_id_t prev_type_id;
+    /** Most recent device ID with enough consecutive detections. */
+    pbdrv_legodev_type_id_t connected_type_id;
+    /** Previous device ID with enough consecutive detections. */
+    pbdrv_legodev_type_id_t prev_connected_type_id;
+    /** Number of consecutive detections of the same device ID. */
+    uint8_t dev_id_match_count;
+    // Intermediate state values to preserve in between async calls.
+    dev_id_group_t dev_id1_group;
     uint8_t gpio_value;
     uint8_t prev_gpio_value;
-    uint8_t dev_id_match_count;
 } dcm_data_t;
 
 typedef struct {
+    /**
+     * Main protothread for the legodev. Each external legodev port has one,
+     * all driven by a single process that runs until the hub shuts down.
+     */
+    struct pt pt;
+    /**
+     * Child protothread used for the device connection manager and the uart
+     * device process. These don't run at the same time, so we can reuse it. */
+    struct pt child;
     const pbdrv_ioport_pup_pins_t *pins;
     const pbdrv_legodev_pup_ext_platform_data_t *pdata;
     pbdrv_legodev_pup_uart_dev_t *uart_dev;
-    struct pt uart_dev_pt;
     dcm_data_t dcm;
-    struct pt pt;
     struct etimer timer;
-    pbdrv_legodev_type_id_t connected_type_id;
-    pbdrv_legodev_type_id_t prev_type_id;
     pbio_angle_t angle;
 } ext_dev_t;
 
@@ -99,7 +111,7 @@ static void legodev_pup_enable_uart(const pbdrv_ioport_pup_pins_t *pins) {
 // on the port to see when devices are connected or disconnected.
 // It is expected for there to be a 2ms delay between calls to this function.
 PT_THREAD(poll_dcm(ext_dev_t * dev)) {
-    struct pt *pt = &dev->dcm.pt;
+    struct pt *pt = &dev->child;
     dcm_data_t *data = &dev->dcm;
     const pbdrv_ioport_pup_pins_t *pins = dev->pins;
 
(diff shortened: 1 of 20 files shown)


laurensvalk:
> We could save code size on move hub and have faster start up on SPIKE hubs if we make this a configurable option and only enable it on the hubs that need it.

_Originally posted by @dlech in https://github.com/pybricks/pybricks-micropython/pull/175#discussion_r1258282647_

lib/pbio/drv/legodev/legodev_pup.c
```diff
@@ -384,6 +376,35 @@ struct _pbdrv_legodev_dev_t {
...
+PROCESS_THREAD(pbio_legodev_pup_process, ev, data) {
+    #if PBDRV_CONFIG_LEGODEV_PUP_POWER_CYCLE_PORTS
+    static struct etimer timer;
+    #endif
+
+    static int i;
+
+    PROCESS_BEGIN();
+
+    #if PBDRV_CONFIG_LEGODEV_PUP_POWER_CYCLE_PORTS
```
dlech, line 388: I've been adding `QUIRK` to config options for hacks like this.

```suggestion
    #if PBDRV_CONFIG_LEGODEV_PUP_QUIRK_POWER_CYCLE_PORTS
```

lib/pbio/drv/legodev/legodev_pup.c
```diff
@@ -384,6 +376,35 @@ struct _pbdrv_legodev_dev_t {
 
 static pbdrv_legodev_dev_t devs[PBDRV_CONFIG_LEGODEV_PUP_NUM_INT_DEV + PBDRV_CONFIG_LEGODEV_PUP_NUM_EXT_DEV];
 
+PROCESS_THREAD(pbio_legodev_pup_process, ev, data) {
+    #if PBDRV_CONFIG_LEGODEV_PUP_POWER_CYCLE_PORTS
+    static struct etimer timer;
+    #endif
```
dlech, line 382: This can be moved into the other `#if` to reduce noise.

lib/pbio/drv/legodev/legodev_pup_uart.h
```diff
@@ -6,8 +6,9 @@
 
 #include <pbdrv/config.h>
 
-#include <pbio/dcmotor.h>
+#include <contiki.h>
```
dlech, line 9: To be consistent, the contiki include should be before pbdrv.

lib/pbio/drv/legodev/legodev_test.c
```diff
@@ -46,6 +81,9 @@ void pbdrv_legodev_init(void) {
             legodev->uart_dev = pbdrv_legodev_pup_uart_configure(0, 0, dcmotor);
         }
     }
+
+    // We would normally start the process here, but this also causes it to
+    // start in other tests where this is not needed.
```
dlech, line 86: So perhaps we should have a `#if !...TEST...` here and start the process here in normal firmware builds?

lib/pbio/drv/uart/uart.h
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2022 The Pybricks Authors
```
dlech, line 2:
```suggestion
// Copyright (c) 2023 The Pybricks Authors
```

lib/pbio/drv/uart/uart_stm32f0.c
```diff
@@ -265,6 +272,8 @@ PROCESS_THREAD(pbdrv_uart_process, ev, data) {
         uart->initialized = true;
     }
 
+    pbdrv_init_busy_down();
```
dlech, line 275: I don't see any yields between busy up and down, so it doesn't seem like it is needed here.

lib/pbio/drv/uart/uart_stm32f4_ll_irq.c
```diff
@@ -281,6 +288,8 @@ PROCESS_THREAD(pbdrv_uart_process, ev, data) {
         LL_USART_Enable(pdata->uart);
     }
 
+    pbdrv_init_busy_down();
```
dlech, line 291: Same here, no yields, so we don't need to bother with busy up/down.

lib/pbio/drv/uart/uart_stm32l4_ll_dma.c
```diff
@@ -469,6 +476,8 @@ PROCESS_THREAD(pbdrv_uart_process, ev, data) {
         LL_USART_Enable(pdata->uart);
     }
 
+    pbdrv_init_busy_down();
```
dlech, line 479: same here

lib/pbio/drv/legodev/legodev_pup.c
```diff
@@ -384,6 +376,35 @@ struct _pbdrv_legodev_dev_t {
...
+
+    PROCESS_BEGIN();
+
+    #if PBDRV_CONFIG_LEGODEV_PUP_POWER_CYCLE_PORTS
+    // Some hubs turn on power to the I/O ports in the bootloader. This causes
+    // UART sync delays after boot. It is faster to power cycle the I/O devices
+    // than it is to wait for long sync in most cases.
+    pbdrv_ioport_enable_vcc(false);
+    etimer_set(&timer, 500);
+    PROCESS_WAIT_EVENT_UNTIL(ev == PROCESS_EVENT_TIMER && etimer_expired(&timer));
```
dlech, line 394: Since this yields, we *might* need to surround it with busy up/down. However, if we are 100% sure that nothing in pbio/pbsys depends on this process actually being running or the VCC power being on, then it can be omitted.

lib/pbio/drv/legodev/legodev_pup.c
```diff
@@ -48,23 +44,37 @@ typedef enum {
 
 // Device connection manager state for each port
 typedef struct {
-    dev_id_group_t dev_id1_group;
+    /** Most recent one-off device ID candidate. */
```
dlech, line 47: I'm not sure what `one-off` means in this context. Perhaps there is a better word choice?

lib/pbio/drv/legodev/legodev_pup.c
```diff
@@ -314,62 +330,38 @@ static void debug_state_change(ext_dev_t *dev) {
 #define debug_state_change(...)
 #endif
 
-PROCESS_THREAD(pbio_legodev_pup_process, ev, data) {
-    static struct etimer timer;
+static PT_THREAD(pbdrv_legodev_pup_thread(ext_dev_t * dev)) {
 
-    static ext_dev_t *dev;
+    PT_BEGIN(&dev->pt);
```
dlech, line 335: It might save a few byte in the firmware to make a few local variables like `struct *pt = &dev->pt` to avoid multiple dereferences of the same pointer.

lib/pbio/drv/legodev/legodev_pup.c
```diff
@@ -314,62 +330,38 @@ static void debug_state_change(ext_dev_t *dev) {
...
-
-                // Nothing more to do if it's the same device as before.
-                if (dev->connected_type_id == dev->prev_type_id) {
-                    continue;
-                }
+        // Initially assume nothing is connected.
+        dev->dcm.connected_type_id = PBDRV_LEGODEV_TYPE_ID_NONE;
+        dev->dcm.dev_id_match_count = 0;
 
+        // Run the device detection manager until a UART device is detected.
```
dlech, line 343: It wouldn't hurt to add some more explanation here about the difference between UART devices and non-UART devices, in particular in relation to how disconnect is detected and why UART is a special case that stops the DCM but other devices don't.

lib/pbio/drv/legodev/legodev_pup_uart.c
```diff
@@ -156,12 +156,12 @@ typedef struct {
  * struct ev3_uart_port_data - Data for EV3/LPF2 UART Sensor communication
  */
 struct _pbdrv_legodev_pup_uart_dev_t {
-    /**< Protothread for main communication protocol. */
+    /**< Main protothread, first used for synchronization thread and then for data send thread. */
```
dlech, line 159: None of these comments, including existing ones should have `<` after `**`.

lib/pbio/drv/legodev/legodev_test.c
laurensvalk: The issue isn't in test vs. normal builds, but between test_x vs. test_y.

lib/pbio/drv/legodev/legodev_pup.c
laurensvalk: it gets 12 byte bigger

PR #195, opened 2023-08-29
Add setting for disabling Bluetooth

nkarstens:
Adds a button control to change the setting, and the light around the button to indicate current state.

Implements https://github.com/pybricks/support/issues/1123

### DIFF
--- a/lib/pbio/sys/hmi.c
+++ b/lib/pbio/sys/hmi.c
@@ -3,7 +3,7 @@
 
 // Provides Human Machine Interface (HMI) between hub and user.
 
-// TODO: implement additional buttons and Bluetooth light for SPIKE Prime
+// TODO: implement additional buttons for SPIKE Prime
 // TODO: implement additional buttons and menu system (via screen) for NXT
 
 #include <stdbool.h>
--- a/lib/pbio/sys/hmi.c
+++ b/lib/pbio/sys/hmi.c
@@ -19,6 +19,7 @@
 #include <pbio/color.h>
 #include <pbio/event.h>
 #include <pbio/light.h>
+#include <pbsys/bluetooth.h>
 #include <pbsys/config.h>
 #include <pbsys/status.h>
 
@@ -87,6 +88,9 @@ void pbsys_hmi_handle_event(process_event_t event, process_data_t data) {
  */
 void pbsys_hmi_poll(void) {
     pbio_button_flags_t btn;
+    bool bluetooth_disabled;
+
+    static bool bluetooth_toggled = false;
 
     if (pbio_button_is_pressed(&btn) == PBIO_SUCCESS) {
         if (btn & PBIO_BUTTON_CENTER) {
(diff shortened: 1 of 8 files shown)


lib/pbio/include/pbsys/program_load.h
```diff
@@ -41,6 +41,13 @@ typedef struct _pbsys_program_load_data_header_t {
      */
     volatile uint32_t checksum_complement;
     #endif
+    /**
```
nkarstens, line 44: I'm not sure if this is correct. It looks like one of the drivers requires a granularity of 64 bits.

lib/pbio/sys/hmi.c
```diff
@@ -87,19 +87,34 @@ void pbsys_hmi_handle_event(process_event_t event, process_data_t data) {
...
-        if (btn & PBIO_BUTTON_CENTER) {
+        if ((btn & PBIO_BUTTON_CENTER) == PBIO_BUTTON_CENTER) {
             pbsys_status_set(PBIO_PYBRICKS_STATUS_POWER_BUTTON_PRESSED);
             user_program_start(true);
 
             // power off when button is held down for 3 seconds
             if (pbsys_status_test_debounce(PBIO_PYBRICKS_STATUS_POWER_BUTTON_PRESSED, true, 3000)) {
                 pbsys_status_set(PBIO_PYBRICKS_STATUS_SHUTDOWN_REQUEST);
             }
+        } else if ((btn & PBIO_BUTTON_RIGHT_UP) == PBIO_BUTTON_RIGHT_UP) {
```
nkarstens, line 103: TODO: Add compile guards around this and set based on the hub type.

lib/pbio/sys/program_load.c
```diff
@@ -82,6 +82,15 @@ pbio_error_t pbsys_program_load_get_user_data(uint32_t offset, uint8_t **data, u
     return PBIO_SUCCESS;
 }
 
+void pbsys_program_load_set_bluetooth_disabled(bool bluetooth_disabled) {
+    map->header.bluetooth_disabled = bluetooth_disabled;
```
nkarstens, line 86: This doesn't seem to be working and I'm having trouble understanding why. Any ideas would be appreciated.

If I replace this code with reading/writing to a file-static variable then everything behaves as expected (though of course the setting is not persisent). As it is, it always reads false, both during startup and after I attempt to modify it.

laurensvalk:
Thanks for the contribution! We’ll review this when we can.

There’s some things we still need to decide about persistent settings, so maybe we can start without that part of it. It shouldn’t be too much trouble for people to press the Bluetooth button if they want to turn it off.

nkarstens:
@laurensvalk Sounds good, I'll update it to remove the patch for persistence and mark it ready for review.

nkarstens:
@dlech It would be great if I could get your thoughts on this one as well if you have time. I think there may be an outstanding question about persistent settings. @laurensvalk had mentioned there were some things still being worked on there, have you had a chance to figure anything out?

dlech:
My thinking right now is to finish USB support first, then come back to this. Having USB support could change the behavoir we want here.

nkarstens:
> Thanks @nkarstens. To add to David comment, I think it would be nice if the default behavior stays consistent with what we had, so Bluetooth on by default.
> 
> Besides consistency, I've always found it much more convenient too. With the official app you always have to do multiple clicks, button pushes, and so on, just to connect and run a program.

I can update this to change the default value. I wanted it off by default for our team so that it's one less thing they have to worry about during competition.

Ideally we could get setting storage figured out so that it would remember the previous state. If that won't happen for a while, then I would be fine merging this in with the default being Bluetooth enabled and I can just do a custom build for our team.

nkarstens:
Rebased for v3.3.0 release / v7.10.0 firmware.

nkarstens:
Fixed formatting issues.

sodes-proxy:
Hi!
@nkarstens @laurensvalk 
Is there a way for the artifacts to be created again?
or do you have any documentation on how I could generate a custom firmware with these changes?

I want to use Pybrick for WRO competition, having bluetooth enabled permanently is problematic for competition purposes. The rules (Section 5.11 in https://wro-association.org/wp-content/uploads/WRO-2023-RoboMission-General-Rules.pdf)
I appreciate any help you can provide me!!!

laurensvalk:
@sodes-proxy - please keep an eye on https://github.com/pybricks/support/issues/1615 later this week. Thank you!

laurensvalk:
A version of this has been merged in https://github.com/pybricks/pybricks-micropython/pull/247 and was made persistent in https://github.com/pybricks/pybricks-micropython/pull/249, so we'll close this one.

Thanks for your contribution @nkarstens !

------------------

> * Bluetooth is on by default, same as now (no breaking change).
> * Toggle Bluetooth on/off with button.
> * Bluetooth light: blue is on, red is off.
> * Make it _persistent_ after reboot. So you can achieve _off-on-boot_, for judge review.
> * Spike Prime / Robot Inventor only.

If you want to try it out before we release a new beta, just follow [these instructions](https://pybricks.com/install/spike-mindstorms/#installing-the-latest-build-advanced). That page includes a link to the latest build.

PR #209, opened 2023-10-26
Broadcasting from SPIKE in a tight loop

laurensvalk:
This is an exploratory draft. It fixes https://github.com/pybricks/support/issues/1151, but has problems when cancelling a program while broadcasting while disconnected from PC. The second time you start the example code, it won't broadcast.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,10 +23,13 @@
 - Fixed `surface=False` not working in `ColorSensor` ([support#1232]).
 - Fixed `PUPDevice.write` not selecting correct mode ([support#1213]).
 - Fixed City Hub turning back on after shutdown ([support#1195]).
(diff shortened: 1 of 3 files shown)


laurensvalk:
> but has problems when cancelling a program while broadcasting while disconnected from PC. The second time you start the example code, it won't broadcast.

This is because on cancel, `pb_type_BLE_cleanup` may be called before the `broadcast_task` completes, which sets `is_broadcasting`. So on the next start, it will think it is already broadcasting and not start.

laurensvalk:
We can avoid this problem if we don't mind _always_ waiting 10 ms on broadcast since then the wait logic can come after the broadcast and cancellation is no problem.

laurensvalk:
> We can avoid this problem if we don't mind always waiting 10 ms on broadcast since then the wait logic can come after the broadcast and cancellation is no problem.

I'm going with this simpler solution to expedite the release.

PR #210, opened 2023-11-04
drv/bluetooth_stm32_cc2640: work around broadcast lock up

dlech:
The Bluetooth chips on the Technic and City hubs seem to lock up if the advertisement date is updated too many times. So far, we have only found one condition where this doesn't happen. The Bluetooth chip can't be connected to any other devices and advertising has to be stopped and restarted between each update. This is quite limiting, but seems to be the best we can do for now.

Issue: https://github.com/pybricks/support/issues/1095


dlech:
I'm not very optimistic, but I'm still hoping we can find a better solution than this. Making the PR so others can have a firmware to try and can poke around with the code.

dlech:
Found a better solution: https://github.com/pybricks/support/issues/1095#issuecomment-1793560688

PR #212, opened 2023-11-13
pbio/drivebase: Reverse turn rate for negative drive speed.

laurensvalk:
Changed steering direction with `DriveBase.drive()` when reversing. Now,   a positive angle or turn rate means the robot travels along a circle to   its right. The speed or radius can be negated to travel in reverse along  that same circle.

See https://github.com/pybricks/support/issues/1191

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@
 - Added `MoveHub.imu.tilt()` ([support#539]).
 - Enabled hub init orientation support for Move Hub ([support#539]).
 
+### Changed
+- Changed steering direction with `DriveBase.drive()` when reversing. Now, 
+  a positive angle or turn rate means the robot travels along a circle to
+  its right. The speed or radius can be negated to travel in reverse along
+  that same circle ([support#1191]).
+
 ### Fixed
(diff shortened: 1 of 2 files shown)


laurensvalk:
@JDwyer009, would you like to give this a try?

To try it, download one of the firmware links in the post above.

In Pybricks, begin installing the firmware as usual. Instead of choosing a hub on the first dialog, pick "Advanced" and pick the file you just downloaded.

Proceed normally and try out the updated `drive` command :smile: 

Thanks!

dlech:
I feel like this would make things a bit weird when the drive speed is zero. Since zero is neither positive nor negative, it seems arbitrary that when drive speed is zero that the positive rules should apply. And the fact that the requested turn rate would no longer match the turn rate return by `state()` only when speed is negative also makes it feel weird.

So I still think it makes more sense to leave `drive()` as it is (and introduce a new method if we want something else).

laurensvalk:
For what it's worth, this variant doesn't have special treatment or siding it one way for the zero case.

laurensvalk:
> So I still think it makes more sense to leave `drive()` as it is (and introduce a new method if we want something else).

Agreed, so we should close this. Further discussion for other methods is in https://github.com/pybricks/support/issues/1414

PR #213, opened 2023-11-15
Create Drivebase_Tests_incl_Asserts

JDwyer009:
This is the same code as posted in the forum.
Let me know if I've not followed any particular protocols.


laurensvalk:
Thank you!

I've made a few updates and merged it:
- Changes none extension to `.py` and applied auto-formatting.
- Used ports A and B.
- Removed the prints, leaving only the asserts. When the CI tests see no output at all, we'll know the test succeeded.
- Updated `drive` directions to match the current situation until we figure out what to do with #212 :smile:

PR #214, opened 2023-11-23
Fix hang when animations are called too quickly.

davidwu226:
Fix hang from calling animations too quickly.

Fixes: pybricks/support#1295

### DIFF
--- a/lib/pbio/src/light/animation.c
+++ b/lib/pbio/src/light/animation.c
@@ -120,7 +120,11 @@ PROCESS_THREAD(pbio_light_animation_process, ev, data) {
         clock_time_t interval = animation->next(animation);
         if (pbio_light_animation_is_started(animation)) {
             etimer_reset_with_new_interval(&animation->timer, interval);
-            etimer_restart(&animation->timer);
+            /* If the timer is to fire in the past, restart it immediately instead. */
+            if (clock_time() > etimer_expiration_time(&animation->timer)) {
+                etimer_reset_with_new_interval(&animation->timer, 0);
+                etimer_restart(&animation->timer);
+            }
         }
     }
 
--- a/lib/pbio/src/light/animation.c
+++ b/lib/pbio/src/light/animation.c
@@ -50,10 +50,10 @@ void pbio_light_animation_start(pbio_light_animation_t *animation) {
     pbio_light_animation_list_head = animation;
 
     process_start(&pbio_light_animation_process);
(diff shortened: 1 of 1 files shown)


lib/pbio/src/light/animation.c
```diff
@@ -120,6 +120,7 @@ PROCESS_THREAD(pbio_light_animation_process, ev, data) {
         clock_time_t interval = animation->next(animation);
         if (pbio_light_animation_is_started(animation)) {
             etimer_reset_with_new_interval(&animation->timer, interval);
+            etimer_restart(&animation->timer);
```
dlech, line 123: We are already calling `etimer_reset_with_new_interval()` here, why do we also need `etimer_restart()`? It seems redundant.

lib/pbio/src/light/animation.c
```diff
@@ -50,10 +50,10 @@ void pbio_light_animation_start(pbio_light_animation_t *animation) {
...
 
     process_start(&pbio_light_animation_process);
-    // HACK: init timer since we don't call etimer_set()
-    timer_set(&animation->timer.timer, 0);
-    // fake a timer event to load the first cell and start the timer
-    process_post_synch(&pbio_light_animation_process, PROCESS_EVENT_TIMER, &animation->timer);
+
+    PROCESS_CONTEXT_BEGIN(&pbio_light_animation_process);
+    etimer_set(&animation->timer, 0);
+    PROCESS_CONTEXT_END(&pbio_light_animation_process);
```
dlech, line 56: This is better than the old hack. 👍

dlech (commented):
Can you add an explanation of why this fixes the bug?

Also we need to fix the test that broke because of this change.

And we should add a changelog entry.

lib/pbio/src/light/animation.c
davidwu226: I've never done any Contiki programming before so this might be wrong, but according to their docs, `etimer_reset()` sets the timer to expire from the previous expire time whereas `etimer_restart()` restarts the timer from the current time. Looking at the implementation for `etimer_reset_with_new_interval()`, it internally calls `etimer_reset()` (as you'd expect). I suspect with a low interval, it is possible that the new timer is set to fire in the past, which seems to cause problems.

davidwu226:
I've made a slight change to the fix so that it doesn't introduce drift unless it is necessary.

PR #215, opened 2023-11-23
Fix hang when animation intervals are too low.

davidwu226:
If the new timer for an animation is in the past, schedule it to fire immediately instead.

Fixes: https://github.com/pybricks/support/issues/1295


davidwu226:
NVM, this doesn't fix the problem.

PR #217, opened 2023-11-27
pybricks/parameters/pb_type_color: add distance method

Novakasa:
Exposing the new HCL-bicone color cost function (https://github.com/pybricks/pybricks-micropython/pull/104) makes it possible for users to create their own smarter color detection algorithms, for example by adding a hysteresis. This PR adds it as a `color.distance(other_color)` method to a `Color` instance. On movehub, this adds 84 bytes. The function name `get_distance_to` would have added 100 bytes.

Example use:

```python
hsv = sensor.hsv()
if hsv.distance(Color.GREEN) < hsv.distance(Color.RED):
    print("green!")
else:
    print("red!")
```

Not sure if this would add less firmware size if it were just a module-level function, but to me it makes a bit more sense as a `Color` method.

### DIFF
--- a/pybricks/parameters/pb_type_color.c
+++ b/pybricks/parameters/pb_type_color.c
@@ -217,6 +217,7 @@ STATIC void pb_type_Color_attr(mp_obj_t self_in, qstr attr, mp_obj_t *dest) {
             default:
                 break;
         }
+        dest[1] = MP_OBJ_SENTINEL;
         return;
     }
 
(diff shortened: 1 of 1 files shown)


laurensvalk:
Can you add some end-user Python samples?

Is this primarily to map(color1, color2) --> cost, or would you eventually like to be able to pass a custom cost function to `detectable_colors`?

PR #223, opened 2023-12-30
pbio/drv/ioport_debug_uart: Simplify debugging.

laurensvalk:
This adds a config flag to replace the last I/O port with UART to make debugging Bluetooth much easier.

```c
PBDRV_CONFIG_IOPORT_DEBUG_UART
```
This function is nonblocking and can be called from anywhere. If called again while busy printing, it will just skip the new message:

```c
#include <pbdrv/../../drv/ioport/ioport_debug_uart.h>

pbdrv_ioport_debug_printf("Hello, world! %d\n", 123456);
```

And this can be called within protothreads. Prints async until ready. Useful when full message is needed but timing is less critical.

```c
PBDRV_IOPORT_PT_DEBUG_PRINTF(pt, "Hello %d\n", 123456);
```

This is generalized to work on all hubs.

-----------

It works great on Technic Hub, Essential Hub, Prime Hub.

But not on the Move Hub and City Hub, so I must have missed something on those.


dlech (commented):
The basic idea seems reasonable, but the implementation seems like it could be easy to break things.

Perhaps another approach to accomplish the same thing in a safer way would be to create a new contiki process and a new debug print event. Then to do a non-blocking maybe-print just send an event to that process with the string as the data. (So `pbdrv_ioport_debug_uart_printf()` would be mostly the same but drop `pbdrv_uart_write_end()` and replace `pbdrv_uart_write_begin()` with `process_post_synch(&pbdrv_uart_debug_process, pbdrv_uart_debug_event, buffer)`.)

If a blocking version is needed (could have both PT and non-PT versions of this), Then it would just be:

```
wait for pbdrv_uart_debug_process not busy
pbdrv_ioport_debug_uart_printf(...)
wait for pbdrv_uart_debug_process not busy
```

laurensvalk:
Thanks for the feedback. I've looked into setting up such a process. How should it wait for write completion?

```c
 // Wait for uart to finish, expecting broadcast event from UART. 
PROCESS_WAIT_EVENT_UNTIL(pbdrv_uart_write_end(debug_uart) != PBIO_ERROR_AGAIN);
```

This line is mostly stuck waiting. Apparently it is not continuing on getting the completion broadcast event from the uart driver.

```c
PROCESS(pbdrv_ioport_debug_uart_process, "ioport_debug_uart");

void pbdrv_ioport_debug_uart_init(void) {
    pbdrv_init_busy_up();
    process_start(&pbdrv_ioport_debug_uart_process);
}

static bool busy_writing = false;
static char print_buffer[64];

pbio_error_t pbdrv_ioport_debug_uart_printf_maybe(const char *format, ...) {
    if (busy_writing) {
        return PBIO_ERROR_AGAIN;
    }
    busy_writing = true;

    // Buffer result.
    va_list args;
    va_start(args, format);
    vsnprintf(print_buffer, sizeof(print_buffer), format, args);
    va_end(args);

    // Request write to uart.
    process_poll(&pbdrv_ioport_debug_uart_process);
    return PBIO_SUCCESS;
}

PROCESS_THREAD(pbdrv_ioport_debug_uart_process, ev, data) {

    static struct etimer timer;

    const pbdrv_ioport_pup_port_platform_data_t *port_data = &pbdrv_ioport_pup_platform_data.ports[PBDRV_CONFIG_IOPORT_NUM_DEV - 1];

    PROCESS_BEGIN();

    // Wait for uart to be ready. This may not emit an event, so use our own.
    for (;;) {
        etimer_set(&timer, 100);
        PROCESS_WAIT_EVENT_UNTIL(ev == PROCESS_EVENT_TIMER && etimer_expired(&timer));
        if (pbdrv_uart_get(port_data->uart_driver_index, &debug_uart) == PBIO_SUCCESS) {
            break;
        }
    }

    // Enable and configure uart.
    const pbdrv_ioport_pup_pins_t *pins = &port_data->pins;
    pbdrv_gpio_alt(&pins->uart_rx, pins->uart_alt);
    pbdrv_gpio_alt(&pins->uart_tx, pins->uart_alt);
    pbdrv_gpio_out_low(&pins->uart_buf);

    pbdrv_uart_set_baud_rate(debug_uart, 115200);
    pbdrv_uart_flush(debug_uart);

    pbdrv_init_busy_down();

    for (;;) {
        // busy_writing was set before poll
        PROCESS_WAIT_EVENT_UNTIL(ev == PROCESS_EVENT_POLL);

        // Should be able to start writing.
        if (pbdrv_uart_write_begin(debug_uart, (uint8_t *)print_buffer, strlen(print_buffer), UART_TIMEOUT) != PBIO_SUCCESS) {
            busy_writing = false;
            printf("how did we get here\n");
            continue;
        }

        // Wait for uart to finish, expecting broadcast event from UART.
        PROCESS_WAIT_EVENT_UNTIL(pbdrv_uart_write_end(debug_uart) != PBIO_ERROR_AGAIN);
        busy_writing = false;
    }


    PROCESS_END();
}
```

dlech:
Which hub is getting stuck, all of them?

laurensvalk:
Currently on Technic.

SPIKE does seem to get through some messages, though sometimes slowly, suggesting it is only continuing due to other external events.

laurensvalk:
Hmm, I starting to think we might be overthinking this.

If the whole purpose is to make a process to manage the `is_busy` state, isn't this a check that belongs in the UART driver? Then we could make all of this without any additional processes or protothreads, I think.

> It seems wrong/dangerous to call pbdrv_uart_write_end() before pbdrv_uart_write_begin().

On Technic and Spike, `pbdrv_uart_write_end()` is essentially checking for "busy", so this is more or less doing as above.

So could we perhaps add `pbdrv_uart_write_busy()`?

dlech:
`pbdrv_uart_write_end()` has to be called after completion to reset some things, so I don't think it is quite that simple without major changes to the UART drivers. 

I kind of remember having issues like this with the sensors not always getting notified when new data was ready.

laurensvalk:
Since it is working well for the debugging purposes we use it for, I think we can merge it. It doesn't affect live builds.

I think we can further generalize this if / when we have generic UART access.

PR #229, opened 2024-01-21
drv/bluetooth/stm32_cc2640: send correct max size

dlech:
This finishes a TODO in the CC2640 driver to send the correct max payload size to the host via the Pybricks hub capabilities characteristic. Previously, this was hardcoded to use the hub max MTU size. But some Bluetooth adapters have a smaller MTU size, so we need to use the smaller of the two.

Fixes: https://github.com/pybricks/support/issues/947


dlech:
I added another commit to be sure we are responding to prepared (long) write requests. 

Previously, BlueZ would hang for a long time waiting for a response from these. Now it works as shown in the screenshot.

image

laurensvalk:
Thanks!

Is this expected to have any side effects at all or would it only affect the Pybricks Code connection on the affected hosts?

In order words, any platforms this should be extensively tested on? Otherwise happy to just merge.

dlech:
The changes are quite minimal so I feel like we have done adequate testing. The only behavior change will be for hosts that report an MTU less than 158. And we have verified that case is working as expected now since it was previously broken.

lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c
```diff
@@ -1094,12 +1095,17 @@ static void handle_event(uint8_t *packet) {
...
+
+                    // REVISIT: Just saving the main connection MTU for now.
+                    // If we allow multiple connections, this will need to be
+                    // changed.
+                    if (connection_handle == conn_handle) {
+                        conn_mtu = MIN(client_mtu, PBDRV_BLUETOOTH_MAX_MTU_SIZE);
+                    }
 
+                    attExchangeMTURsp_t rsp;
                     rsp.serverRxMTU = PBDRV_BLUETOOTH_MAX_MTU_SIZE;
```
laurensvalk, line 1108: Just curious, why does this response not use the `conn_mtu` value?

dlech: Per the Bluetooth spec, both the central and peripheral send their own value and take the lesser of the two. This way, it doesn't matter who asks first.

laurensvalk: Thanks! If you have time at some point, I have a few more Bluetooth questions 😄

PR #231, opened 2024-02-01
pbio/platform: Disable interrupts before data init.

laurensvalk:
We appear to be getting corrupted data or nonzero static global structs when running the firmware after leaving BLE firmware update mode. Depending on the build, this would make the hub not boot, or indefinitely reboot after installing the firmware.

Fixes https://github.com/pybricks/support/issues/1408


laurensvalk:
> I would even make it the first instruction in the file, so even before updating the stack pointer. (and mention in the comment that the bootloader doesn't do this which is why we have to do it here)

Done.

---------------

Is there any reason not to do this on Move Hub and City Hub as well? I've taken them along in this PR.

I'll be testing them now.

lib/pbio/platform/technic_hub/startup.s
```diff
@@ -76,6 +76,7 @@ defined in linker script */
 	.weak	Reset_Handler
 	.type	Reset_Handler, %function
 Reset_Handler:
+  cpsid i   /* Disable interrupts in case LEGO bootloader left them enabled before handover. */
```
dlech, line 79: nit: we know for certain that it leaves them enabled (from reverse engineering), so saying "in case" seems a bit misleading

lib/pbio/drv/core.c
```diff
@@ -92,8 +92,10 @@ void pbdrv_init(void) {
     pbdrv_usb_init();
     pbdrv_watchdog_init();
 
-    // Some hubs have a bootloader that disables interrupts. Has to be done
-    // here otherwise Essential hub can hang on boot if it is earlier.
+    // Some hubs have a bootloader that disables interrupts, and on other hubs
+    // we disable them early in the boot process to avoid corrupting statically
```
dlech, line 96: Could be simpler to just say that "interrupts are disabled when transitioning from the bootlaoder to our firmware" and leave it at that. (The bit about corrupting memory is just one of several possible side effects but isn't really relevant here. The really important information here is the last sentence explaining why we can't enable interrupts any sooner than this.)

dlech (commented):
Yes, this needs to be done on City and Move hubs since they have similar bootloaders that don't disable interrupts (unless there is something in platform.c that I forgot about). Just a few suggestions to improve the comments.

lib/pbio/drv/core.c
laurensvalk: Thanks, will update. But to nitpick further I couldn't quite follow the existing comment on why it was there and not before, or perhaps after setting up pbsys, etc.

:smile:

lib/pbio/platform/technic_hub/startup.s
dlech: yes

laurensvalk:
Thanks for suggesting this fix and the review!

lib/pbio/drv/core.c
dlech: Fair point. :smile: 

I never tracked it down to the exact problem so it is a bit vague. But basically, if an interrupt happens before the first init of all of the drivers is called, it can cause problems. But they need to be enabled for drivers with init_up/down waiting to be able to finish initializing.

PR #234, opened 2024-02-11
pbio/sys/program_stop: Drop hack to stop Bluetooth.

laurensvalk:
We have an old hack that shuts down the bluetooth process. When starting a new task after that in a cleanup routine such as added in 04fd2e770d97ab4771e77072bab0d9d61bcdeb45, the hub will never finish the task, which prevents the user program from stopping.

We have several proper fixes in place now, and we are about to merge some updates that will make all Bluetooth task objects static, so we should remove this hack.

Fixes https://github.com/pybricks/support/issues/1438

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Fix Bluetooth locking up when connecting Bluetooth adapter with small MTU to Technic and City hubs ([support#947]).
 - Fix Technic Hub not always starting automatically after firmware exiting
   update mode ([support#1408]). Also apply this to Move Hub and City Hub.
+- Fix hub not shutting down while a program runs ([support#1438]).
 
 ### Changed
 - Changed polarity of output in the `Light` class. This makes no difference for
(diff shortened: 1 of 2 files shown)


dlech:
This change looks good. 

But I suspect there is still more to do to ensure the hub can always be shut down without having to remove the batteries...

Are we sure we have fixed all places where tasks aren't cancellable?

If we use static memory for tasks now and force a program to stop while Bluetooth is still using a task, then start a new program that tries to reuse a statically allocated task that is still in use, it is going to likely cause a crash or other undefined behavior.

laurensvalk:
One step at a time :smile:

This change affects only forced stop on shutdown, so program restart should not be affected.

Thanks for the review. This can get merged for V3.4.0b2.

laurensvalk:
Merged via https://github.com/pybricks/pybricks-micropython/commit/0945da5e75fe44b2c9534bd8f7e5da4a2a53c28f

PR #239, opened 2024-02-27
pybricks.common: Dynamically create button enum.

laurensvalk:
This simplifies the button enum to make them more suitable beyond just pbio-specific keypad mappings. This will let them be used for other devices, like the Xbox Controller. (https://github.com/pybricks/support/issues/1488)

This also saves about 100 bytes on Move Hub. About half of this comes from omitting buttons that can't be used on Move Hub or with the remote, which we could theoretically also get without the dynamic enum (https://github.com/pybricks/support/issues/1453)


laurensvalk:
Got this down to about 200 bytes of savings, so it's probably worth the extra RAM used in scripts with buttons.

PR #240, opened 2024-03-01
bricks/movehub: Use ERROR_REPORTING_NONE.

laurensvalk:
This saves about 2.6KB without making the errors any less terse than they already were.

:exploding_head: 

For example

```python
from pybricks.hubs import MoveHub2
```

will still give the slightly more unhelpful error:

```
Traceback (most recent call last):
  File "movespace.py", in <module>
ImportError: 
```

So no great loss. I dug into this when adding a call to `raise_msg`, which added a bit too much build size for something that isn't supposed to do much.


dlech:
If the error output is the same, what is actually being removed that saves so much size?

laurensvalk:
Presumeably code with message handling now skipping that:

```c
#if MICROPY_ERROR_REPORTING == MICROPY_ERROR_REPORTING_NONE
#define mp_obj_new_exception_msg(exc_type, msg) mp_obj_new_exception(exc_type)
#define mp_obj_new_exception_msg_varg(exc_type, ...) mp_obj_new_exception(exc_type)
```

laurensvalk:
But I wasn't quite right about the savings, since the some of our own modules check for `MICROPY_ERROR_REPORTING_TERSE` but not for the stricter `NONE` option just enabled. So it was accidentally bringing back the verbose Pybricks error tips.

Now it saves about 4KB.

laurensvalk:
@BertLindeman , @JorgePe, would you be interested to try out this firmware to see if anything major is missing or broken?

If we really saved 4KB, we should be able to keep your move hubs going for quite a long time :smile:

JorgePe:
> @BertLindeman , @JorgePe, would you be interested to try out this firmware to see if anything major is missing or broken?
> 
> If we really saved 4KB, we should be able to keep your move hubs going for quite a long time 😄
> 
> The download links are shown in the posts above.

I sure can - I was just flashing another Move Hub for my snake. Is it available with Beta IDE or do I need to download it?

BertLindeman:
> @BertLindeman , @JorgePe, would you be interested to try out this firmware to see if anything major is missing or broken?
> 
> If we really saved 4KB, we should be able to keep your move hubs going for quite a long time 😄
> 
> The download links are shown in the posts above.

will do, Laurens

JorgePe:
Looks good to me. Tested Broadcasting (just observe), internal LED color and the 2 internal motors in multitask.

BertLindeman:
OK by me too, tested broadcast, observe, various import, name and other errors.
Made a testprogram with 20 functions so 200 subfunctions to test the program size, like this
```
def main_function0():
    def sub_function1(): 
        data1 = list(range(100))
    def sub_function2(): 
        data1 = list(range(100))
    def sub_function3(): 
        data1 = list(range(100))
    def sub_function4(): 
        data1 = list(range(100))
    def sub_function5(): 
        data1 = list(range(100))
    def sub_function6(): 
        data1 = list(range(100))
    def sub_function7(): 
        data1 = list(range(100))
    def sub_function8(): 
        data1 = list(range(100))
    def sub_function9(): 
        data1 = list(range(100))
    def sub_function10(): 
        data1 = list(range(100))
```

 
The compiler reports:
```
This program is too big. Compiled size is 6,160 bytes but the hub can only fit 3,956 bytes.

Try removing unused code or making names and strings smaller.
```

laurensvalk:
But I don't think the `TERSE` errors were especially helpful to begin with, so the `NONE` errors might be just fine, given the savings.

laurensvalk:
> OK by me too, tested broadcast, observe, various import, name and other errors.

Thanks @BertLindeman !

laurensvalk:
> Looks good to me. Tested Broadcasting (just observe), internal LED color and the 2 internal motors in multitask.

Awesome, thank you!

PR #241, opened 2024-03-05
bricks/ev3: New arm none port.

laurensvalk:
This adds a new bare arm Pybricks port for EV3 using the same `arm-none-eabi ` makefile used for the other ports.

The goal is to get to a standalone firmware that works just like the Powered Up hubs with Pybricks. This means they can share almost all code except for low-level drivers. So without Linux underneath. And with drivers to make it compatible with WebUSB for compatibility with Pybricks Code for easy use.

If we get that far, it should be easier to maintain this brick just like the other hubs long term, and solve most of the Linux-related issues we currently have such as motor control loop time problems. It also wouldn't require low-capacity microSD cards which are increasingly hard to get.

--------------

It boots all the way to MicroPython with standard output over UART on sensor port 1. It has a dummy static program preloaded. The expected output is:

```
ImportError: no module named 'pybricks.tools'
Traceback (most recent call last):
```

Which shows that MicroPython is running. Hardware peripherals are not implemented.

Unlike previous `nxt` and `ev3rt` ports, this properly starts and runs via `pbsys`, which means that we could build gradually from here, adding one hardware driver at a time.

To build:
```
make -C bricks/ev3 -j  uImage
```
Prepare a microSD card (16 GB or less) and format it as a single FAT32 partition. Then copy the `uImage` to it.

This should be modified to add a build target for a firmware that can be installed via the EV3 update mode so it can run without a microSD card and installed with a web USB driver via Pybricks Code.

-------------------

The minimalist linker file and `startup.s` was inspired by @pepijndevos ' [devos](https://github.com/pepijndevos/devOS).

-----

Inspiration for future hardware implementation:
- [ev3dev](https://www.ev3dev.org/docs/kernel-hackers-notebook/ev3dev-linux-kernel/): Well documented resources for EV3 on debian Linux.
- [EV3RT](https://github.com/pybricks/ev3rt-lib): RTOS (not Linux). Closer to the metal, but drivers mixed in with RTOS.
- [am18xlib](https://github.com/pybricks/am18x-lib-ev3): Good inspiration for minimal peripheral drivers, including PRU and so on.


pepijndevos:
Super exciting developments!

winklerp:
super exciting! Thank you very much!

jaguilar:
This is very exciting. To the extent you need anyone to help beta test, or even if you have small-scoped programming tasks that could be done by someone with limited knowledge of the system, please feel welcome to reach out to me.

BertLindeman:
I see the assets listed above.
I looked at the LEGO NXT software and the nxt-firmware generated here has no *.rfw file that the nxt software wants to update the firmware. 
EV3 wants a *.bin file.

It's probably too soon to try.
That's fine.
If it's time to do some preliminary testing. let me know.

J is so right: exiting!

jaguilar:
@BertLindeman I believe the bin files for the PRU are here: https://github.com/mindboards/ev3sources/tree/master/extra/linux-03.20.00.13/pru-firmware-05-31-2011-1423-v3.0/pru/firmware/bin

I'm trying to read and understand what the deal is with these things. As far as I can tell they are similar to the RP2040's programmable IO subsystem -- they're little coprocessors that can run a small instruction set and control IO pins. It looks like one of the purposes of this coprocessor is to implement UART functionality, which requires realtime scheduling, without involving the main CPU.

BertLindeman:
> @BertLindeman I believe the bin files for the PRU are here: https://github.com/mindboards/ev3sources/tree/master/extra/linux-03.20.00.13/pru-firmware-05-31-2011-1423-v3.0/pru/firmware/bin
> 
> I'm trying to read and understand what the deal is with these things. As far as I can tell they are similar to the RP2040's programmable IO subsystem -- they're little coprocessors that can run a small instruction set and control IO pins. It looks like one of the purposes of this coprocessor is to implement UART functionality, which requires realtime scheduling, without involving the main CPU.

Thanks, J.
This is (just a bit) beyond my hardware knowledge, Will look into it.
Bert

PR #242, opened 2024-03-22
pbio/platform/nxt: Use pbsys

laurensvalk:
This enables executing (but not downloading) programs and starting the REPL via pbsys on NXT, and removes a bunch of hacks that were previously needed to start the REPL. It also cleans up pbdrv/core in the process.

This builds on the previous progress in https://github.com/pybricks/pybricks-micropython/pull/241 and prior commits to simplify using pbsys in new ports. In particular, it now uses a test driver for a block device with a builtin dummy program. This lets us properly implement and run pbsys before we complete a method to download programs via USB or Bluetooth.


laurensvalk:
:rocket: 

image

PR #244, opened 2024-03-29
pybricks.tools.read_input_byte: Allow chr conversion of last byte.

laurensvalk:
The default behavior remains the same, but you can now optionally do:

```python
read_input_byte(chr=True, last=True)
```
to apply the chr() function and discard all up to the last byte.

Applying `chr` is useful for remote controlling something with the keyboard, and simplifies the block code in https://github.com/pybricks/support/issues/1574

`last=True` can be useful for various things, including using the arrow keys, which when combined with `chr` will then simply map to A/B/C/D instead of having to parse three bytes and adding logic to find the last one.

### DIFF
--- a/pybricks/tools/pb_module_tools.c
+++ b/pybricks/tools/pb_module_tools.c
@@ -152,22 +152,58 @@ mp_obj_t pb_module_tools_pbio_task_wait_or_await(pbio_task_t *task) {
 }
 
 /**
- * Reads one byte from stdin without blocking.
+ * Reads one byte from stdin without blocking if a byte is available, and
+ * optionally converts it to character representation.
  *
- * @returns The integer value of the byte read or @c None if no data is available.
+ * @param [in]  last    Choose @c True to read until the last byte is read
+ *                      or @c False to get the first available byte.
+ * @param [in]  chr     Choose @c False to return the integer value of the byte.
+ *                      Choose @c True to return a single character string of
+ *                      the resulting byte if it is printable and otherwise
+ *                      return @c None .
+ *
+ * @returns The resulting byte if there was one, converted as above, otherwise @c None .
+ *
  */
-STATIC mp_obj_t pb_module_tools_read_input_byte(void) {
-    if (!(mp_hal_stdio_poll(MP_STREAM_POLL_RD) & MP_STREAM_POLL_RD)) {
-        // No bytes available.
+STATIC mp_obj_t pb_module_tools_read_input_byte(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
+    PB_PARSE_ARGS_FUNCTION(n_args, pos_args, kw_args,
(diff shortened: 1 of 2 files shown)


pybricks/tools/pb_module_tools.c
```diff
@@ -152,22 +152,52 @@ mp_obj_t pb_module_tools_pbio_task_wait_or_await(pbio_task_t *task) {
 }
 
 /**
- * Reads one byte from stdin without blocking.
+ * Reads one byte or character from stdin without blocking.
```
dlech, line 155:
```suggestion
 * Reads one byte or one printable ASCII character from stdin without blocking.
```

to be unambiguous

pybricks/tools/pb_module_tools.c
```diff
@@ -152,22 +152,52 @@ mp_obj_t pb_module_tools_pbio_task_wait_or_await(pbio_task_t *task) {
...
  *
- * @returns The integer value of the byte read or @c None if no data is available.
+ * @param [in]  chr     Choose @c True to return a printable character or @c False to return an integer.
+ * @param [in]  last    Choose @c True to read until the last byte is read or @c False to return the first byte.
+ *
+ * @returns The value read or @c None if no valid data is available.
```
dlech, line 160: or if the ASCII character was not printable.

pybricks/tools/pb_module_tools.c
```diff
@@ -152,22 +152,52 @@ mp_obj_t pb_module_tools_pbio_task_wait_or_await(pbio_task_t *task) {
...
+        // REVISIT: In theory, this should not block if mp_hal_stdio_poll() and
+        // mp_hal_stdin_rx_chr() are implemented correctly and nothing happens
+        // in a thread/interrupt/kernel that changes the state.
+        chr = mp_hal_stdin_rx_chr();
+
+        // For last=True, read up to last, otherwise return first byte.
+        if (!mp_obj_is_true(last_in)) {
+            break;
+        }
+    }
```
dlech, line 179: Should we be skipping non-printable ASCII characters in the loop when `chr=True`?

laurensvalk: Good point. I've changed the argument order to make it clearer we first care about the first/last _byte_, and then look at converting to a single character string, if `chr` is `True`.

laurensvalk:
The block for this will have a dropdown (Pybricks style) instead of two bool options so we can keep it concise while still giving a more verbose description than we could on the block.

image

pybricks/tools/pb_module_tools.c
dlech: My question is if `last=False, chr=True`, and the buffer contains 3 non-printable characters and one printable character, is the desired behavoir to return `None` 3 times and then return the printable character? Because it seems like that's what would happen now.

PR #245, opened 2024-04-04
Unify three bluetooth de-init tasks and allow exit on shutdown request

laurensvalk:
We have previously implemented the MicroPython VM abort to exit MicroPython under all circumstances, but we have recently added a few pbio tasks that were started after this, and they are not cancellable. This prevents shutdown.

This unifies awaiting on the deinitialization of the bluetooth peripheral, bluetooth broadcasting/observing and bluetooth stdout.

It also adds a break for a shutdown request so the hub can shut down even if a pbio task is stuck (e.g. https://github.com/pybricks/support/issues/1419).

This also finishes implementing the remote disconnect task, which on some platforms was only started but not awaited inside the task. This could be used for a dedicated awaitable remote disconnect method in a future release, which has been requested several times.

This is largely untested, which is a work in progress.


lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c
```diff
@@ -1051,17 +1061,16 @@ static PT_THREAD(peripheral_disconnect_task(struct pt *pt, pbio_task_t *task)) {
     if (peri->con_handle != NO_CONNECTION) {
         PT_WAIT_WHILE(pt, write_xfer_size);
         GAP_TerminateLinkReq(peri->con_handle, 0x13);
-        PT_WAIT_UNTIL(pt, hci_command_status);
+        PT_WAIT_UNTIL(pt, peri->con_handle == NO_CONNECTION);
```
dlech, line 1064: I would keep the waiting for command status. The disconnect could happen for a different reason other than the request made here, so the program flow would continue while the command is still being processed on the BT chip.

pybricks/pybricks.c
```diff
@@ -133,10 +136,25 @@ void pb_package_pybricks_init(bool import_all) {
...
-    pb_type_BLE_cleanup();
+    pb_type_ble_start_cleanup();
     #endif
     // Disconnect from remote.
     #if PYBRICKS_PY_PUPDEVICES
-    pb_type_Remote_cleanup();
+    pb_type_lwp3device_start_cleanup();
     #endif // PYBRICKS_PY_PUPDEVICES
+
+    #if PYBRICKS_PY_COMMON_BLE && PYBRICKS_PY_PUPDEVICES
```
dlech, line 146: Should this be `||` instead of `&&`?

lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c
laurensvalk: So (in nominal conditions) are we guaranteed to get two events and always in that order?

pybricks/pybricks.c
laurensvalk: Thanks. I'll probably just have to be specific and add `PYBRICKS_PY_PUPDEVICES_REMOTE` to account for the virtualhub which has pupdevices but no remote, and then we can use `||`.

lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c
dlech: yes. and even if the disconnect event happens first due to non-nominal conditions `peri->con_handle == NO_CONNECTION` is still going to be true after the status event.

PR #246, opened 2024-05-09
pybricks.tools: Add hostbuffer to receive IDE data

laurensvalk:
This adds a class that allows a user program to allocate a buffer for receiving data from an IDE.

It has similarities to receiving data over standard in, but it is a dedicated separate command. The data is not queued but (over)written to a fixed size buffer. This is a bit like downloading a program, but with user-readable data.

This is useful for applications where the buffer contains "sensor data"-like contents prepared by the IDE.

Usage is as follows:

```python
from pybricks.tools import HostBuffer

# Allocates buffer 4 bytes. From here, the program is ready to receive
# data write commands to this buffer from a connected host like Pybricks Code. 
buffer = HostBuffer("hbb")

# Returns tuple with 3 values (int16, int8, int8)
values = buffer.get_values()

# Returns bytes copy of the 4-byte buffer.
data = buffer.get_bytes()

# gc collection of the buffer will automatically stop processing
# received data write commands. If Pybricks Code sends more, they
# are just ignored.
```

Before merging this as-is, it might be useful to thinking about the complementary case as well, where the user program can write data back to the IDE. This could facilitate things like live sensor port view without messing with stdout.

### DIFF
--- a/lib/pbio/sys/command.c
+++ b/lib/pbio/sys/command.c
@@ -15,15 +15,15 @@
 #include "./program_stop.h"
 #include "./user_program.h"
 
-static pbsys_command_write_program_data_buffer_callback_t write_program_data_buffer_callback = NULL;
+static pbsys_command_write_app_data_callback_t write_app_data_callback = NULL;
 
 /**
  * Sets callback for the write program data buffer command.
  *
  * @param [in]  callback  The callback to set or @c NULL to unset.
  */
-void pbsys_command_set_write_program_data_buffer_callback(pbsys_command_write_program_data_buffer_callback_t callback) {
-    write_program_data_buffer_callback = callback;
+void pbsys_command_set_write_app_data_callback(pbsys_command_write_app_data_callback_t callback) {
+    write_app_data_callback = callback;
 }
 
 /**
@@ -64,11 +64,11 @@ pbio_pybricks_error_t pbsys_command(const uint8_t *data, uint32_t size) {
             // If no consumers are configured, goes to "/dev/null" without error
             return PBIO_PYBRICKS_ERROR_OK;
         case PBIO_PYBRICKS_COMMAND_WRITE_PROGRAM_DATA_BUFFER:
-            if (write_program_data_buffer_callback && size > 3) {
+            if (write_app_data_callback && size > 3) {
                 uint16_t offset = pbio_get_uint16_le(&data[1]);
                 uint16_t data_size = size - 3;
                 const uint8_t *data_write = &data[3];
-                write_program_data_buffer_callback(offset, data_size, data_write);
+                write_app_data_callback(offset, data_size, data_write);
             }
             // If no consumers are configured, goes to "/dev/null" without error
             return PBIO_PYBRICKS_ERROR_OK;
--- a/lib/pbio/sys/command.c
+++ b/lib/pbio/sys/command.c
@@ -6,6 +6,7 @@
 
 #include <pbdrv/reset.h>
 #include <pbio/protocol.h>
+#include <pbsys/command.h>
 
 #include <pbsys/storage.h>
 
@@ -13,6 +14,17 @@
 #include "./storage.h"
 #include "./program_stop.h"
 #include "./user_program.h"
+
+static pbsys_command_write_program_data_buffer_callback_t write_program_data_buffer_callback = NULL;
+
+/**
+ * Sets callback for the write program data buffer command.
+ *
+ * @param [in]  callback  The callback to set or @c NULL to unset.
+ */
+void pbsys_command_set_write_program_data_buffer_callback(pbsys_command_write_program_data_buffer_callback_t callback) {
+    write_program_data_buffer_callback = callback;
+}
 
 /**
  * Parses binary data for command and dispatches handler for command.
@@ -49,6 +61,15 @@ pbio_pybricks_error_t pbsys_command(const uint8_t *data, uint32_t size) {
             }
             pbsys_bluetooth_rx_write(&data[1], size - 1);
             #endif
+            // If no consumers are configured, goes to "/dev/null" without error
+            return PBIO_PYBRICKS_ERROR_OK;
+        case PBIO_PYBRICKS_COMMAND_WRITE_PROGRAM_DATA_BUFFER:
+            if (write_program_data_buffer_callback && size > 3) {
+                uint16_t offset = pbio_get_uint16_le(&data[1]);
+                uint16_t data_size = size - 3;
+                const uint8_t *data_write = &data[3];
+                write_program_data_buffer_callback(offset, data_size, data_write);
+            }
             // If no consumers are configured, goes to "/dev/null" without error
             return PBIO_PYBRICKS_ERROR_OK;
         default:
--- a/lib/pbio/sys/command.c
+++ b/lib/pbio/sys/command.c
@@ -63,15 +63,23 @@ pbio_pybricks_error_t pbsys_command(const uint8_t *data, uint32_t size) {
             #endif
             // If no consumers are configured, goes to "/dev/null" without error
             return PBIO_PYBRICKS_ERROR_OK;
-        case PBIO_PYBRICKS_COMMAND_WRITE_PROGRAM_DATA_BUFFER:
-            if (write_app_data_callback && size > 3) {
-                uint16_t offset = pbio_get_uint16_le(&data[1]);
-                uint16_t data_size = size - 3;
-                const uint8_t *data_write = &data[3];
-                write_app_data_callback(offset, data_size, data_write);
+        case PBIO_PYBRICKS_COMMAND_WRITE_APP_DATA: {
+            if (!write_app_data_callback) {
+                // No errors when no consumer is configured. This avoids errors
+                // when data is sent after the program ends.
+                return PBIO_PYBRICKS_ERROR_OK;
             }
-            // If no consumers are configured, goes to "/dev/null" without error
-            return PBIO_PYBRICKS_ERROR_OK;
+
+            // Requires at least the message type and data offset.
+            if (size <= 3) {
+                return PBIO_PYBRICKS_ERROR_VALUE_NOT_ALLOWED;
+            }
+
+            uint16_t offset = pbio_get_uint16_le(&data[1]);
+            uint16_t data_size = size - 3;
+            const uint8_t *data_to_write = &data[3];
+            return pbio_pybricks_error_from_pbio_error(write_app_data_callback(offset, data_size, data_to_write));
+        }
         default:
             return PBIO_PYBRICKS_ERROR_INVALID_COMMAND;
     }
--- a/pybricks/tools/pb_type_hostbuffer.c
+++ b/pybricks/tools/pb_type_hostbuffer.c
@@ -5,6 +5,10 @@
 
 #if PYBRICKS_PY_TOOLS_HOSTBUFFER
 
+#include <string.h>
+
+#include <pbsys/command.h>
+
 #include "py/mphal.h"
 
 #include <pybricks/tools.h>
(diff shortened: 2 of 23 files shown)


lib/pbio/sys/command.c
```diff
@@ -48,6 +60,15 @@ pbio_pybricks_error_t pbsys_command(const uint8_t *data, uint32_t size) {
             #endif
             // If no consumers are configured, goes to "/dev/null" without error
             return PBIO_PYBRICKS_ERROR_OK;
+        case PBIO_PYBRICKS_COMMAND_WRITE_PROGRAM_DATA_BUFFER:
+            if (write_program_data_buffer_callback && size > 3) {
+                uint16_t offset = pbio_get_uint16_le(&data[1]);
+                uint16_t data_size = size - 3;
+                const uint8_t *data_write = &data[3];
+                write_program_data_buffer_callback(offset, data_size, data_write);
```
dlech, line 68: Why not return an error code here so we can notify the sender if they sent invalid data (e.g. too big offset)?

pybricks/tools/pb_type_hostbuffer.c
```diff
@@ -0,0 +1,118 @@
...
+    // object is constant as would be expected. Revisit: enable and return
+    // a memoryview, especially if using large buffers.
+    return mp_obj_new_bytes(self->bytes_obj.data, self->bytes_obj.len);
+}
+STATIC MP_DEFINE_CONST_FUN_OBJ_1(pb_type_hostbuffer_get_bytes_obj, pb_type_hostbuffer_get_bytes);
+
+STATIC mp_obj_t pb_type_hostbuffer_get_values(mp_obj_t self_in) {
+
+    // One-off import for ustruct.unpack.
+    static mp_obj_t ustruct_unpack = NULL;
```
dlech, line 51: This seems a bit dangerous. It might work on bare-metal hubs if ustruct_unpack is an address in flash memory that is always the same. Otherwise, we could end up with a pointer to random data after restarting the MicroPython runtime.

If there isn't a measurable performance gain for caching this in a normal program, I would just always do the lookup and not worry about caching it.

pybricks/tools/pb_type_hostbuffer.c
```diff
@@ -0,0 +1,118 @@
...
+
+static void handle_write_data_buffer(uint16_t offset, uint32_t size, const uint8_t *data) {
+    // Can't write if buffer does not exist or isn't big enough.
+    if (!hostbuffer_instance || offset + size > hostbuffer_instance->bytes_obj.len) {
+        return;
+    }
+    memcpy(hostbuffer_instance->buffer + offset, data, size);
+}
+
+STATIC mp_obj_t pb_type_hostbuffer_get_bytes(mp_obj_t self_in) {
```
dlech, line 39: To keep things a simple as possible, maybe we don't need this method?

dlech:
`HostBuffer` sounds like a very technical name. Have you been programming too much? :laughing: 

Maybe we could call it something a little less technical like `CommsData`?

(Being more technical in the Pybricks BLE protocol side of things is good though)

pybricks/tools/pb_type_hostbuffer.c
laurensvalk: I suppose that works. Giving full access to the bytes seemed more generic, i.e. the host could even dictate the format if it is unknown and send it as a zero terminated string at the start of the buffer.

lib/pbio/sys/command.c
laurensvalk: That could work, although I think success would be OK if the program isn't running, so we don't throw errors if a simple web app keeps sending all the time.

laurensvalk:
Thanks for the name suggestions. I was thinking this could be the base class for various applications, e.g. BallSensor, giving `x`, `y`, `width`, `height` as `%` of the canvas, with `bbbb` format. These could probably live in the user code or as a separate module instead of in the firmware.

The challenge with `CommsData` is that it's maybe less clear where the data is coming from. If we do that do, maybe `AppData` works too since it is coming from the Pybricks Code app.

dlech:
> maybe `AppData` works too since it is coming from the Pybricks Code app.

I like it.

laurensvalk:
> Before merging this as-is, it might be useful to thinking about the complementary case as well, where the user program can write data back to the IDE. This could facilitate things like live sensor port view without messing with stdout.

Looking ahead, since `AppData` doesn't imply a direction, maybe the other way around could be part of the same class.

Or possibly something like `AppDataReader` for this one and `AppDataWriter` for the potential future counterpart since it will work rather differently.

Higher level uses cases that could make use of a lower level `AppDataWriter`:
- sensor "port view" (this is requested a lot)
- variable monitor (sometimes requested)

dlech:
> Looking ahead, since `AppData` doesn't imply a direction, maybe the other way around could be part of the same class.

Makes sense.

```python
# rx-only
app = AppData("b")
# same as
app = AppData(rx="b")

# tx-only
app = AppData(tx="b")

# bidirectional
app = AppData("b", "b")
# same as
app = AppData(rx="b", tx="b")

# get latest
x, = app.rx()

# send
app.tx(1)
```

laurensvalk:
Depending on the buffer sizes we think could be used, it might be nice to somehow specify an offset, index, or slice, or something like it when reading and writing so you only have to send relevant or new data.

For example

```
app = AppData("bbbHhh")
distance = app.rx(index=3) # reads the value corresponding to H
```

dlech:
Not as pretty, but can already do that if it returns a tuple:

```python
distance = app.rx()[3]
```

laurensvalk:
> maybe the other way around could be part of the same class

`AppData.send(bytes_data)`

could be the sending side of this. Since the PC has infinite resources, we don't have to pre-allocate any buffers or use offsets. It would pretty much work like raw standard output, but it wouldn't be printed to the output pane.

afarago:
Are there any builds/versions where the AppData is also implemented for the bidirectional transfer?

As far as I understand current implementation of `HostBuffer` is able to receive data, but is not able to transmit.

laurensvalk:
That is not implemented yet. It would probably work a bit like `pbsys_bluetooth_tx`*, but with a new `PBIO_PYBRICKS_EVENT_WRITE_APPDATA = 2` to augment the existing `PBIO_PYBRICKS_EVENT_WRITE_STDOUT = 1` so Pybricks Code can tell them apart and not display the incoming app data.

In the firmware, it could perhaps have a separate buffer, similar to `stdout_ring_buf`. Or maybe there is a way to do it with a single ring buffer.

------

`*` Separately, in the bigger picture, `pbsys_bluetooth_tx` probably needs to become something like `pbsys_stdio_tx` so we can generalize it to BLE + USB.

dlech:
> Or maybe there is a way to do it with a single ring buffer.

Since this is only transmitting/receiving a fixed sized state, no ring buffer should be needed. 

If we want to guarantee that every state change is sent, the write should be a blocking/awaitable function.

The ring buffer for stdout is mainly to improve throughput of `print()` since it has variable-sized data (and also makes it non-blocking for small enough data rates)

laurensvalk:
> the write should be a blocking/awaitable function.

Excellent point, thanks. Yes, this should keep it fairly simple.

laurensvalk:
> > the write should be a blocking/awaitable function.
> 
> Excellent point, thanks. Yes, this should keep it fairly simple.

Since the Move Hub (MTU 23) probably won't support these features anyway, we could keep this even simpler by restricting outgoing messages to `PBDRV_CONFIG_BLUETOOTH_MAX_MTU_SIZE - 1` bytes in size. The next minimum is City/Technic at `158 - 1` bytes.

----

@afarago , what is the largest message you would typically need for a single portview information message? Is `157` bytes acceptable?

laurensvalk:
Just pushed an in-progress version that includes sending. The following works:

```python
from pybricks.tools import AppData

app = AppData("bb")
print(app.get_values())

print("Seen on stdout...")
app.write_bytes(b"Won't see this, but it will be sent!")  # This is also awaitable
```

`stdout` messages are labeled with `PBIO_PYBRICKS_EVENT_WRITE_STDOUT = 1`, while the app data output is labeled with `PBIO_PYBRICKS_EVENT_WRITE_APP_DATA = 2` so it doesn't pollute the terminal pane.

Each line is also guaranteed to be sent in one chunk rather than occasionally buffered like `stdout`, so this should be easier to parse on the Pybricks Code size.

afarago:
> Since the Move Hub (MTU 23) probably won't support these features anyway, we could keep this even simpler by restricting outgoing messages to `PBDRV_CONFIG_BLUETOOTH_MAX_MTU_SIZE - 1` bytes in size. The next minimum is City/Technic at `158 - 1` bytes.

I think so.
At the moment based on your suggestion I am going with one message per device, that is "BBll" for port number, port type, value1, value2, ... we might go on with this.
For the hub, we still have character data, but 157 shall be more than enough.

laurensvalk:
Sending longer `AppData` messages on Technic Hub fails. I think it boils down to the following:

```c
bStatus_t ATT_HandleValueNoti(uint16_t connHandle, attHandleValueNoti_t *pNoti) {
    uint8_t buf[32]; // <------- This seems very small. Should it be 158 - 3 + 5 = 160?

    buf[0] = connHandle & 0xFF;
    buf[1] = (connHandle >> 8) & 0xFF;
    buf[2] = 0; // authenticated link not required
    buf[3] = pNoti->handle & 0xFF;
    buf[4] = (pNoti->handle >> 8) & 0xFF;
    memcpy(&buf[5], pNoti->pValue, pNoti->len);

    return HCI_sendHCICommand(ATT_CMD_HANDLE_VALUE_NOTI, buf, 5 + pNoti->len);
}
```

What do you think, @dlech ?

I'm also wondering if this could potentially also explain some other related printing issues.

dlech:
Assuming that the central has a MTU >= the hub is probably a fairly safe assumption, but it is entirely possible that the central could be something like a Move Hub with max MTU of 23 bytes.

So ideally, the `AppData` Python object would be aware of the negotiated MTU and use that to calculate the max size rather than. the hub's max value.

As for the CC2640 chips, it could very well be that the 20 byte max payload for notifications is hard-coded in the Bluetooth chip firmware. But I do see a `GATT_UpdateMTU()` function, so maybe we need to call this after the MTU exchange?



> I'm also wondering if this could potentially also explain some other related printing issues.

The stdout buffer is chunked into 20 bytes (`MAX_CHAR_SIZE`) so wouldn't be overflowing the 32 byte buffer or exceeding the limit you found when trying to send more, if that is what you mean.

laurensvalk:
> Assuming that the central has a MTU >= the hub is probably a fairly safe assumption, but it is entirely possible that the central could be something like a Move Hub with max MTU of 23 bytes.
> 
> So ideally, the `AppData` Python object would be aware of the negotiated MTU and use that to calculate the max size rather than using the hub's hard-coded max value.
> 
> As for the CC2640 chips, it could very well be that the 20 byte max payload for notifications is hard-coded in the Bluetooth chip firmware. But I do see a `GATT_UpdateMTU()` function, so maybe we need to call this after the MTU exchange to tell the Bluetooth chip resize it's notification buffer?

:+1: 

I think I'll adjust this PR to use `MAX_CHAR_SIZE` for each chunk, and then we can look at improving throughput separately.

> > I'm also wondering if this could potentially also explain some other related printing issues.
> 
> The stdout buffer is chunked into 20 bytes (`MAX_CHAR_SIZE`) so wouldn't be overflowing the 32 byte buffer or exceeding the limit you found when trying to send more, if that is what you mean.

I was thinking of https://github.com/pybricks/support/issues/1152, but maybe this is just a different issue. That should indeed be protected by the `MAX_CHAR_SIZE`.

techcatgato:
Hello, I have a few questions about AppData.
1.is the sending of the messages from the hub only available on beta.pybricks.com?
2. How can I send a list between the hub and pc?
3. How can the pc recieve the list in a regular python script?

PR #247, opened 2024-06-04
Allow Bluetooth to be toggled off/on with Bluetooth button.

laurensvalk:
Implements toggling Bluetooth on and off with the Bluetooth button on Spike Prime and Mindstorms Robot Inventor.

See https://github.com/pybricks/support/issues/1615 for context, use cases, and user interface considerations.

This is a slightly more minimalist approach compared to #195. It adds `PBIO_PYBRICKS_STATUS_BLUETOOTH_BLE_ENABLED` to the protocol instead of `PBIO_PYBRICKS_STATUS_BLUETOOTH_BUTTON_PRESSED` and `PBIO_PYBRICKS_STATUS_BLUETOOTH_SHUTDOWN`.

The `hmi` is implemented similar to the existing code for the start/stop program button, and separated from the actual Bluetooth start/stop logic. The `hmi` may request a Bluetooth toggle, but it is up to `sys/bluetooth` whether to do anything with that. This prevents Bluetooth from being toggled if is currently being used or while a program is running.

The `PBIO_PYBRICKS_STATUS_BLUETOOTH_BLE_ENABLED` flag is only used for raising exceptions when the user tries to create remote peripheral or broadcast/observe.

Persistency is a separate task that can be added later. This was holding up previous merges, but we should be able to get started without it.

Thanks to everyone for asking about this feature and thanks to @nkarstens for creating #195!

### DIFF
--- a/lib/pbio/sys/bluetooth.c
+++ b/lib/pbio/sys/bluetooth.c
@@ -47,18 +47,6 @@ PROCESS(pbsys_bluetooth_process, "Bluetooth");
 
 // Internal API
 
-static void bluetooth_start(void) {
-    pbsys_status_set(PBIO_PYBRICKS_STATUS_BLUETOOTH_BLE_ENABLED);
-    pbsys_status_light_bluetooth_set_color(PBIO_COLOR_BLUE);
-    process_start(&pbsys_bluetooth_process);
-}
-
-static void bluetooth_stop(void) {
-    pbsys_status_clear(PBIO_PYBRICKS_STATUS_BLUETOOTH_BLE_ENABLED);
-    pbsys_status_light_bluetooth_set_color(PBIO_COLOR_RED);
-    process_exit(&pbsys_bluetooth_process);
-}
-
 /** Initializes Bluetooth. */
 void pbsys_bluetooth_init(void) {
     // enough for two packets, one currently being sent and one to be ready
@@ -70,7 +58,7 @@ void pbsys_bluetooth_init(void) {
     lwrb_init(&stdout_ring_buf, stdout_buf, PBIO_ARRAY_SIZE(stdout_buf));
     lwrb_init(&stdin_ring_buf, stdin_buf, PBIO_ARRAY_SIZE(stdin_buf));
 
-    bluetooth_start();
+    process_start(&pbsys_bluetooth_process);
 }
 
 /**
@@ -210,6 +198,8 @@ bool pbsys_bluetooth_tx_is_idle(void) {
     return !send_busy && lwrb_get_full(&stdout_ring_buf) == 0;
 }
 
+static bool pbsys_bluetooth_user_enabled = true;
+
 void pbsys_bluetooth_enabled_state_request_toggle(void) {
 
     // Ignore toggle request in all but idle system status.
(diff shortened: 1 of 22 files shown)


lib/pbio/sys/bluetooth.c
```diff
@@ -195,6 +210,28 @@ bool pbsys_bluetooth_tx_is_idle(void) {
     return !send_busy && lwrb_get_full(&stdout_ring_buf) == 0;
 }
 
+void pbsys_bluetooth_enabled_state_request_toggle(void) {
```
dlech, line 213: Instead of all of these status checks and starting/stopping the bluetooth thread, I think we could just add some checks in the bluetooth thread at the appropriate times. Stopping the thread seems like an easy way to break things unintentionally.

```diff
diff --git a/lib/pbio/sys/bluetooth.c b/lib/pbio/sys/bluetooth.c
index 2eea566b..84268ee0 100644
--- a/lib/pbio/sys/bluetooth.c
+++ b/lib/pbio/sys/bluetooth.c
@@ -290,6 +290,9 @@ PROCESS_THREAD(pbsys_bluetooth_process, ev, data) {
     pbdrv_bluetooth_set_receive_handler(handle_receive);
 
     while (!pbsys_status_test(PBIO_PYBRICKS_STATUS_SHUTDOWN)) {
+        // TODO: may need to also check shutdown status here
+        PROCESS_WAIT_WHILE(pbsys_status_test(PBIO_PYBRICKS_STATUS_BLE_OFF));
+
         // make sure the Bluetooth chip is in reset long enough to actually reset
         etimer_set(&timer, 150);
         PROCESS_WAIT_EVENT_UNTIL(ev == PROCESS_EVENT_TIMER && etimer_expired(&timer));
@@ -305,7 +308,8 @@ PROCESS_THREAD(pbsys_bluetooth_process, ev, data) {
         PROCESS_WAIT_UNTIL(
             pbdrv_bluetooth_is_connected(PBDRV_BLUETOOTH_CONNECTION_LE)
             || pbsys_status_test(PBIO_PYBRICKS_STATUS_USER_PROGRAM_RUNNING)
-            || pbsys_status_test(PBIO_PYBRICKS_STATUS_SHUTDOWN));
+            || pbsys_status_test(PBIO_PYBRICKS_STATUS_SHUTDOWN)
+            || pbsys_status_test(PBIO_PYBRICKS_STATUS_BLE_OFF));
 
         if (!pbdrv_bluetooth_is_connected(PBDRV_BLUETOOTH_CONNECTION_LE)) {
             // if a connection wasn't made, we need to manually stop advertising
@@ -314,6 +318,10 @@ PROCESS_THREAD(pbsys_bluetooth_process, ev, data) {
 
         pbsys_status_clear(PBIO_PYBRICKS_STATUS_BLE_ADVERTISING);
 
+        if (pbsys_status_test(PBIO_PYBRICKS_STATUS_BLE_OFF)) {
+            goto ble_off;
+        }
+
         PT_INIT(&status_monitor_pt);
 
         while (pbdrv_bluetooth_is_connected(PBDRV_BLUETOOTH_CONNECTION_LE)
@@ -351,6 +359,7 @@ PROCESS_THREAD(pbsys_bluetooth_process, ev, data) {
             PROCESS_WAIT_EVENT();
         }
 
+ble_off:
         reset_all();
         PROCESS_WAIT_WHILE(pbsys_status_test(PBIO_PYBRICKS_STATUS_USER_PROGRAM_RUNNING));
 
```

laurensvalk: Thank you! I ended up doing something similar last night. We can even omit the `goto` and achieve the same by testing the BLE active status in that loop.

laurensvalk:
In rare occasions it can get stuck at the bottom of the `pbsys/bluetooth` process.

This seems odd as it would have traversed this path too all the time so far when running a really short program. Or maybe we just haven't noticed it before?

```
        // reset Bluetooth chip
        pbsys_status_light_bluetooth_set_color(PBIO_COLOR_GREEN); // debug light
        pbdrv_bluetooth_power_on(false);
        PROCESS_WAIT_WHILE(pbdrv_bluetooth_is_ready());             // light can get stuck staying green
        pbsys_status_light_bluetooth_set_color(PBIO_COLOR_RED); // debug light
```

nothingtodowu:
I have a question that if I want set bluetooth off by default, how to change this? Thanks.

nothingtodowu:
Thanks a lot.

PR #250, opened 2024-06-07
pbio/sys/hmi: Reduce time needed to power off hub.

laurensvalk:
Reduces the power off time from 3 to 2 seconds. This has been requested occasionally. I think this is OK.

If it makes certain use cases harder, like long press menus, we could now quite easily add a user setting to make this timeout configurable due to https://github.com/pybricks/support/issues/1622 now being implemented.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
   correction of the `hub.imu.heading()` value ([support#1678]).
 - Added `update_heading_correction` to interactively set the heading
   correction value ([support#1678]).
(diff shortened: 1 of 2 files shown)


laurensvalk:
This was requested among others by @biasedlogic.

laurensvalk:
We will merge this since there haven't been any objections for a month. If anyone needs a longer shutdown-time out, please feel free to open a new issue. Then we make make the duration a user-configurable setting. Perhaps as a second argument to the existing `set_stop_button` method.

PR #251, opened 2024-06-10
platform/ev3: Integrate am18x-lib.

laurensvalk:
This integrates the library part of https://github.com/turmary/am18x-lib (not its many examples) into the bare metal EV3 port to do some experimentation with AM1808 peripherals, interrupts, and so on. It's not quite a full low level HAL, but it's a lot more than nothing.

We don't have to merge or use this, but having something that actually works and can print debugging information is helpful, so I wanted to at least archive this intermediate result as a pull request.

For example, this could be used to implement the clock driver and the `__wfi` to get basic waiting working, as done in the final commit.

Maybe I should now repeat this exercise with the TI Starterware... :nerd_face:


laurensvalk:
This can probably be merged. The only existing code this touches is `bricks/_common/arm_none_eabi.mk`, but the additions are protected by `ifeq ($(PB_MCU_FAMILY),TIAM1808)`.

This will require some minor changes to the automatically starting program, so I'll leave it open for the time being.

laurensvalk:
This pull request is a nice way to archive this exploration. Let's close it for the time being. Keeping the existing bare metal EV3 build as-is makes it a nice starting point for new ports.

PR #252, opened 2024-06-25
pbio/imu: Add user adjustable gyro correction.

laurensvalk:
For rationale and usage, see https://github.com/pybricks/support/issues/1678


laurensvalk:
I'm going to merge this as is. This way the persistent data feature will get some fresh testing before we release it in beta.

We can still improve the user-facing Python calibration function as we go. See https://github.com/pybricks/support/issues/1678 for follow ups on that.

PR #255, opened 2024-07-31
pybricks.iodevices.LWP3Device: match incomplete uuid list

dlech:
The new Technic Move hub uses the correct advertisement data type for the LEGO Wireless Protocol (LWP) service UUID. Older hubs use the flag for a complete list instead of an incomplete list. So to match all hubs, we need to check for both.

TODO:

- [x] add changelog entry


laurensvalk:
Looks good to me. Nice to see the generalized Bluetooth scanning refactoring pay off. 

I'm not around today, so feel free to merge at your convenience.

PR #256, opened 2024-08-01
Update _builtin_port_view.py

BertLindeman:
An unknown device would not be removed on unplug.
It would be reported as plugged-in until another device came on on that port.

### DIFF
--- a/bricks/_common/modules/_builtin_port_view.py
+++ b/bricks/_common/modules/_builtin_port_view.py
@@ -124,8 +124,9 @@ def update_dc_motor(port, type_id):
 
 
 # Any unknown Powered Up device.
-def unknown_pup_device(port, type_id, dev):
+def unknown_pup_device(port, type_id):
     while True:
+        PUPDevice(port)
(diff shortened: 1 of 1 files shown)


bricks/_common/modules/_builtin_port_view.py
```diff
@@ -124,8 +124,9 @@ def update_dc_motor(port, type_id):
 
 
 # Any unknown Powered Up device.
-def unknown_pup_device(port, type_id, dev):
+def unknown_pup_device(port, type_id):
     while True:
+        dev = PUPDevice(port)  # reset on the un-plug of the unknown device
```
dlech, line 129:
```suggestion
        PUPDevice(port)  # reset on the un-plug of the unknown device
```

`dev` isn't used, so can be left out.

BertLindeman:
Have updated the patch-1 branch But No idea how to update this pull request.

dlech:
It got updated already. We can squash the commits when we merge it.

laurensvalk:
Merged, thank you!

PR #261, opened 2024-08-23
Refactor Bluetooth light UI. Use SPIKE Prime Bluetooth Light.

laurensvalk:
~~tl;dr: Use a subset of original light UI when Bluetooth is off (instead of inventing something new) so we don't lock ourselves into something when we release the upcoming beta.~~

----

~~https://github.com/pybricks/pybricks-micropython/pull/247 introduced the ability to toggle Bluetooth on and off. The main light UI was kept unchanged, and a blue (on) and red (off) Bluetooth light was added. As already stated in that issue and pull request, this UI was not ideal since there are too many lights, which can be confusing for users and competition judges. So we'd have to come up with something else before we can release the firmware in beta.~~

~~Instead of adding new lights, this PR makes it a subset of the existing UI. When Bluetooth is off, the hub status light is off when it would otherwise be on or blink (when connected or advertising). There is no light indication around the Bluetooth light. This seems fairly intuitive. We could potentially still have the red light around the Bluetooth button when off, or save this for when we redo the whole light UI.~~

See updated below.

### DIFF
--- a/lib/pbio/include/pbio/protocol.h
+++ b/lib/pbio/include/pbio/protocol.h
@@ -323,6 +323,12 @@ typedef enum {
      * @since Pybricks Profile v1.2.0
      */
     PBIO_PYBRICKS_STATUS_SHUTDOWN_REQUEST = 8,
+    /**
+     * Hub is connected to a host (like Pybricks Code) via BLE.
+     *
+     * @since Pybricks Profile v1.4.0
+     */
+    PBIO_PYBRICKS_STATUS_BLE_HOST_CONNECTED = 9,
     /** Total number of indications. */
     NUM_PBIO_PYBRICKS_STATUS,
 } pbio_pybricks_status_t;
--- a/lib/pbio/include/pbio/protocol.h
+++ b/lib/pbio/include/pbio/protocol.h
@@ -294,7 +294,7 @@ typedef enum {
      */
     PBIO_PYBRICKS_STATUS_BLE_ADVERTISING = 3,
     /**
-     * Bluetooth Low Energy has low signal.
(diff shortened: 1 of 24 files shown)


laurensvalk:
This refactors the status light patterns:

It split the single `pbsys_status_light_indication_t` into separate `pbsys_status_light_indication_ble_t` and `pbsys_status_light_indication_warning_t`. On hubs with separate main and bluetooth lights, these can be displayed separately. On hubs with one light, the warnings can overlay/override the bluetooth state.

Overall, the behavior on Move/City/Technic/Essential Hub is mostly the same as before. While we are doing this, the patterns are slightly tweaked for better visibility across hubs.

On Prime Hub, this moves all Bluetooth state information to the Bluetooth light. This also disables the blue fading animation and the blue blinking shutdown animation on the main button, which now look out of place. This information is also conveyed by the 5x5 light matrix.

Could potentially be refactored further with https://github.com/pybricks/support/issues/1798.

lib/pbio/include/pbio/protocol.h
```diff
@@ -277,6 +277,12 @@ typedef enum {
      * @since Pybricks Profile v1.2.0
      */
     PBIO_PYBRICKS_STATUS_SHUTDOWN_REQUEST = 8,
+    /**
+     * Hub is connected to a host (like Pybricks Code) via BLE.
```
dlech, line 281: Do we also need a peripheral connected status?

dlech (commented):
LGTM

lib/pbio/include/pbio/protocol.h
laurensvalk: Deferred to https://github.com/pybricks/support/issues/1716

PR #262, opened 2024-08-24
HACK: fix await wait(0)

dlech:
This is what my late-night brain came up with. Likely a better way to do this.

Fixes: https://github.com/pybricks/support/issues/1775

PR #263, opened 2024-08-26
pybricks.tools.wait: Fix awaiting on wait(0).

laurensvalk:
Fixes https://github.com/pybricks/support/issues/1429
Fixes https://github.com/pybricks/support/issues/1775
Fixes https://github.com/pybricks/support/issues/1460

Alternate for https://github.com/pybricks/pybricks-micropython/pull/262.

The following program now prints normally, while before it would print nothing.

```python
from pybricks.tools import multitask, run_task, wait

async def main1():
    while True:
        await wait(0)

async def main2():
    while True:
        print('Hello!')
        await wait(500)


async def main():
    await multitask(main1(), main2())

run_task(main())
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,9 +28,11 @@
 ### Fixed
 - Fixed not able to connect to new Technic Move hub with `LWP3Device()`.
 - Removed `gc_collect()` from `tools.run_task()` loop to fix unwanted delays.
+- Fixed `await wait(0)` never yielding, so parallel tasks could lock up ([support#1429]).
 
 [pybricks-micropython#250]: https://github.com/pybricks/pybricks-micropython/pull/250
 [pybricks-micropython#253]: https://github.com/pybricks/pybricks-micropython/pull/253
+[support#1429]: https://github.com/pybricks/support/issues/1429
 [support#1615]: https://github.com/pybricks/support/issues/1615
 [support#1622]: https://github.com/pybricks/support/issues/1622
 [support#1678]: https://github.com/pybricks/support/issues/1678
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,9 +30,16 @@
 - Removed `gc_collect()` from `tools.run_task()` loop to fix unwanted delays.
 - Fixed `await wait(0)` never yielding, so parallel tasks could lock up ([support#1429]).
 
+### Removed
+- Removed `loop_time` argument to `pybricks.tools.run_task` as this wasn't
+  having the desired effect, and would cause loop round trips to take `10 ms`
+  for every `await wait(1)` ([support#1460]). Passing the argument is still
+  allowed for backwards compatibility, but it won't do anything.
+
 [pybricks-micropython#250]: https://github.com/pybricks/pybricks-micropython/pull/250
 [pybricks-micropython#253]: https://github.com/pybricks/pybricks-micropython/pull/253
 [support#1429]: https://github.com/pybricks/support/issues/1429
+[support#1460]: https://github.com/pybricks/support/issues/1460
 [support#1615]: https://github.com/pybricks/support/issues/1615
 [support#1622]: https://github.com/pybricks/support/issues/1622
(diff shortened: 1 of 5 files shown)


dlech (commented):
It think this is fine.

Another approach I was considering was replacing the `mp_obj_t obj, uint32_t end_time` parameters with the awaitable object itself so that callback could do whatever it wanted. But that struct is private for now, so I didn't know if there way a good reason to keep it that way.

pybricks/tools/pb_type_awaitable.c
```diff
@@ -54,6 +54,23 @@ static mp_obj_t pb_type_awaitable_close(mp_obj_t self_in) {
 }
 static MP_DEFINE_CONST_FUN_OBJ_1(pb_type_awaitable_close_obj, pb_type_awaitable_close);
 
+/**
+ * Special completion test for awaitable that is always complete.
+ */
+static bool pb_type_awaitable_test_completion_always_complete(mp_obj_t obj, uint32_t end_time) {
```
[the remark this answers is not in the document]
laurensvalk: Thanks! I missed it.

pybricks/tools/pb_module_tools.c
```diff
@@ -226,31 +229,15 @@ static mp_obj_t pb_module_tools_run_task(size_t n_args, const mp_obj_t *pos_args
...
-    uint32_t start_time = mp_hal_ticks_ms();
-    uint32_t loop_time = pb_obj_get_positive_int(loop_time_in);
-
     mp_obj_iter_buf_t iter_buf;
     mp_obj_t iterable = mp_getiter(task_in, &iter_buf);
 
     nlr_buf_t nlr;
     if (nlr_push(&nlr) == 0) {
-
+        run_loop_is_active = true;
```
laurensvalk, line 237: Need to check if we should be doing this here or before calling `mp_getiter` above.

The thinking to move it here was to ensure `run_loop_is_active` is only set if `mp_getiter` doesn't raise, but we'll need to make sure the flag isn't set too late.

dlech: Could potentially be a problem if someone did something like `run_task(wait(0))`.

Seems like it should be fine to move the `mp_get_iter()` inside of the nlr block too (after setting run loop active flag) to avoid this.

dlech:
>  Passing the argument is still allowed for backwards compatibility.

It's going to change the behavior, so not really compatible. So probably better to just remove the argument completely. It wasn't documented anyway. Unless block language was using this?

pybricks/tools/pb_module_tools.c
laurensvalk: I'll do that, but running `run_task(wait(123))` isn't going to work either way and didn't previously, since the argument is always evaluated before `run_task` starts.

 (This is a known limitation, but it isn't restrictive. Probably because the one use case where it does make sense, `run_task(multitask(main1(), main2()))`, does work).

laurensvalk:
> > Passing the argument is still allowed for backwards compatibility.
> 
> It's going to change the behavior, so not really compatible. So probably better to just remove the argument completely. It wasn't documented anyway. Unless block language was using this?

Sounds good to me. The block language also wasn't using it.

CHANGELOG.md
```diff
@@ -28,9 +28,18 @@
 ### Fixed
 - Fixed not able to connect to new Technic Move hub with `LWP3Device()`.
 - Removed `gc_collect()` from `tools.run_task()` loop to fix unwanted delays.
+- Fixed `await wait(0)` never yielding, so parallel tasks could lock up ([support#1429]).
+
+### Removed
+- Removed `loop_time` argument to `pybricks.tools.run_task` as this wasn't
+  having the desired effect, and would cause loop round trips to take `10 ms`
+  for every `await wait(1)` ([support#1460]). Passing the argument is still
+  allowed for backwards compatibility, but it won't do anything.
```
dlech, line 37: Looks like we missed fixing up the changelog for removing loop_time completely.

laurensvalk: Thanks, will fix as part of revision in #254

PR #264, opened 2024-08-28
Allow for multiple builtin program including Port View, take 2.

laurensvalk:
I closed https://github.com/pybricks/pybricks-micropython/pull/254 accidentally. This is a follow up. It is conceptually similar, but incorporates David's feedback given there. 

There is now a one byte program identifier instead of (4 bytes + 1 bool for the type). Instead of modifying the `START_REPL` and` START_PROGRAM` protocol commands with an identifier payload as before, this introduces a single new command `PBIO_PYBRICKS_COMMAND_START_USER_OR_BUILTIN_PROGRAM` with a one byte payload.

It also adds this byte to the hub status report, but I'm open to implementing this differently. The available builtin programs are now individually included in the feature flags.

I haven't tried making this work with any particular port view visualization yet. We can do that after this is finalized.

Making this work required some refactoring in the way we wait for programs to start. This is a bit simpler now, which makes this an overal reduction in code size of about 100 bytes on Move Hub, despite already making some progress towards out-of-the-box multi-program support on the bigger hubs.


CHANGELOG.md
```diff
@@ -24,6 +32,8 @@
   the hub ([pybricks-micropython#250]).
 - Improved font for the digits ``0--9`` when displaying them
   with `hub.display.char(str(x))` ([pybricks-micropython#253]).
+- Added an optional 32-bit program identifier payload to the protocol commands
```
dlech, line 35: Looks like leftover from the previous approach.

lib/pbio/sys/status.c
```diff
@@ -35,6 +37,30 @@ static void pbsys_status_update_flag(pbio_pybricks_status_t status, bool set) {
         (process_data_t)status);
 }
 
+/**
+ * Writes Pybricks status report command to @p buf
+ *
+ * @param [in]  buf        The buffer to hold the binary data.
+ * @return                 The number of bytes written to @p buf.
+ */
+uint32_t pbsys_status_write_status_report(uint8_t *buf) {
```
dlech, line 46: This looks out of place. Would expect it to be with the rest of the similar functions in `pybricks.c`.

We can just add a `pbsys_status_get_program_id()` instead.

laurensvalk: I considered this initially, but assembling the `status` payload in `status.c` (where the status state is kept) does seem about right.

We could add a pure function to `protocol.c` to be complete I suppose. This was initially omitted it since it almost repeats the other one.

Both would seem a bit more natural than the current approach, where the Bluetooth system driver is passing its version of the system state to build the status message.

CHANGELOG.md
laurensvalk: Thanks!

lib/pbio/include/pbio/protocol.h
```diff
@@ -147,8 +183,24 @@ typedef enum {
...
+     *            downloaded user programs. Slots 128--255 are for builtin user
+     *            programs.
+     *
+     * Errors:
+     * - ::PBIO_PYBRICKS_ERROR_BUSY if another program is already running.
+     * - ::PBIO_PYBRICKS_ERROR_INVALID_COMMAND if the requested program is not available.
+     *
+     * @since Pybricks Profile v1.4.0.
+     */
+    PBIO_PYBRICKS_COMMAND_START_USER_OR_BUILTIN_PROGRAM = 8,
```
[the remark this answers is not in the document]
laurensvalk: :+1: for less verbose.

> And I still think it would be fine to just add an optional parameter to `PBIO_PYBRICKS_COMMAND_START_USER_PROGRAM` instead of adding a new command if we want to go that route instead.

The new one will also return different errors (e.g. when requested id is not available). But I'll look at merging them.

lib/pbio/include/pbsys/config.h
```diff
@@ -8,6 +8,16 @@
...
 
+#include <pbio/protocol.h>
+
+#define PBSYS_CONFIG_APP_FEATURE_FLAGS (0 \
+    + PBSYS_CONFIG_FEATURE_BUILTIN_USER_PROGRAM_REPL * PBIO_PYBRICKS_FEATURE_FLAG_BUILTIN_USER_PROGRAM_REPL \
+    + PBSYS_CONFIG_FEATURE_BUILTIN_USER_PROGRAM_PORT_VIEW * PBIO_PYBRICKS_FEATURE_FLAG_BUILTIN_USER_PROGRAM_PORT_VIEW \
+    + PBSYS_CONFIG_FEATURE_BUILTIN_USER_PROGRAM_IMU_CALIBRATION * PBIO_PYBRICKS_FEATURE_FLAG_BUILTIN_USER_PROGRAM_IMU_CALIBRATION \
+    + PBSYS_CONFIG_FEATURE_PROGRAM_FORMAT_MULTI_MPY_V6 * PBIO_PYBRICKS_FEATURE_FLAG_USER_PROG_FORMAT_MULTI_MPY_V6 \
+    + PBSYS_CONFIG_FEATURE_PROGRAM_FORMAT_MULTI_MPY_V6_1_NATIVE * PBIO_PYBRICKS_FEATURE_FLAG_USER_PROG_FORMAT_MULTI_MPY_V6_1_NATIVE \
+    )
```
dlech, line 19: Perhaps a bit unconventional but it works. :smile: 

Maybe worth adding a comment too since it isn't super-obvious how this works.

dlech (commented):
This is probably good enough for now.

One thing I not so sure about though is how to deal with multiple "user" slots in Pybricks Code.  

It would probably be a good idea to add some kind of way to read which slots have a valid program in them.

With WebBluetooth, we can't get the actual error code for failures. So if we attempt to run a program in a slot with a bad program, there is no way to know why the error occurred. Was it out of range? Was it a bad program? Did the hub battery die? Right now, we don't have any "expected" errors, so we kind of ignore this problem.

So that's why I'm thinking it would be better to have something where we can "ask first" before trying something that would fail, so that we don't actually ever try to do something that might fail under normal circumstances.

laurensvalk:
@afarago, as requested: here is how you can easily hack/modify the existing `createStartReplCommand` in Pybricks Code to trigger Port View.

```ts
/**
 */
export function createStartReplCommand(): Uint8Array {
    const msg = new Uint8Array(2);
    const view = new DataView(msg.buffer);
    view.setUint8(0, 1);      // < ---- start program command
    view.setUint8(1, 129);  // <---- program ID, 129 is port view
    return msg;
}
```

PR #267, opened 2024-10-01
bricks/_common/mpconfigport.h: enable MICROPY_GC_ALLOC_THRESHOLD

dlech:
This can be a useful performance tweak for some applications.

Probably not worth turning on on hubs with smaller RAM though.


laurensvalk:
> Probably not worth turning on on hubs with smaller RAM though.

Is it useful on Technic Hub, or perhaps Prime/Inventor/Essential only?

laurensvalk:
Merged and enabled for hubs with `PYBRICKS_OPT_EXTRA_LEVEL2.`

PR #268, opened 2024-10-03
pbio/light/animation: Fix lockup on restarting animation during animation frame.

laurensvalk:
Fixes https://github.com/pybricks/support/issues/1295 for both the hub light and the light matrix

### DIFF
--- a/lib/pbio/src/light/animation.c
+++ b/lib/pbio/src/light/animation.c
@@ -50,10 +50,10 @@ void pbio_light_animation_start(pbio_light_animation_t *animation) {
     pbio_light_animation_list_head = animation;
 
     process_start(&pbio_light_animation_process);
-    // HACK: init timer since we don't call etimer_set()
-    timer_set(&animation->timer.timer, 0);
-    // fake a timer event to load the first cell and start the timer
-    process_post_synch(&pbio_light_animation_process, PROCESS_EVENT_TIMER, &animation->timer);
+    // Fake a timer event to load the first cell.
+    PROCESS_CONTEXT_BEGIN(&pbio_light_animation_process);
+    etimer_set(&animation->timer, 0);
+    PROCESS_CONTEXT_END(&pbio_light_animation_process);
 
     assert(animation->next_animation != PBIO_LIGHT_ANIMATION_STOPPED);
 }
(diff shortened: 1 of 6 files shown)


lib/pbio/src/light/animation.c
```diff
@@ -50,10 +50,10 @@ void pbio_light_animation_start(pbio_light_animation_t *animation) {
...
 
     process_start(&pbio_light_animation_process);
-    // HACK: init timer since we don't call etimer_set()
-    timer_set(&animation->timer.timer, 0);
-    // fake a timer event to load the first cell and start the timer
-    process_post_synch(&pbio_light_animation_process, PROCESS_EVENT_TIMER, &animation->timer);
+    // Fake a timer event to load the first cell.
+    PROCESS_CONTEXT_BEGIN(&pbio_light_animation_process);
+    etimer_set(&animation->timer, 0);
+    PROCESS_CONTEXT_END(&pbio_light_animation_process);
```
laurensvalk, line 56: I thought I'd clean this up while touching this code, but it looks like the tests expect this to be synchronous using the aforementioned hack. We can revert this bit or add

```
    while (pbio_do_one_event()) {
    }
```
to the tests. I'll update later.

lib/pbio/src/light/animation.c
```diff
@@ -117,10 +117,15 @@ PROCESS_THREAD(pbio_light_animation_process, ev, data) {
         PROCESS_WAIT_EVENT_UNTIL(ev == PROCESS_EVENT_TIMER);
         struct etimer *timer = data;
         pbio_light_animation_t *animation = PBIO_CONTAINER_OF(timer, pbio_light_animation_t, timer);
-        clock_time_t interval = animation->next(animation);
-        if (pbio_light_animation_is_started(animation)) {
-            etimer_reset_with_new_interval(&animation->timer, interval);
+
+        // Like all etimer events, only handle expired etimers. Need this check
+        // in case the timer was stopped or reset before this event was handled.
+        if (!etimer_expired(&animation->timer) || !pbio_light_animation_is_started(animation)) {
```
dlech, line 123:
```suggestion
        if (!etimer_expired(timer) || !pbio_light_animation_is_started(animation)) {
```

We already have a pointer to the timer.

dlech, line 123: It seems like continuing when timer is expired (due to animation not started) could be problematic since we wouldn't get another timer event. Should we be resetting the timer in that case?

lib/pbio/test/src/test_color_light.c
```diff
@@ -45,6 +46,15 @@ static const pbio_color_light_funcs_t test_light_funcs = {
     .set_hsv = test_light_set_hsv,
 };
 
+/**
+ * Handles pending events to ensure animation has started and first frame
+ * has been activated.
+ */
+static void wait_for_first_animation_frame(void) {
```
dlech, line 53: This function would be better name something like `handle_pending_events` since it isn't actually waiting for any specific condition other than pending events.

And since other tests need this too, might as well make this a helper function accessible to all tests (e.g. a static inline function in `test-pbio.h`).

lib/pbio/src/light/animation.c
laurensvalk: Isn't this the same as how it was, wrapping around the loop if there was nothing to do?

laurensvalk: Thank you. I suppose then we can use the conventional pattern as we do everywhere else:

```c
PROCESS_WAIT_EVENT_UNTIL(ev == PROCESS_EVENT_TIMER && etimer_expired(data));
```

PR #269, opened 2024-10-04
pybricks.common.BLE: Implement observe_enable()

laurensvalk:
This allows users to toggle observing off and on so that they can alternate between observing and broadcasting.

This results in better stability on Technic Hub which can lock up when simultaneous advertising and broadcasting.

Fixes https://github.com/pybricks/support/issues/1806

Also fix the remaining FIXME about the default broadcast channel while we are working on this code (first commit).

Usage:

```
hub.ble.observe_enable(True) # Enable observing (on by default)
hub.ble.observe_enable(False) # Disable observing
```

Can be combined with `hub.ble.broadcast(data)` and `hub.ble.broadcast(None)` to alternate between broadcasting and observing for better reliability on Technic Hub.

This isn't extensively tested yet, but making this available for others to try.

-------------

@NStrijbosch, @JJHackimoto

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 - Relaxed speed limit from 1000 deg/s to 1200 deg/s for external Boost
   motor ([support#1623]).
+- Make `broadcast_channel` optional instead of defaulting to `0`.
 
 ### Fixed
 - Fixed persistent data not being deleted when swapping
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 - Allow color objects to be iterated as h, s, v = color_object or indexed
   as color_object[0]. This allows access to these properties in block
   coding ([support#1661]).
+- Added `observe_enable` to the hub `BLE` class to selectively turn observing
+  on and off, just like you can with broadcasting ([support#1806]).
 
 ### Changed
 
@@ -32,6 +34,7 @@
 [support#1623]: https://github.com/pybricks/support/issues/1623
 [support#1661]: https://github.com/pybricks/support/issues/1661
 [support#1668]: https://github.com/pybricks/support/issues/1668
+[support#1806]: https://github.com/pybricks/support/issues/1806
 [support#1846]: https://github.com/pybricks/support/issues/1846
 [support#1858]: https://github.com/pybricks/support/issues/1858
 [support#1863]: https://github.com/pybricks/support/issues/1863
--- a/pybricks/common/pb_type_ble.c
+++ b/pybricks/common/pb_type_ble.c
@@ -52,7 +52,7 @@ static pbio_task_t broadcast_task;
 
 typedef struct {
     mp_obj_base_t base;
-    uint8_t broadcast_channel;
+    mp_obj_t broadcast_channel;
     pbio_task_t *broadcast_task;
     observed_data_t observed_data[];
 } pb_obj_BLE_t;
@@ -263,14 +263,9 @@ static mp_obj_t pb_module_ble_broadcast(size_t n_args, const mp_obj_t *pos_args,
     // move hub is connected to Pybricks Code. Also, broadcasting interferes
     // with observing even when not connected to Pybricks Code.
 
-    // FIXME: This check is (and should only be) done in the BLE constructor,
-    // but it may still pass there since 0 is a valid broadcast channel. That
-    // should be fixed by defaulting to None if no broadcast channel is provided.
-    #if PBSYS_CONFIG_BLUETOOTH_TOGGLE
-    if (!pbsys_storage_settings_bluetooth_enabled()) {
-        mp_raise_msg(&mp_type_RuntimeError, MP_ERROR_TEXT("Bluetooth not enabled"));
+    if (self->broadcast_channel == mp_const_none) {
+        mp_raise_msg(&mp_type_RuntimeError, MP_ERROR_TEXT("no broadcast channel selected"));
     }
-    #endif // PBSYS_CONFIG_BLUETOOTH_TOGGLE
 
     // Stop broadcasting if data is None.
     if (data_in == mp_const_none) {
@@ -309,7 +304,7 @@ static mp_obj_t pb_module_ble_broadcast(size_t n_args, const mp_obj_t *pos_args,
     value.v.data[0] = index + 4; // length
     value.v.data[1] = MFG_SPECIFIC;
     pbio_set_uint16_le(&value.v.data[2], LEGO_CID);
-    value.v.data[4] = self->broadcast_channel;
+    value.v.data[4] = mp_obj_get_int(self->broadcast_channel);
 
     pbdrv_bluetooth_start_broadcasting(self->broadcast_task, &value.v);
     return pb_module_tools_pbio_task_wait_or_await(self->broadcast_task);
@@ -516,7 +511,7 @@ static MP_DEFINE_CONST_OBJ_TYPE(pb_type_BLE,
  *
  * Do not call this function more than once unless pb_type_ble_start_cleanup() is called first.
  *
- * @param [in]  broadcast_channel_in    (int) The channel number to use for broadcasting.
+ * @param [in]  broadcast_channel_in    (int) The channel number to use for broadcasting or None for no broadcasting.
  * @param [in]  observe_channels_in     (list[int]) A list of channels numbers to observe.
  * @returns                             A newly allocated object.
  * @throws ValueError                   If either parameter contains an out of range channel number.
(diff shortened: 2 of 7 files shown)


laurensvalk:
@@JJHackimoto, if you want to try this with **blocks**:

1. Create a new _python_ file called `issue_1806_lib.py` with the following contents:

```python
from pybricks.tools import run_task

async def observe_enable_async(hub, enable):
    await hub.ble.observe_enable(enable)

def observe_enable(hub, enable):
    if run_task():
        return observe_enable_async(hub, enable)
    hub.ble.observe_enable(enable)
```

2. Then you can use this into your block program as follows:

image

Naturally, the Technic Hub still has the usual issue of not observing very well while still connected to the PC. So you may want to build a different example to test things.

3. Here is the sending program I used:

image

JJHackimoto:
Sadly not working. I'm getting the following error in the issue_1806_lib.py file when trying to run my blocks code:

File "issue_1806_lib.py", line 4, in observe_enable_async
AttributeError: 'BLE' object has no attribute 'observe_enable'

BertLindeman:
> Sadly not working. I'm getting the following error in the issue_1806_lib.py file when trying to run my blocks code:
> 
> File "issue_1806_lib.py", line 4, in observe_enable_async AttributeError: 'BLE' object has no attribute 'observe_enable'
> 
> I'm using PyBricks Beta (Checking for updates returns up to date)

Did you install the firmware for this update?
See [artifacts for this update](https://github.com/pybricks/pybricks-micropython/actions/runs/11184426727)

JJHackimoto:
Must have completely missed that! Thanks!

Happy to see the firmware installation bug on M1 Macs have been adressed, installing went smoothly on all three hubs!
I'll test the observing enable feature and report back.

JJHackimoto:
I can't seem to get any observing going after having called the new functions. It runs fine until I've disabled observing, then after enabling it again, it never observes again. Here's part of my code. Both of the print-functions in the "Broadcast" function runs, but after this, the "Hello" print no longer appears. It seems to get stuck at observing as if it was still disabled.
Any clues? I'll go ahead and create a simpler program if you cannot find anything obvious here.

Skärmavbild 2024-10-10 kl  18 15 47

laurensvalk:
It’s hard to say for sure since we don’t see the whole code, but your ‘hello’ loop ends when you set that loop condition to false, so this seems like the expected result.

JJHackimoto:
Yes, you are absolutely correct. My bad.

It seems to be working now, I'll do some more testing in the coming days, but it surely looks promising :)
Quick question, how long do you think I should be broadcasting for to make sure the signal is picked up by the other hubs? Is there any reason why it wouldn't pick it up right away (unless I've turned off observing at that specific time ofc)? Also, for how long after broadcasting will the signal still be able to be picked up, or is there no delay like that?

laurensvalk:
For Technic Hubs, observing works best when disconnected from the PC. I don't know that there are any guarantees, but it should pick it up within half a second?

Especially if you also occasionally intentionally disable observing in this way. This is ultimately just a workaround to trade speed for better long term reliability in advanced creations like yours :slightly_smiling_face: 

Depending on the application specifics, I suppose you can find something that works well enough for you, but there might not be a generic solution that covers all cases in an optimal way.

> It seems to be working now, I'll do some more testing in the coming days, but it surely looks promising :)

Nice!

JJHackimoto:
Thanks for the info! I didn't have any broadcasts that wasn't also observed so that works well!

However, I'm sorry to let you know that after about 10 minutes today, one of my hubs crashed. Same as before, no response and only resets when holding the power button.

All three hubs are flashed with the latest firmware for this fix and I'm using the loop in the screenshot so I don't think it should be able to broadcast and observe at the same time.
Skärmavbild 2024-10-13 kl  15 21 53

laurensvalk:
Can you please share your whole program that reproduces the issue? It's hard to say what could be wrong otherwise. Thanks!

laurensvalk:
I've merged this so it can be part of today's beta release. We can still refine it further before the release as needed.

JJHackimoto:
> In your example, I don't see broadcasting being disabled. Remember that broadcasting happens continuously until you choose `broadcast(None)`.

Oh, that's news to me. I had no idea broadcasting could also be turned off like this. I'll try this setup when I got some time (I guess this is how it would be done when using blocks). Thanks for letting me know. Will share my full program if this doesn't work.
Skärmavbild 2024-10-16 kl  21 11 15

JJHackimoto:
Has this feature been implemented or do I still need to call the observe_enable external task (issue_1806_lib.py) for this to work? I have been putting this off for quite some time but I'm finally getting back into it. Last time, I kept receiving the value 0 instead of None when no recent value was broadcasted. I want to try again with the new firmware if an implementation has been made.

laurensvalk:
There is no dedicated block for it, so importing as you have sounds like a good option.

JJHackimoto:
Ok thanks!

Here's my findings from today.

When broadcasting and observing a list, the default value is not "none", but 0. Broadcasting and observing a single variable has a default value of "none". Is this intended?

I have tested my programs multiple times during the day, and observing is very inconsistent. More than half the time, the broadcasted value won't be registered at all. If I connect my computer and print every single observation, they stay at "none", even when I know a broadcast has been sent. I'm broadcasting for a full second, and observation is done each 150ms. Worth noting, only the hubs that runs both broadcasting and observing are having this issue (maybe enabling broadcasting after being disabled is not instant?) The hub that only broadcasts can be observed by the other hubs all the time, no issues.

I'll share all my code in case someone can help me out.

For my observation function, I'm using a temporary variable to detect if the value observed is not "none". If so, that value is then applied to the actual variable. This is because the rest of the code would break if the variable was "none" when expected result is bool. My wish is that the variable observed should stay as last observed if no new observation was made.

If any part of the code needs explaining, then please let me know and I'll help. I feel like we are getting closer with this. No crashes were detected at all today!
[pybricks-backup-2024-12-01_19-19-57.zip](https://github.com/user-attachments/files/17970105/pybricks-backup-2024-12-01_19-19-57.zip)

laurensvalk:
Could you include screenshots of your code? You can use the camera icon on the right hand side and just drag the files in here. Thanks!

> When broadcasting and observing a list, the default value is not "none", but 0. Broadcasting and observing a single variable has a default value of "none". Is this intended?

I don't really understand  what you mean about the default value of a list having a value. A list is either empty or contains the values you have put in it.

> For my observation function, I'm using a temporary variable to detect if the value observed is not "none". If so, that value is then applied to the actual variable. This is because the rest of the code would break if the variable was "none" when expected result is bool. My wish is that the variable observed should stay as last observed if no new observation was made.

I suppose this would make it harder to detect the absence of incoming data, but maybe it could make sense to have this "data deletion timeout" extended or make it user configurable.

> I have tested my programs multiple times during the day, and observing is very inconsistent. More than half the time, the broadcasted value won't be registered at all. If I connect my computer and print every single observation, they stay at "none", even when I know a broadcast has been sent.

Combining broadcasting, observing **_and_** being connected to your PC is a bit too much for the Technic Hub to handle. Observing really only works well if disconnected. Maybe you could use the hub light to indicate if something was received?

JJHackimoto:
> Could you include screenshots of your code? You can use the camera icon on the right hand side and just drag the files in here. Thanks!
> 
Sure! I've attached the screenshots of my current code.
> 
> I don't really understand what you mean about the default value of a list having a value. A list is either empty or contains the values you have put in it.
> 
Well, if observing channel 4 (and nothing is broadcasting there), the observed value is "none". However, if observing and expecting a list of two variables, those two variables will instead get the value 0 (after unpacking the list). I expected them to also get the value "none".
> 
> I suppose this would make it harder to detect the absence of incoming data, but maybe it could make sense to have this "data deletion timeout" extended or make it user configurable.
> 
Yeah, I'm not sure what a good implementation would be for this. It could also easily get confusing, so maybe it's fine as is.
> 
> Combining broadcasting, observing **_and_** being connected to your PC is a bit too much for the Technic Hub to handle. Observing really only works well if disconnected. Maybe you could use the hub light to indicate if something was received?
> 
Good idea! I'll try that and report back.

Sensor Hub:
pybricks_program_2024-12-02T16_26_53
Drive Hub:
pybricks_program_2024-12-02T16_26_58
Distributor Hub:
pybricks_program_2024-12-02T16_27_02

JJHackimoto:
I just tested again for about 15 minutes, and the hub light went cyan each time the hub sends a message, but watching the other hubs made it clear that the message got through only about 50% of the time or less. I couldn't see any pattern as to when it worked and when it didn't.

laurensvalk:
Thanks. I think I'd need to try running this.

If I understand correctly you have two hubs communicating bidirectionally and a third hub that is broadcasting to the first two also. I'll try setting up a small example to see how well this can work.

JJHackimoto:
Yes that's correct. Thanks!

JJHackimoto:
Hi! Have you had time to test? :)
I made a youtube video about a day ago showcasing my project (And someone at Pybricks commented!), but it was difficult to make since I had to do multiple takes until it all worked. I made it look like it works since I believe it soon will :)

laurensvalk:
I started giving it a go, but quickly realized it gets pretty cumbersome at the user level. It should still work though.

So I've been thinking it might make sense to toggle between observing and broadcasting at the firmware level instead.

Since we know now that this works around the freezing, this might be worth doing and then call it good on bluetooth for this hub :) 

We are already doing something like this to constantly restart observing, so having a toggle like this shouldn't change the overal driver code by that much.

JJHackimoto:
Alright that sounds awesome to me! So basically it would take care of itself under the hood and all I do is to broadcast and observe at the same time like I used to do? (When this is implemented ofc)

JJHackimoto:
Hi! How's it going with this? Or maybe there's been a Christmas break :)

JJHackimoto:
@laurensvalk  Any news on this?

laurensvalk:
No news at the firmware level. I know @JanderII has been playing with this in a user script --- that is to say working on a parallel task that alternates between sending and receiving.

Maybe we can come up with something generic enough that can be shared? Then your block program could just import it and run it.

JJHackimoto:
Sounds good to me! Let me know the progress and if I can help with anything :)

JJHackimoto:
Hi! When can I expect to have something to test?

JJHackimoto:
Tested the new beta version today. Still have major issues. Most of the broadcasted data will never be received or gets received a few seconds after initially broadcasted. I've tried both with and without the issue_1806_lib observe_enable. Broadcasting seems to work fine, but it's observing that's the biggest issue right now.

My test includes a sequence where one hub will be waiting to receive a value, while another hub will start off by broadcasting a value upon startup. I've had to restart 3-4 times until the value is finally received on the other hub. I'm also using the LED color on the hub to know if the value was sent, which it is, every time.

PR #270, opened 2024-10-25
pbio/imu: Implement 3D attitude and fusion of IMU data.

laurensvalk:
This is an exciting update that implements 3D attitude estimation for the gyro with fusion of accelerometer data.

The main changes:
- Implement 3D attitude estimation.
- Implement fusion with accelerometer data. 
- Improve heading estimation for `hub.imu.heading()`. Now heading will still be close to the correct value when you drive across ramps or lift it from the table.
- Implement `hub.imu.tilt()` to use the new fused gyro data. This should be much less noisy than the pure accelerometer result, and it is no longer affected by horizontal or vertical acceleration.
- Implement `hub.imu.orientation()` which gives you the rotation matrix of your robot. This is almost always more useful than some kind of (non-)singular angular representation. We'll add some examples to show why this is useful. You could always extract your favorite angular representation from the rotation matrix elements.
- Add calibration options for gyro and accelerometer offset and bias. 

-----------------

Things that still need to be done:
- Replace the (still temporary) `hub.imu.update_heading_calibration()` with an interactive routine that estimates the 3D calibration constants.
- Make improvements for robustness. Since the new additions depend on user calibration constants saved on flash, we'll want to make sure nothing can go wrong (like zero division or singularities for some attitudes) if the user sets unexpected calibration values.
- See how well everything works in practice in different use cases.
- Verify that the 3D heading estimation is no worse than the pure 1D integration. If it is, we need to consider making that separately available.

-----------------

Other notes:
- I did not use a standard fusion algorithm. Since we don't have any magnetic sensors and since most LEGO robots don't fly, I think we can benefit from a simpler algorithm that makes use of some of the assumptions that we can make about LEGO robots. Roughly, it works like this:
- The angular velocity bias is currently still estimated independently as a long term moving average. Since we can make the robot sit still (unlike a drone mid flight), this works quite well.
- Attitude estimation works by converting the angular velocity to the quaternion rate of change and integrating it over time.
- Fusion happens by steering the estimated gravity vector towards the measured gravity vector when the hub is (close to) stationary, as if connecting them by a virtual spring.
- The initial quaternion is set to align with the first available gravity sample (with an arbitrary initial heading).


laurensvalk:
To get this over the line, perhaps we should restore 1D heading and also keep the original heading correction setting available since people are using it.

We can make the 3D-based heading the default as it gets more testing organically, instead of having it sit here as a pull request for a long time.

PR #271, opened 2024-10-27
bricks/_common/mpconfigport: enable unicode support

dlech:
Enable `MICROPY_PY_BUILTINS_STR_UNICODE` on hubs with `PYBRICKS_OPT_EXTRA_MOD`.

Without this, there is not any difference between str and bytes, which can make it hard to work with multi-byte characters.

This takes up about 1700 bytes on Technic hub.

Related discussion: https://github.com/orgs/pybricks/discussions/1882#discussioncomment-10962184


laurensvalk:
Will this have any effect on functions or methods that currently accept both bytes and strings? (I have not looked into this yet, just wondering).

dlech:
> Will this have any effect on functions or methods that currently accept both bytes and strings? 

No, anything that accepts both would be using the buffer protocol, so nothing would change there.

laurensvalk:
Merged and enabled on hubs with `PYBRICKS_OPT_EXTRA_LEVEL2`

PR #272, opened 2024-10-28
Update storage_settings.c to fix #1905

BertLindeman:
default documented imu.acceleration_threshold is 250. The installed threshold is 2500. Reducing to 250 as documented. Related to issue #1905


laurensvalk:
Thank you, but it was set to `2500` in https://github.com/pybricks/support/issues/1105, with the documentation still to be updated.

PR #273, opened 2024-11-10
pybricks.ev3devices.ColorSensor: Use blocking read calls

JakubVanek:
Other EV3 sensors in this module use the `pb_type_device_get_data_blocking()` method. I think that using `pb_type_device_get_data()` as it is used (only) in the color sensor class the sensor will not work for reading non-default modes.

I have found this by accident - I wanted to know if Pybricks waits for a sensor to change its mode. I think this is what happens for most EV3 sensors and it is only disabled for the EV3 color sensor.

However, I unfortunately have no way to test this and also I'm not entirely sure the bug is there.

### DIFF
--- a/pybricks/ev3devices/pb_type_ev3devices_colorsensor.c
+++ b/pybricks/ev3devices/pb_type_ev3devices_colorsensor.c
@@ -29,7 +29,7 @@ static mp_obj_t ev3devices_ColorSensor_make_new(const mp_obj_type_t *type, size_
 
 // pybricks.ev3devices.ColorSensor.color
 static mp_obj_t get_color(mp_obj_t self_in) {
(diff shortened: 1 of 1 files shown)


laurensvalk:
Thank you. The EV3 build hasn't been tested much lately, and won't be continued in its current form. The `pybricks.ev3devices` has not been updated to be `async` compatible.

I'm currently working on a bare metal version of Pybricks for EV3. It will probably get a rewrite of `pybricks.ev3devices` to be a bit more like `pybricks.pupdevices` and be `async` compatible.

JakubVanek:
Oh, I see. In that case, feel free to close this PR.

laurensvalk:
I think we can still merge it to make sure the legacy EV3 version is still complete. Thanks!

PR #274, opened 2024-11-16
pybricks.common.charger: change return value of connected()

dlech:
Change return value of `connected()` property from `bool` to `int` using the value of `pbdrv_usb_get_bcd()`. This will allow pro users to be able to tell if they have a "nonstandard" charger that could prevent proper charging.

Related discussion: https://github.com/orgs/pybricks/discussions/1872#discussioncomment-11278180

Alternative to consider: We could have the battery status light reflect this status somehow to make it even easier for users. Although having programmatically accessible in addition to that still seems useful for advanced use cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,12 +22,17 @@
 - Re-implemented `hub.imu.heading()` to use optionally use the projection of 3D
   orientation to improve performance when the hub is lifted off the ground.
   The 1D-based heading remains the default for now.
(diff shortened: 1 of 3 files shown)


laurensvalk:
This looks good to me. Thanks!

Could you add a notice to the `CHANGELOG`? I will be going through that one later to make the documentation updates to match.

laurensvalk:
Merged, thanks!

PR #275, opened 2024-12-02
pbio/drv/uart: Refactor async read and write

laurensvalk:
**Motivation**

I was working on a UART driver for a new device (not part of this PR) and wanted to take on some of the open tasks before adding another driver with the same open tasks/issues:

_Get rid of the autostart for the UART process_

Changing the initialization was easy enough, but then I realized that this wasn't actually a process like we have in the other drivers. The UART drivers are only used as part of other drivers, like the `legodev` driver (for LUMP devices), which would be running its own event-driven process.

The UART processes were there for a different reason, which was to postpone handling of IRQ data to the event loop, to update the transmission state. Since the UART drivers are only used as part of other processes, we can drive the poll handling from there. In practice, this amounts to just updating the transmission state before the completion check so no explicit separate poll step is necessary. (Which also fixes the subtle bug mentioned below.)

_Get rid of event broadcasts_

The UART poll handlers were broadcasting events to every other process even though only the processes that use UARTS need polling. This was a longstanding `TODO` in the drivers. This is now replaced with configurable callback. There is one callback per driver since in the general case different device drivers may use UARTS (not just `legodev`).

Potentially this is a change to watch out for, though in a good way. If there were any processes that were inadvertently relying on these unrelated broadcast events to move things along, these may now reveal themselves.

_Better understanding of read and write tasks_

We used to have a `read_begin` function and a `read_end` function we kept calling in a loop from within a protothread:

```c
#define PBIO_PT_WAIT_READY(pt, expr) PT_WAIT_UNTIL((pt), (expr) != PBIO_ERROR_AGAIN)

    PBIO_PT_WAIT_READY(&ludev->pt, ludev->err = pbdrv_uart_read_begin(ludev->uart, ludev->rx_msg, 1, 10));
    if (ludev->err != PBIO_SUCCESS) {
        DBG_ERR(ludev->last_err = "UART Rx error during baud");
        PT_EXIT(&ludev->pt);
    }

    PBIO_PT_WAIT_READY(&ludev->pt, ludev->err = pbdrv_uart_read_end(ludev->uart));
    // ludev->err is then processed below
```

This separation made the implementation of `pbdrv_uart_read_begin` and `pbdrv_uart_read_end` somewhat hard to follow. For this PR I have combined these into a single protothread that we can spawn.

```c
PT_SPAWN(&ludev->pt, &ludev->read_pt, pbdrv_uart_read(&ludev->read_pt, ludev->uart, ludev->rx_msg, 1, 10, &ludev->err));
    // ludev->err is then processed below
```

**Fixing a subtle bug with single-byte reads**

This also fixes a subtle bug where using begin/end to read a single byte always required a new byte, even if that was not the byte that was going to be read. I noticed this when I used the existing implementation for a MicroPython REPL. If more than one byte was pasted, it would be received into the ringbuffer by the IRQ but reading from it would be `(N-1)` steps behind, so you needed to type more characters before the old ones would be processed by `pbdrv_uart_read_begin/end`

Now, reading a byte will read an existing byte if it is available without requiring additional polling. Flush is now implemented on all platforms and can be called if a fresh byte is needed, so there is no loss of generality.

I almost don't see how this could have worked before, but perhaps the event broadcasting (and slightly more than necessary, one too many for each write) was providing enough ticks to move things along.

**Build size**
This also saves 360 bytes on Move Hub, which is a nice bonus.

------

This squashes the following intermediate commits:

------

pbio/drv/uart: Don't broadcast uart events to all processes.

Instead poll subscribers, which is just the legodev
process for now. This can be expanded to poll the
specific uart instead of all of them.

This also drops the uart processes, which were not
doing anything, saving some code size.

-----

pbio/drv/uart: Replace begin/end with single protothread.

UART read and write always used begin/end sequentially, each wrapped
in a wait for err != PBIO_ERROR_AGAIN. This replaces the begin/end
pattern with one awaitable protothread.

This makes the code easier to follow and reduces code size.

----

pbio/drv/uart: Set callback per device.

Not all devices may be used by the legodev driver.


laurensvalk:
We we are doing this, let's also drop the autostart dependency for the one remaining driver that used it: ADC.

laurensvalk:
Move Hub savings: -400 bytes
Technic Hub savings: -180 bytes

lib/pbio/drv/bluetooth/bluetooth_stm32_cc2640.c
```diff
@@ -2277,6 +2271,14 @@ PROCESS_THREAD(pbdrv_bluetooth_spi_process, ev, data) {
 
     PROCESS_BEGIN();
 
+    #if DEBUG_ON_LAST_UART_PORT
+    // Wait for the UART to be ready for debugging.
+    while (pbdrv_ioport_debug_uart_init(uart_poll_callback) != PBIO_SUCCESS) {
+        etimer_set(&timer, 100);
+        PROCESS_WAIT_EVENT_UNTIL(ev == PROCESS_EVENT_TIMER && etimer_expired(&timer));
```
dlech, line 2278: Why delay boot for 100 ms?

It should be fine to just yield here and skip the timer.

lib/pbio/drv/ioport/ioport_debug_uart.h
```diff
@@ -9,28 +9,18 @@
...
 #include <pbdrv/config.h>
+#include <pbdrv/uart.h>
 
 #if PBDRV_CONFIG_IOPORT_DEBUG_UART
 
-extern struct pt printf_thread;
+extern char debug_buffer[];
+extern struct pt debug_printf_thread;
+extern pbdrv_uart_dev_t *debug_uart;
+extern pbio_error_t debug_err;
```
dlech, line 19: Please add a namespace prefix to public symbols.

lib/pbio/drv/legodev/legodev_pup.c
```diff
@@ -343,6 +349,7 @@ static PT_THREAD(pbdrv_legodev_pup_thread(ext_dev_t * dev)) {
     while (true) {
 
         // Initially assume nothing is connected.
+        legodev_pup_disable_uart(dev->pins);
```
dlech, line 352: I didn't dig into it, but this looks suspect to me. It seems like this could interfere with the device detection code that expects pins to be in a certain state.

lib/pbio/drv/legodev/legodev_pup.c
```diff
@@ -431,7 +447,15 @@ void pbdrv_legodev_init(void) {
         pbio_dcmotor_get_dcmotor(legodev, &dcmotor);
         legodev->ext_dev->uart_dev = pbdrv_legodev_pup_uart_configure(legodev_data->ioport_index, port_data->uart_driver_index, dcmotor);
 
+        // legodev driver is started after all other drivers, so we
+        // assume that we do not need to wait for this to be ready.
```
dlech, line 451: I would add an assert statement here so that we could at least catch this in debug builds to catch future changes that might break this assumption.

lib/pbio/drv/legodev/legodev_pup.c
```diff
@@ -405,6 +413,14 @@ PROCESS_THREAD(pbio_legodev_pup_process, ev, data) {
     PROCESS_END();
 }
 
+/**
+ * We get notified when the uart driver has completed sending or receiving data.
+ */
+static void uart_poll_callback(pbdrv_uart_dev_t *uart) {
+    // REVISIT: Only need to poll the specified uart device.
```
dlech, line 420: We could add a flag that gets set here so that we only poke the uart instances that need it in the poll handler.

lib/pbio/drv/legodev/legodev_pup_uart.c
```diff
@@ -743,13 +731,8 @@ static PT_THREAD(pbdrv_legodev_pup_uart_synchronize_thread(pbdrv_legodev_pup_uar
     pbdrv_uart_flush(ludev->uart);
 
     // read one byte to check for ACK
-    PBIO_PT_WAIT_READY(&ludev->pt, ludev->err = pbdrv_uart_read_begin(ludev->uart, ludev->rx_msg, 1, 10));
-    if (ludev->err != PBIO_SUCCESS) {
-        DBG_ERR(ludev->last_err = "UART Rx error during baud");
-        PT_EXIT(&ludev->pt);
-    }
+    PT_SPAWN(&ludev->pt, &ludev->read_pt, pbdrv_uart_read(&ludev->read_pt, ludev->uart, ludev->rx_msg, 1, 10, &ludev->err));
```
dlech, line 734: Add another macro if this timeout is different than `EV3_UART_IO_TIMEOUT` with a comment that explains the difference.

lib/pbio/drv/uart/uart_stm32f0.c
```diff
@@ -202,60 +191,29 @@ void pbdrv_uart_stm32f0_handle_irq(uint8_t id) {
...
-            // TODO: this should only be sent once per write_begin
-            process_post(PROCESS_BROADCAST, PROCESS_EVENT_COM, NULL);
+        if (uart->poll_callback) {
+            uart->poll_callback(&uart->uart_dev);
         }
     }
 }
 
-static void handle_exit(void) {
+// Currently not used
```
dlech, line 208: Surround it with `#if 0` or delete it if not used. Don't make it non-static.

lib/pbio/drv/uart/uart_stm32l4_ll_dma.c
```diff
@@ -357,21 +359,17 @@ void pbdrv_uart_stm32l4_ll_dma_handle_uart_irq(uint8_t id) {
...
 }
 
-static void handle_poll(void) {
-    // TODO: only broadcast when read or write is complete
-    process_post(PROCESS_BROADCAST, PROCESS_EVENT_COM, NULL);
-}
-
-static void handle_exit(void) {
+// Currently not used
+void handle_exit(void) {
```
dlech, line 372: Prefer `#if 0`.

dlech (commented):
Nice one. Just a few suggestions.

laurensvalk:
Thanks for your review! 

I will update it as suggested, but not merge it just yet. I think we can do that after we make the next release so we can avoid breaking anything.

I'm also inclined to maybe drop that (unused) uart debug port interface as it is, and come up with something more useful in a future iteration. I'd like it to work from anywhere instead of from other processes. And also from the first port instead of the last (or any port) so it will work better with EV3.

lib/pbio/drv/legodev/legodev_pup.c
laurensvalk: Currently the behavior is/was:

```
boot
while True:
    run device connection manager until unknown uart detected
    legodev_pup_enable_uart(dev->pins);       // <--- there was an enable but never a disable
    spawn LUMP uart thread, which runs until failure
```

I think the uart (buffer) is disabled initially, so in the current case the second pass was not the same as the first. Setting it back to disabled was intended to take care of your suspicion :smile: 

On closer look, it seems that the DCM immediately sets the buffer state again so it doesn't currently seem to make any difference.

The goal is to eventually get this into a multimodal port interface, where `legodev` is one of several options.

laurensvalk:
Merging this into the pybricks-v4.x branch. This can eventually be rebased onto (and become) the master branch after the next stable release.

PR #277, opened 2024-12-16
pbio/platform/prime_hub: Enable slots.

laurensvalk:
Support for basic multi-program slots has been added a while ago, but it is not enabled by default while we wait on related featured to be completed.

This enables it for those who wish to try it out early.

**Try it out**
- Download the Prime Hub firmware file using the link in the post below.
- Install that file [using these instructions](https://pybricks.com/learn/getting-started/install-pybricks/#trying-the-nightly-version-optional).
- Code as usual using https://beta.pybricks.com/
- On the hub, choose slot 1 to 5 using the left/right buttons.
- A new program is downloaded to the selected slot.

**Caveats**
- There is no file manager yet. To delete programs, just download an empty program to that slot.
- Upgrading to a new firmware will erase the programs from the hub.
- All programs are kept in memory at once. If you write 5 giant programs, it will have a bit less memory remaining to run them.

### DIFF
--- a/lib/pbio/platform/prime_hub/pbsysconfig.h
+++ b/lib/pbio/platform/prime_hub/pbsysconfig.h
@@ -10,7 +10,7 @@
 #define PBSYS_CONFIG_BLUETOOTH                      (1)
 #define PBSYS_CONFIG_BLUETOOTH_TOGGLE               (1)
 #define PBSYS_CONFIG_BLUETOOTH_TOGGLE_BUTTON        (512) // PBIO_BUTTON_RIGHT_UP, but enum value cannot be used here.
-#define PBSYS_CONFIG_HMI_NUM_SLOTS                  (0)
+#define PBSYS_CONFIG_HMI_NUM_SLOTS                  (5)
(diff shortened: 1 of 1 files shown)


laurensvalk:
This no longer works.

Since updating Pybricks Code, it looks like we are now always sending 0 for the slot payload, so using the "current slot" never hits.

```c
        case PBIO_PYBRICKS_COMMAND_START_USER_PROGRAM: {
            if (size > 2) {
                return PBIO_PYBRICKS_ERROR_VALUE_NOT_ALLOWED;
            }
            // Use payload as program ID, otherwise use active user slot.
            return pbio_pybricks_error_from_pbio_error(
                pbsys_main_program_request_start((size == 2 ? data[1] : pbsys_hmi_get_selected_program_slot()), PBSYS_MAIN_PROGRAM_START_REQUEST_TYPE_REMOTE));
```

 I suppose it is not critical since we are not shipping this with the current firmware.

laurensvalk:
> This no longer works. Since updating Pybricks Code, it looks like we are now always sending 0 for the slot payload, so using the "current slot" never hits.

This can be solved with https://github.com/pybricks/pybricks-code/pull/2343

PR #279, opened 2024-12-28
pybricks.common.Speaker: fix negative duration in beep()

dlech:
Return immediately if duration is negative as per documentation.

This works the same as with motor methods, with `wait=False` in that the method is not awaitable when duration is negative.

Issue: https://github.com/pybricks/support/issues/1996

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,12 +35,14 @@
 - Fixed `DriveBase.angle()` getting an incorrectly rounded gyro value, which
   could cause `turn(360)` to be off by a degree ([support#1844]).
 - Fixed `hub` silently ignoring non-orthogonal base axis when it should raise.
+- Fixed not handling negative duration in `Speaker.beep()` ([support#1996]).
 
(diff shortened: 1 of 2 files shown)


dlech:
Tested with:

```python

hub = PrimeHub()

while True:
    hub.speaker.beep(100, -1)
    wait(100)
    hub.speaker.beep(0, -1)
    wait(100)
```

laurensvalk:
Thank you! This looks good.

Can you also test it with an async program?

dlech:
Works with async like this: 

```python
from pybricks.tools import wait, run_task

hub = PrimeHub()

async def main():
    while True:
        hub.speaker.beep(100, -1)
        await wait(100)
        hub.speaker.beep(0, -1)
        await wait(100)

run_task(main())
```

laurensvalk:
Thank you!

czuryk:
Thanks for quick resolving the issue!

laurensvalk:
@czuryk If you want to try the fix, you can do so with [these instructions](https://pybricks.com/learn/getting-started/install-pybricks/#trying-the-nightly-version-optional).

PR #280, opened 2025-01-08
hack: Bad process

laurensvalk:
This should not be merged, but serves to document and reproduce an odd bug that I am seeing locally.

**Context**
I was working on a major revision of the port subsystem, and was running into issues with etimers and polling processes.

After some time, I was able to boil it down to a very small example that also builds from the current master branch. This makes it seem independent from the current work (which is not submitted here).

**To reproduce**
Build either of the two commits from this PR. If everything is working, you should see `Hello World` printed in Pybricks Code  every second after connecting.

- In the first commit, nothing is printed at all
- In the second, if we ignore the `ev` check, it doesn't work at first but it starts when kicked externally. For example, start the REPL (state changes trigger a event broadcast).

**Caveats**
This may be due to unrelated undefined behavior somewhere else. It seems to vary from build to build. At one point, adding comments made it (dis)appear, presumably because that changes the LC value of the switches. Similarly, experimenting with variations of `etimer_set` such as `etimer_reset` don't make the difference, even if they sometimes appear to make it work, but presumably that is the same as above.

This is one reason to create the CI build --- to see if it reproduces there too. EDIT: It does. I suppose that is a good thing.

### DIFF
--- a/lib/pbio/drv/ioport/ioport_pup.c
+++ b/lib/pbio/drv/ioport/ioport_pup.c
@@ -70,7 +70,10 @@ PROCESS_THREAD(test_process, ev, data) {
 
         etimer_set(&timer, 1000);
 
-        PROCESS_WAIT_EVENT_UNTIL(ev == PROCESS_EVENT_TIMER && etimer_expired(&timer));
+        // And with this change it still doesn't work after connect, but it
+        // does get going when you kick it with an event like starting REPL.
+
+        PROCESS_WAIT_EVENT_UNTIL(/*ev == PROCESS_EVENT_TIMER &&*/ etimer_expired(&timer));
 
         printf("Hello, world\n");
     }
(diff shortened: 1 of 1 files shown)


dlech:
If this isn't working and we have ruled out overflowing the event queue, then it kind smells like possibly a memory corruption bug to me.

One way to possibly debug that could be to keep turning off non-essential drivers (PBDRV_CONFIG) and subsystems (PBIO_CONFIG) until this starts working to narrow it down to what is interfering with proper operation.

laurensvalk:
Thank you.

This seems to be a debug print issue rather than a process issue. I went back to apply this patch to previous released and noticed it started occurring since b460d6dc4706b9ade1c2f5bb06d3fdb07c07c343, which is plausible, because running the event loop from a process is a bad idea.

All the more reason to finish up direct UART access for debugging :smile:

laurensvalk:
Closed as non-issue.

PR #298, opened 2025-03-13
pbio/os: Prototype simpler and unified async OS.

laurensvalk:
Overall, the goals are to reduce complexity, reduce code size, and make it more cross platform.

In this PR, the new approach still drives the original `pbio` event loop as well, so we can convert one process at a time without breaking things. The charger process is done here as a simple test to illustrate what it looks like. Once all processes have been converted and everything is working, we could drop the Contiki dependency.

Not everything we need is implemented yet, but it is already easy to see that `pbio/os` is much easier to get into than the full contiki library. Mainly because we only implement what we need, but also due to the simplifications introduced below.

- Returning errors from protothreads is currently cumbersome. This modification allows protothreads to return error codes. `PBIO_ERROR_AGAIN` essentially takes the role of `PT_YIELDED` / `PT_WAITING`. Everything else indicates completion with the respective error or success.

- Contiki partially distinguishes (but not really) between waiting and yielding. In practice, both yield and the latter yields at least once. And then both macros exist separately for processes and protothreads so you end up with four variants that are almost the same but subtly different. I think we can do everything with a single set of `AWAIT` macros.

- The contiki event timers are more complicated than we need for our use case. Currently, a timer tick first polls the timer process and then posts N events depending on the number of active timers. This is not be safe if we have more timer events than fit in the event queue. Also, etimers use the global `PROCESS_CURRENT` which requires hacks such as `PROCESS_CONTEXT_BEGIN` when accessing them from elsewhere.

- Instead of using various events and polling individual processes throughout, I think we may be able to work with just a single `pbio_os_request_poll()` function. Then all processes will check their current wait condition. This should not be that much more computationally expensive than checking the `needs_poll` flag on each process (and it may be less due to overall simplicity).

- We currently use broadcasting between processes in a few places, which can be hard to follow and risks overflowing the event queue as already stated in a few existing `REVISIT`s around the code. We have recently eliminated this for the uart process. We should be able to do this for the status change too.

- We can move some of the logic for handling pending events and entering sleep mode to `pbio` instead of having it in `mphalport`. Aside from hooks such as `enable_irq` it becomes platform agnostic. This also brings the virtual hub a bit closer to the embedded hubs. It also unifies the EV3 and NXT with Powered Up.

### DIFF
--- a/bricks/_common/micropython.c
+++ b/bricks/_common/micropython.c
@@ -9,6 +9,7 @@
 
 #include <pbio/button.h>
 #include <pbio/main.h>
+#include <pbio/os.h>
 #include <pbio/util.h>
 #include <pbio/protocol.h>
 #include <pbsys/main.h>
@@ -39,14 +40,9 @@
 // Implementation for MICROPY_EVENT_POLL_HOOK
 void pb_event_poll_hook(void) {
 
-    // Drive pbio event loop.
-    while (pbio_do_one_event()) {
-    }
-
     mp_handle_pending(true);
 
-    // Platform-specific code to run on completing the poll hook.
-    pb_event_poll_hook_leave();
+    pbio_os_run_while_idle();
 }
 
 // callback for when stop button is pressed in IDE or on hub
--- a/bricks/_common/micropython.c
+++ b/bricks/_common/micropython.c
@@ -40,9 +40,12 @@
 // Implementation for MICROPY_EVENT_POLL_HOOK
 void pb_event_poll_hook(void) {
 
+    while (pbio_os_run_processes_once()) {
+    }
+
     mp_handle_pending(true);
 
-    pbio_os_run_while_idle();
+    pbio_os_run_processes_and_wait_for_event();
 }
 
 // callback for when stop button is pressed in IDE or on hub
--- a/bricks/_common_stm32/mphalport.c
+++ b/bricks/_common_stm32/mphalport.c
@@ -24,20 +24,6 @@ void pb_stack_get_info(char **sstack, char **estack) {
     *estack = (char *)&_estack;
 }
 
-pbio_os_irq_flags_t pbio_os_hook_disable_irq(void) {
-    mp_uint_t flags = __get_PRIMASK();
-    __disable_irq();
-    return flags;
-}
-
-void pbio_os_hook_enable_irq(pbio_os_irq_flags_t flags) {
-    __set_PRIMASK(flags);
-}
-
-void pbio_os_hook_wait_for_interrupt(pbio_os_irq_flags_t flags) {
-    __WFI();
-}
-
 // using "internal" pbdrv variable
 extern volatile uint32_t pbdrv_clock_ticks;
 
--- a/bricks/_common_stm32/mphalport.c
+++ b/bricks/_common_stm32/mphalport.c
@@ -24,17 +24,17 @@ void pb_stack_get_info(char **sstack, char **estack) {
     *estack = (char *)&_estack;
 }
 
-uint32_t pbio_os_hook_disable_irq(void) {
+pbio_os_irq_flags_t pbio_os_hook_disable_irq(void) {
     mp_uint_t flags = __get_PRIMASK();
     __disable_irq();
     return flags;
 }
 
-void pbio_os_hook_enable_irq(uint32_t flags) {
+void pbio_os_hook_enable_irq(pbio_os_irq_flags_t flags) {
     __set_PRIMASK(flags);
 }
 
-void pbio_os_hook_wait_for_interrupt(void) {
+void pbio_os_hook_wait_for_interrupt(pbio_os_irq_flags_t flags) {
     __WFI();
 }
 
--- a/bricks/_common_stm32/mphalport.c
+++ b/bricks/_common_stm32/mphalport.c
@@ -24,17 +24,18 @@ void pb_stack_get_info(char **sstack, char **estack) {
     *estack = (char *)&_estack;
 }
 
-void pb_event_poll_hook_leave(void) {
-    // There is a possible race condition where an interrupt occurs and sets the
-    // Contiki poll_requested flag after all events have been processed. So we
-    // have a critical section where we disable interrupts and check see if there
-    // are any last second events. If not, we can call __WFI(), which still wakes
-    // up the CPU on interrupt even though interrupts are otherwise disabled.
-    mp_uint_t state = disable_irq();
-    if (!process_nevents()) {
-        __WFI();
-    }
-    enable_irq(state);
+uint32_t pbio_os_hook_disable_irq(void) {
+    mp_uint_t flags = __get_PRIMASK();
+    __disable_irq();
+    return flags;
+}
+
+void pbio_os_hook_enable_irq(uint32_t flags) {
+    __set_PRIMASK(flags);
+}
+
+void pbio_os_hook_wait_for_interrupt(void) {
+    __WFI();
 }
 
 // using "internal" pbdrv variable
--- a/bricks/virtualhub/mp_port.c
+++ b/bricks/virtualhub/mp_port.c
@@ -94,28 +94,29 @@ void pb_event_poll_hook(void) {
     pbio_os_run_while_idle();
 }
 
-// This is used instead of the uint32_t flags used in embedded builds.
-static sigset_t global_origmask;
-
-uint32_t pbio_os_hook_disable_irq(void) {
+pbio_os_irq_flags_t pbio_os_hook_disable_irq(void) {
     sigset_t sigmask;
     sigfillset(&sigmask);
-    pthread_sigmask(SIG_SETMASK, &sigmask, &global_origmask);
-    return 0;
+
+    sigset_t origmask;
+    pthread_sigmask(SIG_SETMASK, &sigmask, &origmask);
+    return origmask;
 }
 
-void pbio_os_hook_enable_irq(uint32_t flags) {
-    pthread_sigmask(SIG_SETMASK, &global_origmask, NULL);
+void pbio_os_hook_enable_irq(pbio_os_irq_flags_t flags) {
+    sigset_t origmask = (sigset_t)flags;
+    pthread_sigmask(SIG_SETMASK, &origmask, NULL);
 }
 
-void pbio_os_hook_wait_for_interrupt(void) {
+void pbio_os_hook_wait_for_interrupt(pbio_os_irq_flags_t flags) {
     struct timespec timeout = {
         .tv_sec = 0,
         .tv_nsec = 100000,
     };
     // "sleep" with "interrupts" enabled
+    sigset_t origmask = flags;
     MP_THREAD_GIL_EXIT();
-    pselect(0, NULL, NULL, NULL, &timeout, &global_origmask);
+    pselect(0, NULL, NULL, NULL, &timeout, &origmask);
     MP_THREAD_GIL_ENTER();
 }
 
--- a/bricks/virtualhub/mp_port.c
+++ b/bricks/virtualhub/mp_port.c
@@ -89,9 +89,12 @@ void pb_virtualhub_port_deinit(void) {
 // Implementation for MICROPY_EVENT_POLL_HOOK
 void pb_event_poll_hook(void) {
 
+    while (pbio_os_run_processes_once()) {
+    }
+
     mp_handle_pending(true);
 
-    pbio_os_run_while_idle();
+    pbio_os_run_processes_and_wait_for_event();
 }
 
 pbio_os_irq_flags_t pbio_os_hook_disable_irq(void) {
--- a/bricks/virtualhub/mp_port.c
+++ b/bricks/virtualhub/mp_port.c
@@ -14,6 +14,8 @@
 
 #include <contiki.h>
 
+#include "pbio_os_config.h"
+
 #include <pbio/main.h>
 #include <pbio/os.h>
 #include <pbsys/core.h>
--- a/bricks/virtualhub/mp_port.c
+++ b/bricks/virtualhub/mp_port.c
@@ -15,6 +15,7 @@
 #include <contiki.h>
 
 #include <pbio/main.h>
+#include <pbio/os.h>
 #include <pbsys/core.h>
 #include <pbsys/program_stop.h>
 #include <pbsys/status.h>
@@ -72,7 +73,8 @@ void pb_virtualhub_port_init(void) {
     pbsys_init();
 
     pbsys_status_set(PBIO_PYBRICKS_STATUS_USER_PROGRAM_RUNNING);
-    while (pbio_do_one_event()) {
+
+    while (pbio_os_run_processes_once()) {
     }
 
     pb_package_pybricks_init(true);
@@ -84,62 +86,44 @@ void pb_virtualhub_port_deinit(void) {
     pb_package_pybricks_deinit();
 }
 
-// MICROPY_VM_HOOK_LOOP
-void pb_virtualhub_poll(void) {
-    while (pbio_do_one_event()) {
-    }
-}
+// Implementation for MICROPY_EVENT_POLL_HOOK
+void pb_event_poll_hook(void) {
 
-// MICROPY_EVENT_POLL_HOOK
-void pb_virtualhub_event_poll(void) {
-start:
     mp_handle_pending(true);
 
-    int events_handled = 0;
+    pbio_os_run_while_idle();
+}
 
-    while (pbio_do_one_event()) {
-        events_handled++;
-    }
+// This is used instead of the uint32_t flags used in embedded builds.
+static sigset_t global_origmask;
 
-    // If there were any pbio events handled, don't sleep because there may
-    // be something waiting on one of the events that was just handled.
-    if (events_handled) {
-        return;
-    }
-
+uint32_t pbio_os_hook_disable_irq(void) {
     sigset_t sigmask;
     sigfillset(&sigmask);
+    pthread_sigmask(SIG_SETMASK, &sigmask, &global_origmask);
+    return 0;
+}
 
-    // disable "interrupts"
-    sigset_t origmask;
-    pthread_sigmask(SIG_SETMASK, &sigmask, &origmask);
+void pbio_os_hook_enable_irq(uint32_t flags) {
+    pthread_sigmask(SIG_SETMASK, &global_origmask, NULL);
+}
 
-    if (process_nevents()) {
-        // something was scheduled since the event loop above
-        pthread_sigmask(SIG_SETMASK, &origmask, NULL);
-        goto start;
-    }
-
+void pbio_os_hook_wait_for_interrupt(void) {
     struct timespec timeout = {
         .tv_sec = 0,
         .tv_nsec = 100000,
     };
-
     // "sleep" with "interrupts" enabled
     MP_THREAD_GIL_EXIT();
-    pselect(0, NULL, NULL, NULL, &timeout, &origmask);
+    pselect(0, NULL, NULL, NULL, &timeout, &global_origmask);
     MP_THREAD_GIL_ENTER();
-
-    // restore "interrupts"
-    pthread_sigmask(SIG_SETMASK, &origmask, NULL);
 }
-
 
 void pb_virtualhub_delay_us(mp_uint_t us) {
     mp_uint_t start = mp_hal_ticks_us();
 
     while (mp_hal_ticks_us() - start < us) {
-        pb_virtualhub_poll();
+        MICROPY_VM_HOOK_LOOP;
     }
 }
 
--- a/lib/pbio/include/pbio/os.h
+++ b/lib/pbio/include/pbio/os.h
@@ -269,7 +269,7 @@ struct _pbio_os_process_t {
 
 bool pbio_os_run_processes_once(void);
 
-void pbio_os_run_while_idle(void);
+void pbio_os_run_processes_and_wait_for_event(void);
 
 void pbio_os_request_poll(void);
 
--- a/lib/pbio/include/pbio/os.h
+++ b/lib/pbio/include/pbio/os.h
@@ -48,6 +48,9 @@
  *
  * The code in this file stems from Contiki's implementation of protothreads as
  * listed above. This adaptation changes the API to work better with Pybricks.
+ *
+ * Caution: it is conceptually very similar to Contiki but it cannot be used as
+ * a drop-in replacement.
  */
 
 #ifndef _PBIO_OS_H_
@@ -148,8 +151,6 @@ struct _pbio_os_process_t {
  *
  * This macro is used for declaring that a protothread ends. It must
  * always be used together with a matching ASYNC_BEGIN() macro.
- *
- * NB: In contrast to Contiki, this does not call ASYNC_INIT() before exiting.
  *
  * @param [in]  err    Error code to return.
  */
--- a/lib/pbio/include/pbio/os.h
+++ b/lib/pbio/include/pbio/os.h
@@ -208,7 +208,7 @@ struct _pbio_os_process_t {
 
 /**
  * Awaits two protothreads until one of them completes successfully or returns
- * an error.
+ * an error. There is no cleanup of the other protothread.
  *
  * @param [in]  host_state             State of the host protothread in which this macro is used.
  * @param [in]  sub_state_1            State of the first sub protothread.
@@ -250,6 +250,13 @@ struct _pbio_os_process_t {
         }                                       \
     } while (0)
 
+/**
+ * Yields the protothread until the specified timer expires.
+ * 
+ * @param [in]  state     Protothread state.
+ * @param [in]  timer     The timer to check.
+ * @param [in]  duration  The duration to wait for in milliseconds.
+ */
 #define PBIO_OS_AWAIT_MS(state, timer, duration)      \
     do {                                              \
         pbio_os_timer_set(timer, duration);           \
@@ -269,7 +276,7 @@ pbio_error_t pbio_port_process_none_thread(pbio_os_state_t *state, void *context
 
 void pbio_os_process_start(pbio_os_process_t *process, pbio_os_process_func_t func, void *context);
 
-void pbio_os_process_reset(pbio_os_process_t *process, pbio_os_process_func_t func);
+void pbio_os_process_init(pbio_os_process_t *process, pbio_os_process_func_t func);
 
 /**
  * Disables interrupts and returns the previous interrupt state.
--- a/lib/pbio/include/pbio/os.h
+++ b/lib/pbio/include/pbio/os.h
@@ -136,9 +136,12 @@ struct _pbio_os_process_t {
  * which the protothread runs. All C statements above the ASYNC_BEGIN()
  * invocation will be executed each time the protothread is scheduled.
  *
+ * The do_yield_now variable is used only to facilitate the AWAIT_ONCE macro.
+ * It shouldn't take any space if unused and optimizations are enabled.
+ *
  * @param [in]  state    Protothread state.
  */
-#define ASYNC_BEGIN(state) { char do_yield_now = 0; if (do_yield_now) {;} PB_LC_RESUME(state)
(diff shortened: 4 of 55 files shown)


laurensvalk:
Example usage:

```c
pbio_error_t pbio_os_wait_example(pbio_os_state_t *state, uint32_t duration) {

    static pbio_os_timer_t timer;

    ASYNC_BEGIN(state);

    printf("Going to wait for %d ms\n", duration);

    // We can return errors!
    if (duration > 2000) {
        return PBIO_ERROR_INVALID_ARG;
    }

    AWAIT_MS(state, &timer, duration);

    ASYNC_END(PBIO_SUCCESS);
}

pbio_error_t pbio_os_sub_example(pbio_os_state_t *state) {

    static pbio_os_state_t child;

    static uint32_t counter;

    pbio_error_t err;

    ASYNC_BEGIN(state);

    for (counter = 0; counter < 100; counter++) {

        AWAIT(state, &child, err = pbio_os_wait_example(&child, 100 * counter));

        if (err != PBIO_SUCCESS) {
            printf("Something bad happened!\n");
            return err;
        }
    }

    ASYNC_END(PBIO_SUCCESS);
}

// Outermost "process thread" looks the same but must only have the state argument
pbio_error_t pbio_os_example_thread(pbio_os_state_t *state) {

    static pbio_os_state_t child;

    pbio_error_t err;

    ASYNC_BEGIN(state);

    AWAIT(state, &child, err = pbio_os_sub_example(&child));
    if (err != PBIO_SUCCESS) {
        // Being able to do this is the addition.
        return err;
    }

    ASYNC_END(PBIO_SUCCESS);
}
```

laurensvalk:
Similarly, the following:

```c
PT_THREAD(pbdrv_uart_write(struct pt *pt, pbdrv_uart_dev_t *uart, uint8_t *msg, uint8_t length, uint32_t timeout, pbio_error_t *err)) {

    PT_BEGIN(pt);

    if (!msg || !length) {
        *err = PBIO_ERROR_INVALID_ARG;
        PT_EXIT(pt);
    }

    if (uart->tx_buf) {
        *err = PBIO_ERROR_BUSY;
        PT_EXIT(pt);
    }
...
}
```

Becomes:

```c
pbio_error_t pbdrv_uart_write(pbio_os_state_t *state, pbdrv_uart_dev_t *uart, uint8_t *msg, uint8_t length, uint32_t timeout) {

    ASYNC_BEGIN(state);

    if (!msg || !length) {
        return PBIO_ERROR_INVALID_ARG;
    }

    if (uart->tx_buf) {
        return PBIO_ERROR_BUSY;
    }
...
}
```

dlech:
I haven't had time yet to have a close look at this yet. If you can wait a bit longer, I can try to have a look this weekend.

laurensvalk:
Sure, no rush -- that's what PRs are for :slightly_smiling_face: Thanks!

BertLindeman:
I assume this is the correct place to report:
'disconnect from pc does not always switch the hub into "I want to pair" mode'

Running `ci-build-3798-v3.6.1-134-g8e25902f on 2025-03-19` on a TechnicHub with nothing on the ports.

All blinking I talk about here is blue.

To reproduce:
- press the hub button to activate the hub
- in the UI press connect and select the hu

- then repeat
  - press in the UI the disconnect button
  - wait for the hub to blink-blink-pause
    this occurs most of the times.
    But it does not always occur:

1. Sometimes the hub light goes off for a short time and then steady blue for a bit more than a second and then blink-blink-pause
2. Sometimes the hub stays in constant blue.
 
Situation 1. occurs more often: 1 in ~4 to 10 attempts
Situation 2. I only saw 3 times since yesterday.

Testing with bluetooth, so I often disconnect for BT to be stable.

I did not succeed in a more stable scenario.

No need to give this a priority, I just wanted to report it. 
Hardly any influence on my testing.

Bert

lib/pbio/src/os.c
```diff
@@ -47,15 +47,23 @@ void pbio_os_request_poll(void) {
...
 
 /**
- * Starts a process.
+ * Placeholder thread that does nothing.
  *
- * NB: Processes can be started only once. They cannot be restarted.
+ * @param [in] state The protothread state.
+ * @param [in] context The context.
+ */
+pbio_error_t pbio_port_process_none_thread(pbio_os_state_t *state, void *context) {
```
dlech, line 55: I don't understand this function name. But if it is just temporary, I guess it doesn't matter.

lib/pbio/include/pbio/os.h
```diff
@@ -176,6 +189,16 @@ struct _pbio_os_process_t {
         }                                       \
     } while (0)
 
+#define AWAIT_RACE(state, child1, child2, statement1, statement2)                    \
```
dlech, line 208: I would include an explanation that we have to be really careful about the fact that there is no cancellation/cleanup of the thread that lost the race.

dlech, line 208: Also, it seems like we would want to know who won the race.

lib/pbio/include/pbio/os.h
```diff
@@ -0,0 +1,217 @@
...
+            return PBIO_ERROR_AGAIN;             \
+        }                                        \
+    } while (0)
+
+
+#define AWAIT(state, child, statement)          \
+    do {                                        \
+        ASYNC_INIT((child));                    \
+        PB_LC_SET(state);                       \
+        if ((statement) == PBIO_ERROR_AGAIN) {  \
```
dlech, line 194: I guess this means that `statement` can never return anything besides `PBIO_ERROR_AGAIN` or `PBIO_SUCCESS`?  Other error will be ignored, which isn't ideal. Could fix this by getting rid of `do`/`while` and introduce a local `err` variable so that this can be an expression statement. Then we could write `err = AWAIT(...);` and actually handle errors.

lib/pbio/src/os.c
```diff
@@ -69,11 +77,27 @@ void pbio_os_start_process(pbio_os_process_t *process, pbio_os_process_func_t fu
...
-    process->func = func;
-    process->context = context;
     process->next = NULL;
+    process->context = context;
+
+    pbio_os_process_reset(process, func);
+}
+
+/**
+ * Resets an existing process to the initial state with a new function and context.
```
dlech, line 87: Description doesn't match. There isn't a context parameter.

dlech, line 87: Resetting a process sounds dangerous without some kind of cleanup/teardown of the current state. Might be worth mentioning that in the description.

dlech (commented):
This stuff is tricky to get right. We will probably introduce lots of bug/regressions by doing this. But in the end, it will be simpler/nicer. So probably worth it in the long run.

laurensvalk:
> This stuff is tricky to get right. We will probably introduce lots of bug/regressions by doing this. But in the end, it will be simpler/nicer. So probably worth it in the long run.

Thanks for the overal and specific feedback!

lib/pbio/include/pbio/os.h
laurensvalk: I'll respond to this one on main thread.

laurensvalk:
> I guess this means that `statement` can never return anything besides `PBIO_ERROR_AGAIN` or `PBIO_SUCCESS`?  Other error will be ignored, which isn't ideal. Could fix this by getting rid of `do`/`while` and introduce a local `err` variable so that this can be an expression statement. Then we could write `err = AWAIT(...);` and actually handle errors.

The intention is to use it like this:

```c
pbio_error_t example_thread_func(pbio_os_state_t *state) {

    pbio_error_t err;

    ASYNC_BEGIN(state);

    ...

    AWAIT(state, &child, err = pbdrv_uart_write(&child, device, data, len, timeout));
    if (err != PBIO_SUCCESS) {
        return err;
    }

    ...

    ASYNC_END(PBIO_SUCCESS);
}
```

The goal here is indeed precisely to handle errors. `PBIO_ERROR_AGAIN` is "reserved" for async functions, but everything else can be returned. Also, the final `ASYNC_END(PBIO_SUCCESS)` can return other errors.

The `AWAIT_RACE` function can be used in the same way (`err1 = ...., err2 = ...`) so you can find out which completed/errored.

> Then we could write `err = AWAIT(...);`

The example above mostly achieves this, but if there is a clever way to make the macros work differently, we could look at it. Right now they work just like Contiki.

bricks/_common_stm32/mphalport.c
```diff
@@ -24,17 +24,18 @@ void pb_stack_get_info(char **sstack, char **estack) {
...
-    // Contiki poll_requested flag after all events have been processed. So we
-    // have a critical section where we disable interrupts and check see if there
-    // are any last second events. If not, we can call __WFI(), which still wakes
-    // up the CPU on interrupt even though interrupts are otherwise disabled.
-    mp_uint_t state = disable_irq();
-    if (!process_nevents()) {
-        __WFI();
-    }
-    enable_irq(state);
+uint32_t pbio_os_hook_disable_irq(void) {
```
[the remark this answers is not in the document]
laurensvalk: Yep, that is the goal indeed, to get all of this away from MicroPython, including things like the _WFI so it can all be cross platform for when we make a non-MicroPython version of `pbio` such as potentially for the Raspberry Pi Build hat.

laurensvalk:
> Hmm... polling every processes every 1ms seems like it would make it harder to catch bugs where we aren't properly polling in response to other events.

Yes, this is one of the main design considerations we'll want to think about thoroughly before going ahead.

It simplifies processes a lot when everything is working, but debugging could be harder. Maybe we could have the auto-polling be optional for each thread. Then certain processes can independently (self-)poll as we used to do.

> It also means we have to be really careful about using timers in cases where we could be waiting on both a timer and another event a the same time.

This is one of the things that caught me out in the Contiki way of doing things -- if it got to a yield-once-then check, it could "miss" the event if it had just got it before getting there. In the new proposal, this should be simpler.

What cases are you thinking of that could now be trickier?

laurensvalk:
> Resetting a process sounds dangerous without some kind of cleanup/teardown of the current state. Might be worth mentioning that in the description.

Yeah, this is probably only to be used for certain special cases. It is something I'd like to use for the `ports`, which has its own handling of safely resetting.

Basically, it "detaches" the LEGO DCM+LUMP process when the port goes into a different mode. It could attach that "none_process" dummy process instead if that mode needs no process. This allows everything to be synchronously handled rather than having `process_exit` needing several cycles to await other things to end.

dlech:
> What cases are you thinking of that could now be trickier?

It's been to long, so I don't remember clearly. But I have a vague recollection that a few times during development I would forget to to put in the `ev == PROCESS_EVENT_TIMER` and only had the `etimer_expired(&timer)` and things wouldn't work quite right until I remembered to also check `ev`. But I never really dug into it to see what was actually happening that caused it not to work correctly.

laurensvalk:
That is one complexities I was hoping to address here. A contiki timer etimer is expired when it is not associated with a process: `et->p == PROCESS_NONE`. There are multiple ways that this can happen, such as when it hasn't been set, or has been unset elsewhere. Particularly with the `etimer` objects in our uart drivers. They would only work correctly if set from that very process, as `et->p` would be set the global `CURRENT_PROCESS()`.

The new timer is just the clock time with a duration.

laurensvalk:
Instead of `PBIO_ERROR_AGAIN`, maybe we can use (introduce) `PBIO_ERROR_IN_PROGRESS` uniquely for this so we don't confuse it with `PBIO_ERROR_AGAIN` in existing code.

dlech:
Why not just return a bool if there are only 2 possible return values?

laurensvalk:
Not sure I follow. The point is to have the ability to return error values from protothreads (all our existing error values). We just need one more value to indicate that there is no error to report yet. I used `EAGAIN` in this PR, but `EINPROGRESS` is the closer to the correct meaning, and ensures no existing code that already uses `EGAIN` gets confused (even if they should never be mixed).

dlech:
Everything I saw only returned `PBIO_ERROR_AGAIN` or `PBIO_SUCCESS` and you didn't want to pass through the error on AWAIT, so I assumed that we weren't passing any other errors either.

Adding a `PBIO_YIELD` makes sense though as this is not actually an error condition that should be propagated. Maybe even `PBIO_YIELD = -1` to have an easy way to differentiate it from errors - assuming that doesn't break the `-fshort-enums` optimization.

laurensvalk:
Errors are passed on everywhere, including in the `AWAIT` through an assignment expression (unless there is a way to make the macros work another way). The goal is to clean up functions such as:

```c
PT_THREAD(pbdrv_uart_write(struct pt *pt, pbdrv_uart_dev_t *uart, uint8_t *msg, uint8_t length, uint32_t timeout, pbio_error_t *err)) {

    PT_BEGIN(pt);

    if (!msg || !length) {
        *err = PBIO_ERROR_INVALID_ARG;
        PT_EXIT(pt);
    }

    if (uart->tx_buf) {
        *err = PBIO_ERROR_BUSY;
        PT_EXIT(pt);
    }

   ...
```

to:
```c
pbio_error_t pbdrv_uart_write(pbio_os_state_t *state, pbdrv_uart_dev_t *uart, uint8_t *msg, uint8_t length, uint32_t timeout) {

    PBIO_OS_ASYNC_BEGIN(state);

    if (!msg || !length) {
        return PBIO_ERROR_INVALID_ARG;
    }

    if (uart->tx_buf) {
        return PBIO_ERROR_BUSY;
    }

   ...
```

The `PBIO_OS_ASYNC_END(err)` can also return an error, which works well with states that should not nominally be reached as we do in many normal functions. (It just needs to be a macro to squeeze in the extra `}`, just like Contiki.

dlech:
> (unless there is a way to make the macros work another way

That is what I was trying to suggest in https://github.com/pybricks/pybricks-micropython/pull/298#discussion_r2009307839 :wink: 

If we drop the do/while, the `{}` can be used as an expression statement. 

```c
#define AWAIT(state, child, expression)  { \
    ASYNC_INIT((child));                   \
    PB_LC_SET(state);                      \
    pbio_error_t _err = expression;        \
    if (_err== PBIO_ERROR_AGAIN) {         \
        return PBIO_ERROR_AGAIN;           \
    }                                      \
    _err;                                  \
}
```

```c
pbio_error_t err;
...
err = AWAIT(some_protothread())
if (err)
    return err;
...
```

lib/pbio/include/pbio/os.h
```diff
@@ -0,0 +1,282 @@
...
+    } while (0)
+
+/**
+ * Awaits a protothread until it is done.
+ *
+ * @param [in]  state      Protothread state.
+ * @param [in]  child      Protothread state of the child.
+ * @param [in]  statement  The statement to await.
+ */
+#define AWAIT(state, child, statement)          \
```
dlech, line 190: Technically, this need to be an _expression_, not a _statement_. And the doc comment needs to say that the expression needs to evaluate to a `pbio_error_t`. Same applies to the `AWAIT_RACE` as well.

laurensvalk:
Very interesting! The expression statement is a nice way of defining a local `pbio_err_t` that I didn't think of.

bricks/virtualhub/mp_port.c
```diff
@@ -84,62 +86,44 @@ void pb_virtualhub_port_deinit(void) {
...
-    if (events_handled) {
-        return;
-    }
+// This is used instead of the uint32_t flags used in embedded builds.
+static sigset_t global_origmask;
 
+uint32_t pbio_os_hook_disable_irq(void) {
     sigset_t sigmask;
     sigfillset(&sigmask);
+    pthread_sigmask(SIG_SETMASK, &sigmask, &global_origmask);
```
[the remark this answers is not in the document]
laurensvalk: This was the original:

c
// MICROPY_EVENT_POLL_HOOK
void pb_virtualhub_event_poll(void) {
start:
    mp_handle_pending(true);

    int events_handled = 0;

    while (pbio_do_one_event()) {
        events_handled++;
    }

    // If there were any pbio events handled, don't sleep because there may
    // be something waiting on one of the events that was just handled.
    if (events_handled) {
        return;
    }

    sigset_t sigmask;
    sigfillset(&sigmask);

    // disable "interrupts"
    sigset_t origmask;
    pthread_sigmask(SIG_SETMASK, &sigmask, &origmask);

    if (process_nevents()) {
        // something was scheduled since the event loop above
        pthread_sigmask(SIG_SETMASK, &origmask, NULL);
        goto start;
    }

    struct timespec timeout = {
        .tv_sec = 0,
        .tv_nsec = 100000,
    };

    // "sleep" with "interrupts" enabled
    MP_THREAD_GIL_EXIT();
    pselect(0, NULL, NULL, NULL, &timeout, &origmask);
    MP_THREAD_GIL_ENTER();

    // restore "interrupts"
    pthread_sigmask(SIG_SETMASK, &origmask, NULL);
}


How can we implement the hooks for the Linux port such that it ends up something like the one below?

(Keeping it simply in the context of Contiki for the moment -- I'll replace it with the equivalent pbio_os stuff later).

c
// MICROPY_EVENT_POLL_HOOK
void pb_virtualhub_event_poll(void) {
start:
    mp_handle_pending(true);

    int events_handled = 0;

    while (pbio_do_one_event()) {
        events_handled++;
    }

    // If there were any pbio events handled, don't sleep because there may
    // be something waiting on one of the events that was just handled.
    if (events_handled) {
        return;
    }

    // I.E. How to re-implement original so it can use the following common definition?

    uint32_t irq_flags = pbio_os_hook_disable_irq();
    if (!process_nevents()) {
        pbio_os_hook_wait_for_interrupt();
    }
    pbio_os_hook_enable_irq(irq_flags);
}


It looks like the Linux version of pbio_os_hook_wait_for_interrupt needs the orig_flags to make this pattern work. So maybe we just need to pass that on (and ignore it on embedded hubs).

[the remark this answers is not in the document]
laurensvalk: We can typdef `pbio_os_irq_flags_t` in pbioconfig so it can be `sigset_t` type on the virtual hub and `uint32_t` everywhere else. If we also pass it to the wait function:

```c
    if (!poll_request_is_pending) {
        pbio_os_hook_wait_for_interrupt(irq_flags);
    }
```

Then we can fix the issue with the global mask pointed out by @dlech as:

```c
void pbio_os_hook_wait_for_interrupt(pbio_os_irq_flags_t flags) {
    struct timespec timeout = {
        .tv_sec = 0,
        .tv_nsec = 100000,
    };
    // "sleep" with "interrupts" enabled
    sigset_t origmask = flags;
    MP_THREAD_GIL_EXIT();
    pselect(0, NULL, NULL, NULL, &timeout, &origmask);
    MP_THREAD_GIL_ENTER();
}
```

bricks/_common/micropython.c
```diff
@@ -39,14 +40,9 @@
 // Implementation for MICROPY_EVENT_POLL_HOOK
 void pb_event_poll_hook(void) {
 
-    // Drive pbio event loop.
-    while (pbio_do_one_event()) {
```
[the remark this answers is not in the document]
laurensvalk: Not all ports had this. But it probably also doesn't hurt to put this back, so let's do it.

laurensvalk:
> It would be nice to avoid #includes in mpconfig* if we can.
>
> Although if the header is very specifically crafted for such purpose (with comments that say so in the header itself), then it would be OK. Otherwise, it's easy to break things with future changes that aren't aware that mpconfig* aren't typical header files. However, we would want something just with the hook functions as os.h is way more than we should be including here.

We can fix:
- the above note
- having them static inline as they used to be before this PR
- having it in `/lib/pbio`
- https://github.com/pybricks/pybricks-micropython/pull/298/commits/9a6cc19d6ead42e234b64ce7a50c346f08334c4b polluting the existing `pbioconfig.h`.

all at once by having these hooks in a dedicated `pbio_os_config.h`, not unlike `contiki-conf.h`. Anything that shouldn't be a static inline can go in `platform.c` for the respective platform.

laurensvalk:
> > (unless there is a way to make the macros work another way
> 
> That is what I was trying to suggest in [#298 (comment)](https://github.com/pybricks/pybricks-micropython/pull/298#discussion_r2009307839) 😉
> 
> If we drop the do/while, the `{}` can be used as an expression statement.
> 
> ```c
> #define AWAIT(state, child, expression)  { \
>     ASYNC_INIT((child));                   \
>     PB_LC_SET(state);                      \
>     pbio_error_t _err = expression;        \
>     if (_err== PBIO_ERROR_AGAIN) {         \
>         return PBIO_ERROR_AGAIN;           \
>     }                                      \
>     _err;                                  \
> }
> ```
> 
> ```c
> pbio_error_t err;
> ...
> err = AWAIT(some_protothread())
> if (err)
>     return err;
> ...
> ```
> 
> I guess we would have to introduce a struct with two `pbio_error_t` values if we wanted to do the same with `AWAIT_RACE`.

That looks really nice, but it appears we cannot jump back into expression statements:

```
error: switch jumps into statement expression
```

lib/pbio/platform/nxt/nxos/memmap.h
```diff
@@ -56,8 +56,8 @@ extern uint8_t __data_ram_end__;
 extern uint8_t __bss_start__;
 extern uint8_t __bss_end__;
 
-extern uint8_t __stack_start__;
-extern uint8_t __stack_end__;
+extern uint8_t _sstack;
+extern uint8_t _estack;
```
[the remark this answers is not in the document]
laurensvalk: Good to know you don't mind the change in the STM32 builds. Agreed that more verbose is better.

lib/pbio/include/pbdrv/stack.h
```diff
@@ -0,0 +1,30 @@
...
+ * Gets the stack information.
+ *
+ * @param [out] sstack     The start of the stack.
+ * @param [out] estack     The end of the stack.
+ */
+void pbdrv_stack_get_info(char **sstack, char **estack);
+
+#else // PBDRV_CONFIG_STACK
+
+static inline void pbdrv_stack_get_info(char **sstack, char **estack) {
```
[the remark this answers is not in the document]
laurensvalk: Good idea.

laurensvalk:
Most points here have been addressed. Thanks for the feedback.

I'll move some of the `mp_hal` cleanups like the stack info to a new, separate pull request since it isn't really related to this one. 

> > Hmm... polling every processes every 1ms seems like it would make it harder to catch bugs where we aren't properly polling in response to other events.
> 
> Yes, this is one of the main design considerations we'll want to think about thoroughly before going ahead.
> 
> It simplifies processes a lot when everything is working, but debugging could be harder. Maybe we could have the auto-polling be optional for each thread. Then certain processes can independently (self-)poll as we used to do.
> 
> > It also means we have to be really careful about using timers in cases where we could be waiting on both a timer and another event a the same time.
> 
> This is one of the things that caught me out in the Contiki way of doing things -- if it got to a yield-once-come-back-later check, it could "miss" the event if it had just got it before getting there. In the new proposal, this should be simpler.
> 
> What cases are you thinking of that could now be trickier?

This is the slightly more fundamental one that is remaining. I'll think about it some more before we merge this.

laurensvalk:
I think we can merge this and develop from here.

Contiki will stay around for the foreseeable future so we can upgrade processes to the new pattern one by one without breaking things, or easily find where things did break. I have not yet tested @BertLindeman's findings. 

Size reductions probably won't be apparent until that time, but we have a bit of space to spare for now.

laurensvalk:
A few more relevant changes have also been merged, so you might want to pick the latest action / nightly build.

laurensvalk:
Thanks @BertLindeman. I'll keep an eye out for these two as we continue to develop. We're making rapid changes at the moment, so it's certainly possible that some issues pop up. I think this in particular is an area we'll be reviewing more closely in the future.

BertLindeman:
So happy that Pybricks 3 runs  on EV3!

laurensvalk:
We're of a mind to call it Pybricks 4.0 by the time we get there =)

laurensvalk:
@BertLindeman, are you still seeing any of these issues with the [latest nightly builds](https://nightly.link/pybricks/pybricks-micropython/workflows/build/master)? Many thanks!

BertLindeman:
will look at it  😃

BertLindeman:
Refer to [remark above](https://github.com/pybricks/pybricks-micropython/pull/298#issuecomment-2786356228)

All tests in this post with firmware at hash dfbcfa83 (nightly build 3821)

Problem 1 went away: was 
> Connected via BT to the UI (beta), press disconnect bluetooth in the UI and look at the hub light:

Now: hub goes after a short wait to blink-blink-pause So OK..

1. Problem I see now (running **cityhub**):
    - connect to beta gui
    - either:
        - "run" a program that refers to a movehub (so it fails on the cityhub)
        - do not run a program
    - while still connected: long press hub button 
    - see power off
    - see auto re-powered on

2. cityhub will not power off **if**:
    - cityhub is **off**
    - either 
        - short press to go to blink-blink-pause
        - press long to go to bootloader mode
    - release the button to go to blink-blink-pause
    - long press to power off
    - hub goes off and restarts to  blink-blink-pause
    - to power off either wait for the connection timeout or pull the batteries.

3. This is OK:
    - connect city hub and run a short program (print version)
    - hub is still connected, led stable blue
    - long press to power-off
    - hub goes off and stays off


[WIll update this post while testing other hubs]
(Lunch time now)

BertLindeman:
tests with `ci-build-3821-v3.6.1-159-gdfbcfa83 on 2025-05-19`


Short for "blink-blink-pause": b-b-p
All tests without any external devices attached.


1. Connect via BT to the UI (beta), press disconnect in the UI and look at the hub light (should go to b-b-p):
    cityhub     5 test OK
    movehub     5 test OK
    primehub    5 test OK
    technichub  5 test Same response as in movie hub.waits.before.BT.pairing.mode.mp4 in [this comment](https://github.com/pybricks/pybricks-micropython/pull/298#issuecomment-2786356228)



2. connect to beta gui | long press | see power off and should STAY off
    cityhub     ? depending on previous state? once wrong it keeps going wrong.
    movehub     5 test OK
    primehub    5 test OK
    technichub  5 test OK



3. hub is off | short press b-b-p | long press | power-off | release the button should NOT go to b-b-p
    cityhub     1 test OK but once wrong it keeps going wrong.
    movehub     5 test OK
    primehub    5 test OK
    technichub  5 test OK



4. hub is off | very long press until bootloadermode | release button | see b-b-p | long press to power-off 
       - Should release of the button go to b-b-p or power-off? I do not know.
    cityhub     in bootloader mode: release button: Hub goes to b-b-p
    movehub     in bootloader mode: release button: Hub goes silently OFF (no b-b-p)
    primehub    Different way to go to bootloader mode,
          - then long press center button (and keep long until BT button flashes white. OK
    technichub  in bootloader mode: release button: Hub goes to b-b-p



6. connect hub | run (print version) | hub still connected, led stable blue | long press to power-off | hub goes off and stays off
    cityhub     5 test OK
    movehub     5 test OK
    primehub    5 test OK
    technichub  5 test OK

Test program for 5:

```python
from pybricks import version
print(f'version {version}')
```

Remark on 5: if the program differs from this print version, also on test 5 the cityhub re-powers-on, 
Just the import of wait is enough to trigger the re-power-on, like on this program:
```python
from pybricks import version
print(f'version {version}')
```
and even re-powers-on if the program is as short as this:
```python
print('version')
```


7. After 5: hub ON | short press to run program from 5 | led b-b-p | long press to power-off | hub goes off and stays off
    cityhub     Re-power-on
    movehub     5 test OK
    primehub    5 test OK
    technichub  5 test OK


8. connect hub | run hub.system.shutdown() | hub goes off and should stay off. (ui gets GATT operation error as the hub silently left)
    cityhub     does Re-power-on 
    movehub     5 test OK
    primehub    5 test OK
    technichub  5 test OK

[Dinertime now]

BertLindeman:
Small thing (to me):
Stared with primehub, but City / Technic / movehub all do the same, all at git hash dfbcfa83.
Looks like a timeout might need a restart after disconnect.

Normally an idle hub keeps trying for some time to BT connect after the gui disconnects.

But, scenario:
- connect a hub to GUI and do not start a program
- wait ? 5 minutes
- disconnect on GUI button
- wait a few seconds
- hub does power-off

If I wanted to shutdown the hub, I would use the hub button (if I can reach it) 
Most of the time I disconnect on the GUI, I meant to start a program. So a miss-click.

PR #299, opened 2025-03-19
pbio/sys: Refactor status events to be handled synchronously

laurensvalk:
We never quite adopted events extensively across `pbio`. We only have a status set/clear handler in pbsys, with only one top level event listener at `pbsys_hmi_handle_status_change`. We can call it directly without casting the event data value to void * onto an event queue, and then back again. Given the limited case we had, this was harder to follow than we ultimately needed, and risks overflowing the event queue as stated in an old `REVISIT` that we can now address.

Doing this simplifies the introduction of https://github.com/pybricks/pybricks-micropython/pull/298 by ensuring we don't need any event queue at all.


dlech:
I think this will be OK, but it likely needs more work. I'm pretty sure that we have some processes that are depending on these broadcast events to "wake up" the process when the status changes even if they aren't explicitly checking the event. The main one that comes to mind is Bluetooth/USB, e.g. when the user program stops or when we get a hub shutdown request. Basically, we need to audit everywhere `pbsys_status_test()` is called an make sure we still have something that is polling that process when the status changes.

laurensvalk:
Thanks for the review!

> I think this will be OK, but it likely needs more work. I'm pretty sure that we have some processes that are depending on these broadcast events to "wake up" the process when the status changes even if they aren't explicitly checking the event.

This is one of those cases where https://github.com/pybricks/pybricks-micropython/pull/298 should really simplify things.

Where this code used to broadcast, it can call `pbio_os_request_poll` to achieve the same. 

We can wait with this one until #298 is merged.

laurensvalk:
> I think this will be OK, but it likely needs more work. I'm pretty sure that we have some processes that are depending on these broadcast events to "wake up" the process when the status changes even if they aren't explicitly checking the event. The main one that comes to mind is Bluetooth/USB, e.g. when the user program stops or when we get a hub shutdown request. Basically, we need to audit everywhere `pbsys_status_test()` is called an make sure we still have something that is polling that process when the status changes.

We can address this for now by maintaining the Contiki broadcast event in the status change until all processes that poll the status have been updated.

PR #302, opened 2025-04-09
Update EV3 README.md

BertLindeman:
Note that the REPL does not start automatically.


laurensvalk:
Thanks! It does start automatically, but you may not see the `>>>`.

Having the REPL here is just temporary though.

PR #304, opened 2025-05-19
pbio/drv/usb_stm32: Don't stall when no app connected.

laurensvalk:
Partially implements the timeout referenced below:

```
    // TODO: return PBIO_ERROR_INVALID_OP if app flag is not set. Also need a
    // timeout in case the app crashes and doesn't clear the flag on exit.
```

I am not merging this since @dlech is working on this, but just archiving the commit here.

It should work much better when subscribing with the app flag above. Then the timeout only affects one disconnect after normal use, rather than every time it is plugged in as in this commit.


laurensvalk:
I am not merging this since @dlech is working on this, but just archiving the commit here.

PR #309, opened 2025-06-14
pbio/port_lump: Fix ColorLightMatrix by sending initial NACK.

laurensvalk:
It turns out the ColorLightMatrix needs the NACK to get going at all, so make sure we always send one before we start waiting on the keep-alive timeout. In b17e5033dcc4edd75722839c4832c791f0e698a8, we removed grace period of 6 keep-alive messages to go unanswered, which was essentially masking this issue so the device worked before.

### DIFF
--- a/lib/pbio/src/port_lump.c
+++ b/lib/pbio/src/port_lump.c
@@ -881,8 +881,6 @@ sync:
         default_mode = LEGO_DEVICE_MODE_PUP_REL_MOTOR__POS;
     } else if (lump_dev->type_id == LEGO_DEVICE_TYPE_ID_COLOR_DIST_SENSOR) {
         default_mode = LEGO_DEVICE_MODE_PUP_COLOR_DISTANCE_SENSOR__RGB_I;
-    } else if (lump_dev->type_id == LEGO_DEVICE_TYPE_ID_EV3_COLOR_SENSOR) {
-        default_mode = LEGO_DEVICE_MODE_EV3_COLOR_SENSOR__AMBIENT;
     }
     if (default_mode) {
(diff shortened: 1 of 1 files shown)


laurensvalk:
Might need some testing to ensure that the extra NACK does not upset any other sensors.

laurensvalk:
The EV3 color sensor needs this too. It was also relying on the grace period of 6 keep-alive failures and stopped working when we removed that in b17e5033dcc4edd75722839c4832c791f0e698a8. The workaround for that was to set an initial mode in 91acfb321b54ac083e00712f087bb2d6f4c51543 but the initial NACK takes care of it too.

laurensvalk:
It looks like we had the allowance for 6 errors all the way back to the start https://github.com/pybricks/pybricks-micropython/commit/ab61ffd since adapting it from ev3dev.

```c
        // make sure we are receiving data
        if (!data->data_rec) {
            data->num_data_err++;
            DBG_ERR(data->last_err = "No data since last keepalive");
            if (data->num_data_err > 6) {
                data->status = PBIO_UARTDEV_STATUS_ERR;
            }
        }
```

@dlech, do you recall why it was there?

laurensvalk:
This is working fine on SPIKE Medium motors, the Technic XL motor, the Force Sensor, the Color Distance Sensor, the Tilt Sensor, Boost motor, Color Sensor and Ultrasonic Sensor. Let's proceed and keep an eye on it as we go.

laurensvalk:
PSA: @antonvh: We're making a slight update to the protocol by sending an extra keep-alive message right after syncing up. All LEGO sensors seem to be expecting this and it might work just fine on yours, but letting you know just in case you want to test it :shipit:

PR #312, opened 2025-06-29
pbio/protocol: add slots info to hub capabilities and status

dlech:
In order to support multiple program slots, connected apps will need to know how many slots are available on the hub and what the currently selected slot is. These are added to the hub capabilities and status report respectively and the Pybricks Profile version is bumped to 1.5.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 ### Added
 - Experimental support for USB connectivity on SPIKE Prime ([pybricks-micropython#208]).
 - Initial support for `pybricks.iodevices.UARTDevice` ([support#220]). 
+- Enabled previously hidden support for multiple code
+  slots ([pybricks-micropython#264], [pybricks-micropython#312]). A new
+  (unreleased) version of Pybricks Code is needed to use this.
 
 ### Changed
 - Extensive overhaul of UART and port drivers on all hubs. This affects all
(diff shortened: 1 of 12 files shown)


laurensvalk:
Thanks for working on this, this is much appreciated.

To make this fully useful, should we also move `pbsys_storage_get_maximum_program_size()` from hub capabilities to the status?

This should resolve most of the limitations of this initial slot approach. 

(Or we can have a closer look at uploads in general, since memory offsets doesn't quite work anymore in this approach - the current way of handling it is a bit of a hack. But this might take a lot longer to get right).

dlech:
> To make this fully useful, should we also move `pbsys_storage_get_maximum_program_size()` from hub capabilities to the status?

No, I think this would take too many bytes in the status message for something only changes in response to an external command.

Instead, I think we could consider a new characteristic that we can read to ask "what is in the slots". This would only be present on hubs that support slots. This could return file names and program sizes so we can see the current layout and how much free space is left. And is useful if you haven't used your hub for a while and forgot what was on it.

And I had a shower thought this morning that we probably also want to add a command to allow remotely selecting the slot on the hub UI.

laurensvalk:
Yeah, it's probably worth brainstorming a bit over :tea: / :coffee: . Would be great if we come up with something that will be generic enough for what we want to do after this, even if we only implement of subset of those ideas for now.

> add a command to allow remotely selecting the slot on the hub UI.

And also delete. Could be part of the broader discussion above.

laurensvalk:
We are going to have this firmware in beta for quite a while I think, so this is quite likely to raise some issues.

> Since the hub is telling Pybricks Code how many slots it has, Pybricks Code should already be doing the right thing in this regard. There aren't any officially released firmwares like this that have slots enabled, so it doesn't matter if we send 0 or no argument, IIRC.

```python

hub = PrimeHub()
hub.light.on(Color.BLUE) # Put red in first slot, green in second slot, blue in third
wait(5000)
```

Slots are downloaded to the selected slot, but it always runs 0, so you'd always see red even if the current program tries to make it blue.

PR #313, opened 2025-06-29
Update to MicroPython v1.25

dlech:
We've missed a few MicroPython releases, so it is about time we updated.

This passed locally with `make all` so we'll see what else CI catches. Also, I haven't actually tried running this on a hub yet, so we'll need to do some manual testing to see if anything else broke.

Some key things that changed to be tested are raising SystemExit/pressing stop button and module imports (both builtin and user),  

Once we've fixed any regressions, I would like to set `MICROPY_QSTR_BYTES_IN_HASH=0` on hubs where we are starting to run out of flash memory to regain some space.


dlech:
Well, clearly imports are broken as seen in the failed virtualhub tests.

And unexpectedly, the code formatter tool is also broken.

dlech:
If you have some ideas already about how you would like to handle imports differently, please go ahead. Otherwise, I will probably get back to it on Friday.

dlech:
Well, we didn't free up any space yet. But the virtualhub has been updated to do it the new "normal" way.

dlech:
The stop button and imports are passing the smoke test, so I'll go ahead an merge this and I'm sure people will let us know if anything unexpected broke.

PR #314, opened 2025-07-04
[EV3] Implement a SPI flash driver for the EV3

ArcaneNibble:
This PR implements a driver for the EV3 SPI flash memory.


ArcaneNibble:
With the latest commit the flash + ADC seem to both work (not at the same time but instead using the suggested hack).

I rewrote the entire ADC sampling logic to use DMA.

dlech:
> but something I've done seems to have made code that (I promise!) was previously working no longer reach the REPL (hanging without any debug output).

Rebasing on some changes I just pushed might help as I found and fixed a bug with the wrong entry point address in the uImage. Different builds could have the linker put the `Entry` at different places. See https://github.com/pybricks/support/issues/2256#issuecomment-3038057599

lib/pbio/drv/adc/adc_ev3.c
```diff
@@ -48,127 +102,94 @@ pbio_error_t pbdrv_adc_get_ch(uint8_t ch, uint16_t *value) {
...
-        SPIIntEnable(SOC_SPI_0_REGS, SPI_TRANSMIT_INT);
-        PROCESS_WAIT_EVENT_UNTIL(ev == PROCESS_EVENT_POLL && !adc_busy);
-
-        for (uint32_t i = 0; i < pbdrv_adc_callback_count; i++) {
-            pbdrv_adc_callbacks[i]();
-        }
-
-        etimer_reset(&etimer);
-    }
+void pbdrv_adc_ev3_configure_data_format() {
```
dlech, line 178: When using DMA, we need to max out all of the delays with `SPIDelayConfigure()`. Even with this, we get slightly inaccurate readings because there isn't enough settle time between readings, so if you read a 0V channel followed by a 5V channel, the reading for the 5V channel is a bit less than 5V because it takes time for the voltage to transition from 0V to 5V when the mux is switched.

Slowing the the SCLK could probably help with this as well.

lib/pbio/drv/block_device/block_device_ev3.c
```diff
@@ -0,0 +1,730 @@
...
+
+void pbdrv_block_device_init(void) {
+    // SPI module basic init
+    PSCModuleControl(SOC_PSC_0_REGS, HW_PSC_SPI0, PSC_POWERDOMAIN_ALWAYS_ON, PSC_MDCTL_NEXT_ENABLE);
+    SPIReset(SOC_SPI_0_REGS);
+    SPIOutOfReset(SOC_SPI_0_REGS);
+    SPIModeConfigure(SOC_SPI_0_REGS, SPI_MASTER_MODE);
+    unsigned int  spipc0 = SPI_SPIPC0_SOMIFUN | SPI_SPIPC0_SIMOFUN | SPI_SPIPC0_CLKFUN | SPI_SPIPC0_SCS0FUN0 | SPI_SPIPC0_SCS0FUN3;
+    SPIPinControl(SOC_SPI_0_REGS, 0, 0, &spipc0);
+    SPIDefaultCSSet(SOC_SPI_0_REGS, (1 << 0) | (1 << 3));
```
dlech, line 676: Let's use macros here so we know what these bits are.

`SPI_DATA_FORMATx` are not really better though. I would introduce a new one that says `SPI_BITS_PER_WORD(x)`

lib/pbio/drv/adc/adc_ev3.c
```diff
@@ -48,127 +102,94 @@ pbio_error_t pbdrv_adc_get_ch(uint8_t ch, uint16_t *value) {
...
+    SPIClkConfigure(SOC_SPI_0_REGS, SOC_SYSCLK_2_FREQ, SPI_CLK_SPEED_ADC, SPI_DATA_FORMAT1);
+    // NOTE: Cannot be CPOL=1 CPHA=1 like SPI flash
+    // The ADC seems to use the last falling edge to trigger conversions (see Figure 1 in the datasheet).
+    SPIConfigClkFormat(SOC_SPI_0_REGS, SPI_CLK_POL_LOW | SPI_CLK_OUTOFPHASE, SPI_DATA_FORMAT1);
+    SPIShiftMsbFirst(SOC_SPI_0_REGS, SPI_DATA_FORMAT1);
+    SPICharLengthSet(SOC_SPI_0_REGS, 16, SPI_DATA_FORMAT1);
+}
```
dlech, line 185: I would also set other config explicitly here even if it is the same as flash.

For example, `SPIDefaultCSSet()` is an interesting one. Omitting `SPI_CSHOLD` there is important for using the ADC with DMA so that it toggles the CS line between each word.

dlech (commented):
I didn't do an in-depth review since I know this is still a work in progress, but I like what I see so far. :smile:

lib/pbio/drv/adc/adc_ev3.c
ArcaneNibble: The ADS7957 datasheet says that the sampling instant is controlled by the falling edge of the chip select, so it sounds like changing `WDELAY` (rather than the `SPIDELAY` delays) should fix this situation?

Am I correct in understanding that this is entirely an analog circuit problem and not a digital timing problem?

lib/pbio/drv/adc/adc_ev3.c
```diff
@@ -25,15 +24,70 @@
...
+#define PBDRV_CONFIG_ADC_EV3_NUM_DELAY_SAMPLES (2)
+
+/**
+ * Constants.
+ */
+enum {
+    // The maximum ADC clock speed according to the datasheet is 20 MHz
+    // which would require a division factor of /7.5 which is not possible
+    // as this peripheral doesn't have a fractional clock generator.
+    // The TI HAL does *not* perform rounding down and only truncates the divider,
```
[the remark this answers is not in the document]
ArcaneNibble: I fixed the HAL to always round up

lib/pbio/drv/adc/adc_ev3.c
```diff
@@ -25,15 +24,70 @@
...
+    MANUAL_ADC_CHANNEL(7),
+    MANUAL_ADC_CHANNEL(8),
+    MANUAL_ADC_CHANNEL(9),
+    MANUAL_ADC_CHANNEL(10),
+    MANUAL_ADC_CHANNEL(11),
+    MANUAL_ADC_CHANNEL(12),
+    MANUAL_ADC_CHANNEL(13),
+    MANUAL_ADC_CHANNEL(14),
+    MANUAL_ADC_CHANNEL(15),
+
```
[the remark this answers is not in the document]
ArcaneNibble: Added

lib/pbio/drv/adc/adc_ev3.c
ArcaneNibble: Added maximum delays for all timing parameters

lib/pbio/drv/adc/adc_ev3.c
ArcaneNibble: Which other config settings were you thinking of? `SPIDefaultCSSet` doesn't configure the `CSHOLD` bit. That bit is in fact cleared by the DMA commands to `SPIDAT1`

ArcaneNibble:
I've cleaned up the history of this PR and marking it as ready for review.

laurensvalk:
Thanks! Rebased and applied `./tools/codeformat.py` to each commit along the way. I've tested storage with `hub.system.storage`. The analog touch sensor is detected and works.

I've been working on https://github.com/pybricks/support/issues/2264, but I propose to merge yours first for simplicity. 

Since EV3 is under heavy development, I think we can merge it as-is and address open TODOs in follow up pull requests.

PR #315, opened 2025-07-05
EV3 u-boot fixes

dlech:
This cleans up some stuff related to using u-boot on the EV3 and updates to the latest mainline u-boot (with some Pybricks modifications).


bricks/ev3/make_bootable_image.py
```diff
@@ -21,14 +22,18 @@
 FIRMWARE_ROUND_SIZE = 256 * 1024
 
 
-def make_firmware(uboot_blob, uimage_blob):
+def make_firmware(uboot_blob: bytes, uimage_blob: bytes) -> bytes:
 
     if len(uboot_blob) > UBOOT_MAX_SIZE:
-        print("u-boot file is bigger than expected. Using only the first 256KiB.")
-        uboot_blob = uboot_blob[:UBOOT_MAX_SIZE]
```
laurensvalk, line 25: This was initially used to use the original LEGO firmware as a source to extract u-boot. But if we have a proper u-boot now, we should indeed no longer need it.

PR #318, opened 2025-07-05
bricks/ev3: use .elf instead of uImage

dlech:
Replace creating a uImage with using the .elf file directly for the EV3 firmware. This simplifies the build process and avoids the need for u-boot-tools. U-Boot already knows how to load the .elf file, so we can use it directly by using `bootelf` instead of `bootm` (this change was made in the pybricks/v2.0.0 release of u-boot).

In order to keep the size of the .elf file small, we do not enable `-ffunction-sections` and `-fdata-sections` for the EV3 firmware. This avoids the large tables of section names that are generated by these flags, which would otherwise increase the size of the .elf file by nearly 200 kB. The .elf file is still a bit larger than the uImage because of a bit more overhead, but only by less than 4kB.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -93,7 +93,7 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
     - name: Install cross-compiler
-      run: sudo apt-get update && sudo apt-get install --yes gcc-arm-none-eabi u-boot-tools
+      run: sudo apt-get update && sudo apt-get install --yes gcc-arm-none-eabi
     - name: Checkout repo
       uses: actions/checkout@v4
       with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Install cross-compiler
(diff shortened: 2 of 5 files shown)


ArcaneNibble:
> avoids the large tables of section names that are generated by these flags

You should be able to fix this with an appropriate linker script, which can also be used to keep important functions so that they are not dropped. I can look into this later if we are really trying to keep code sizes down.

dlech:
> I can look into this later if we are really trying to keep code sizes down.

Thanks, but I'm happy with how it turned out. `-ffunction-sections` creates a separate `.text.function_name` section for every single section, so we ended up with several thousand sections. When I turned off that option, we now just have the most basic section list we can have. So I don't think there is really any more to gain here. We have plenty of flash and RAM on EV3, so a little bit extra doesn't matter. The extra 180kB from the section tables was about 1/3 of the total .elf file size though, so seemed a bit much.

BertLindeman:
installed ev3 firmware from build 3888 at git2c7a38e7. No REPL.
One step back:
installed ev3 firmware from build 3887 at gitd4dfeec7. REPL OK.

Somewhere there is something different between these two builds.

PR #323, opened 2025-07-08
Ev3 display API

schodet:
First step, including:

- a driver abstraction,
- migration to pbio_os of the ev3 display driver,
-  the Image class,
- drawing:
  - clear,
  - sub-images,
  - single pixel,
  - image in image,
  - rectangle, filled rectangle, rounded rectangle, rounded filled rectangle,
  - circle and disc,
- python binding for all this.

Next steps:

- tests, tests, and tests,
- add ev3dev media,
- implement line (easy), and thick line (hum… fun),
- text,
- loading/saving from/to byte strings, or files.


lib/pbio/drv/display/display_ev3.c
```diff
@@ -503,27 +455,67 @@ PROCESS_THREAD(pbdrv_display_ev3_init_process, ev, data) {
...
+
+    // Waking up the SPI1 instance.
+    PSCModuleControl(SOC_PSC_1_REGS, HW_PSC_SPI1, PSC_POWERDOMAIN_ALWAYS_ON, PSC_MDCTL_NEXT_ENABLE);
+
+    // Reset.
+    SPIReset(SOC_SPI_1_REGS);
+    SPIOutOfReset(SOC_SPI_1_REGS);
+
+    // Mode.
+    uint32_t spipc0 = SPI_SPIPC0_SOMIFUN | SPI_SPIPC0_SIMOFUN | SPI_SPIPC0_CLKFUN | SPI_SPIPC0_ENAFUN;
```
dlech, line 502: `SPI_SPIPC0_SOMIFUN` isn't actually wired up, so we could leave this out.

I don't think we are using `SPI_SPIPC0_ENAFUN` either.

lib/pbio/drv/display/display_ev3.c
```diff
@@ -503,27 +455,67 @@ PROCESS_THREAD(pbdrv_display_ev3_init_process, ev, data) {
...
+    SPIModeConfigure(SOC_SPI_1_REGS, SPI_MASTER_MODE);
+    SPIPinControl(SOC_SPI_1_REGS, 0, 0, (unsigned int *)&spipc0);
```
dlech, line 504: Should not need cast here. In fact I would just split `SPIPinControl()` into two separate functions to avoid this oddness.

dlech (commented):
Looks amazing!

This is already very tidy, but I managed to find a few places where we could tidy it up a bit more. And the suggestions on improving the `pbdrv_display_init()` function can be saved for a follow up.

lib/pbio/drv/display/display_ev3.c
schodet: I did not change this code, I moved the function below `pbdrv_display_ev3_process_thread`. I change it to split it in `pbdrv_display_ev3_spi_init` and `pbdrv_display_init` so it’s clearer.

lib/pbio/drv/display/display_ev3.c
schodet: Idem.

I do not get the “split `SPIPinControl()`” part.

laurensvalk: This also comes from the TI examples, so I would leave it in place for now. When the refactored code is in place and merged, we can have a closer look at cleaning it up (while starting from a point that we know is working).

pybricks/media/pb_type_image.c
```diff
@@ -0,0 +1,336 @@
...
+            mp_int_t y1 = pb_obj_get_int(y1_in);
+            mp_int_t x2 = x2_in == mp_const_none ? source->image.width - 1 : pb_obj_get_int(x2_in);
+            mp_int_t y2 = y2_in == mp_const_none ? source->image.height - 1 : pb_obj_get_int(y2_in);
+            self->buffer = source->buffer;
+            self->update = false;
+            int width = x2 - x1 + 1;
+            int height = y2 - y1 + 1;
+            pbio_image_init_sub(&self->image, &source->image, x1, y1, width, height);
+        }
+    } else {
```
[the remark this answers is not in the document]
laurensvalk: That is a good question. I suppose if we want to keep things backwards-compatible, we should use `screen` instead of `display`. I think we've been using `display` on the SPIKE Prime since it clearly isn't a screen and they have a different interface.

pybricks/media/pb_type_image.c
```diff
@@ -0,0 +1,336 @@
...
+#include <pbio/int_math.h>
+
+#include <pbdrv/display.h>
+
+extern const mp_obj_type_t pb_type_Image;
+
+// pybricks.media.Image class object
+typedef struct _pb_type_Image_obj_t {
+    mp_obj_base_t base;
+    uint8_t *buffer;  // m_malloc'd buffer, may be shared, NULL for display.
```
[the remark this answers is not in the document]
dlech: Sure, TODO is fine here too.

lib/pbio/drv/display/display_ev3.c
dlech: Yup, I had the same comment in the ADC code, so we can do that later.

dlech:
> I have no problem with rewriting, but for a reviewer, seeing the `fixup!` commits before the rewrite may be nicer.

GitHub will show a diff of only what changed when you force push, so pushing `!fixup` commits isn't necessary. But be careful - if you rebase on a new master branch and force push, all of the master branch changes will show up in this diff as well! To work around that, you can rebase on master, force push, then make the fixups then force push again.

pybricks/media/pb_type_image.c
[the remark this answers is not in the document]
dlech: To keep things simple for now, I would hold off on a CompressImage - it isn't like we are running out of space on the EV3 already. It would be better to discuss a bit of what we want the whole user experience to be like first, e.g. how do users create or get images? Is there a set of "standard" images that we ship with the firmware? Or do we distribute them with Pybricks Code and only download to the EV3 as needed? And many more questions.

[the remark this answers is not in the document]
dlech: Got it. My original comment in this thread was pointing out that if we pass anything other than an `Image` object or `"_screen_"`, it is treated as `"_screen_"` rather than raising a `TypeError`. Normally, I would expect an invalid argument to raise such an error.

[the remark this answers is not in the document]
schodet: Yes, it was a place for a TODO, well, I added the test, it’s better now.

pybricks/media/pb_type_image.c
```diff
@@ -0,0 +1,353 @@
...
+            mp_int_t y1 = pb_obj_get_int(y1_in);
+            mp_int_t x2 = x2_in == mp_const_none ? source->image.width - 1 : pb_obj_get_int(x2_in);
+            mp_int_t y2 = y2_in == mp_const_none ? source->image.height - 1 : pb_obj_get_int(y2_in);
+            self->buffer = source->buffer;
+            self->is_display = false;
+            int width = x2 - x1 + 1;
+            int height = y2 - y1 + 1;
+            pbio_image_init_sub(&self->image, &source->image, x1, y1, width, height);
+        }
+    } else if (mp_obj_is_qstr(source_in) && MP_OBJ_QSTR_VALUE(source_in) == MP_QSTR__screen_) {
```
dlech, line 90:
```suggestion
    } else if (mp_obj_equal(source_in, MP_OBJ_NEW_QSTR(MP_QSTR__screen_)) {
```
Not sure if it is guaranteed the arg will be a qstr (not all strings are qstrs!). And this is perhaps a bit more like how it would be executed it it was written in Python anyway.

lib/pbio/drv/display/display_ev3.c
```diff
@@ -18,6 +18,7 @@
 #include "../core.h"
 
 #include <pbdrv/gpio.h>
+#include <pbdrv/display.h>
```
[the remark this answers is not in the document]
dlech: My eyes were playing tricks on me. I was thinking it should go after error.h, but `pbdrv/` is not the same as `pbio/`!

lib/pbio/drv/display/display_ev3.c
```diff
@@ -13,13 +13,13 @@
...
 
-#include <contiki.h>
-
 #include "../core.h"
 
+#include <pbdrv/display.h>
 #include <pbdrv/gpio.h>
 #include <pbio/error.h>
 #include <pbio/util.h>
+#include <pbio/os.h>
```
dlech, line 22: Alphabetical order?

pybricks/media/pb_type_image.c
```diff
@@ -0,0 +1,353 @@
...
+        // Screen.
+        self->buffer = NULL;
+        self->is_display = true;
+        self->image = *pbdrv_display_get_image();
+    } else if (mp_obj_is_str(source_in)) {
+        // TODO: load from image "files".
+        mp_raise_ValueError(MP_ERROR_TEXT("'_screen_' is the only supported source string"));
```
dlech, line 97:
```suggestion
        mp_raise_NotImplementedError(MP_ERROR_TEXT("'_screen_' is the only supported source string"));
```

Assuming we intend to implement it.

pybricks/media/pb_type_image.c
```diff
@@ -0,0 +1,353 @@
...
+static int get_color(mp_obj_t obj) {
+    uint8_t max = pbdrv_display_get_max_value();
+    if (obj == mp_const_none) {
+        return max;
+    }
+    if (mp_obj_is_int(obj)) {
+        return mp_obj_get_int(obj);
+    }
+    const pbio_color_hsv_t *hsv = pb_type_Color_get_hsv(obj);
+    int32_t v = pbio_int_math_bind(hsv->v, 0, 100);
```
dlech, line 42: This is fine for now, but if we want to convert true color images to grayscale, then it works better to take hue and saturation into account as well. But that is one of those bigger discussion questions. :smile:

pybricks/media/pb_type_image.c
```diff
@@ -0,0 +1,353 @@
...
+    mp_int_t y = pb_obj_get_int(y_in);
+    mp_int_t r = pb_obj_get_int(r_in);
+    bool fill = mp_obj_is_true(fill_in);
+    int color = get_color(color_in);
+
+    if (!fill) {
+        pbio_image_draw_circle(&self->image, x, y, r, color);
+    } else {
+        pbio_image_fill_circle(&self->image, x, y, r, color);
+    }
```
dlech, line 322:
```suggestion
    if (fill) {
        pbio_image_fill_circle(&self->image, x, y, r, color);
    } else {
        pbio_image_draw_circle(&self->image, x, y, r, color);
    }
```

Minor nit: reads a bit easier without the `!` to my eyes.

pybricks/media/pb_type_image.c
```diff
@@ -0,0 +1,353 @@
...
+        PB_ARG_DEFAULT_INT(width, 1),
+        PB_ARG_DEFAULT_NONE(color));
+
+    mp_int_t x1 = pb_obj_get_int(x1_in);
+    mp_int_t y1 = pb_obj_get_int(y1_in);
+    mp_int_t x2 = pb_obj_get_int(x2_in);
+    mp_int_t y2 = pb_obj_get_int(y2_in);
+    mp_int_t width = pb_obj_get_int(width_in);
+    int color = get_color(color_in);
+
```
dlech, line 253: Might want to `mp_raise_NotImplementedError()` here since the pbio functions silently don't do anything currently. Unless we intend to implement those right away. (Bert, likes to test everything right away, so he might report a bug :grin:)

pybricks/media/pb_type_image.c
```diff
@@ -0,0 +1,353 @@
...
+        PB_ARG_REQUIRED(y),
+        PB_ARG_REQUIRED(source),
+        PB_ARG_DEFAULT_NONE(transparent));
+
+    mp_int_t x = mp_obj_get_int(x_in);
+    mp_int_t y = mp_obj_get_int(y_in);
+    pb_assert_type(source_in, &pb_type_Image);
+    pb_type_Image_obj_t *source = MP_OBJ_TO_PTR(source_in);
+    bool transparent = false;
+    int transparent_value;
```
dlech, line 196: Probably should initialize this as well so static analyzers don't complain about uninitialized use.

pybricks/media/pb_type_image.c
```diff
@@ -0,0 +1,353 @@
...
+        transparent = true;
+        transparent_value = get_color(transparent_in);
+    }
+
+    if (!transparent) {
+        pbio_image_draw_image(&self->image, &source->image, x, y);
+    } else {
+        pbio_image_draw_image_transparent(&self->image, &source->image, x, y, transparent_value);
+    }
+
```
dlech, line 207:
```suggestion
    if (transparent_in != mp_const_none) {
        int transparent_value = get_color(transparent_in);
        
        pbio_image_draw_image_transparent(&self->image, &source->image, x, y, transparent_value);
    } else {
        pbio_image_draw_image(&self->image, &source->image, x, y);
    }

```

Or just combine the if statements to avoid it altogether.

dlech (commented):
I'm starting to get nit-picky, so this must be in good shape, lol. The main suggestion is about the `"_screen_"` equality check. Other than that, nothing too important.

Before we merge it though,we'll want to rebase on the master branch. I updated the MicroPython version recently and that changed a few of the alloc functions, so we'll want to make sure it didn't affect this or fix it if it did.

pybricks/media/pb_type_image.c
schodet: For me, the no-fill is the default normal case, that’s why I put it first, but OK, I can make the change :)

I also change the box if.

pybricks/media/pb_type_image.c
dlech: > should we add a light gray and dark gray in Color?

Not yet. Let's hold off on that until we have a look at the APIs again and reconsider what would be best for both NXT and EV3.

laurensvalk: > This was just to accept a Color parameter. By the way, should we add a light gray and dark gray in Color?

You can do `Color.WHITE * 0.75` and `Color.WHITE * 0.25`.

schodet:
I added a fix for `pbdrv_display_pybricks_logo` which was missing an end-index. This was shown after the rebase due to a change of the half-word after the table.

dlech:
Thanks! Really great work going on here.

PR #327, opened 2025-07-11
pbio/platform/ev3: Read Bluetooth MAC address from boot EEPROM

ArcaneNibble:
This is needed for BT as well as to have a unique serial number for USB.

This is implemented with simple polling because it only needs to be done once on bootup


ArcaneNibble:
Changes made

PR #333, opened 2025-07-14
[EV3] Implement the rest of the clock API

ArcaneNibble:
This implements the rest of the clock API. It currently assumes that IRQs are never disabled on the EV3.

Closes https://github.com/pybricks/support/issues/2270


lib/pbio/drv/clock/clock_ev3.c
```diff
@@ -63,15 +62,17 @@ void systick_isr_C(void) {
  */
 void systick_suspend(void) {
     /* Disable the timer interrupt */
-    TimerDisable(SOC_TMR_0_REGS, TMR_TIMER34);
```
laurensvalk, line 66: I think these hooks stem from another project and we don't use them, so we could drop them.

lib/pbio/drv/clock/clock_ev3.c
```diff
@@ -63,15 +62,17 @@ void systick_isr_C(void) {
...
+    TimerDisable(SOC_TMR_0_REGS, TMR_TIMER12);
+    systick_is_suspended = 1;
 }
 
 /**
  * Enable the timer and therefore the systick
  */
 void systick_resume(void) {
     /* Enable the timer interrupt */
-    TimerEnable(SOC_TMR_0_REGS, TMR_TIMER34, TMR_ENABLE_CONTRELOAD);
```
laurensvalk, line 74: Same hre.

lib/pbio/drv/clock/clock_ev3.c
ArcaneNibble: Removed these functions in the latest version

lib/pbio/drv/clock/clock_ev3.c
```diff
@@ -21,16 +21,16 @@
...
+// The input to the Timer0 module is PLL0_AUXCLK. This will always run at 24 MHz on the EV3,
+// regardless of what the CPU speed is set to. We configure the timer to generate interrupts
+// every millisecond, and we configure the timer to reset to 0 when it reaches a count value
+// corresponding to one millisecond of time. Finer timing resolution can be obtained by
+// combining the software-managed millisecond counter with the timer value.
+//
+// The system tick uses the "12" half of the timer, and the PRU1 (TODO) uses the "34" half.
+const uint32_t auxclk_freq_hz = 24000000;
+const uint32_t timer_ms_period = (auxclk_freq_hz / 1000) - 1;
+const uint32_t timer_us_division = auxclk_freq_hz / 1000000;
```
dlech, line 33:
```suggestion
static const uint32_t auxclk_freq_hz = 24000000;
static const uint32_t timer_ms_period = (auxclk_freq_hz / 1000) - 1;
static const uint32_t timer_us_division = auxclk_freq_hz / 1000000;
```

Also, does it make sense to use `SOC_ASYNC_2_FREQ` or a related macro instead of hard-coding the frequency here?

PR #334, opened 2025-07-14
[EV3] Fix linker script to merge sections

ArcaneNibble:
This allows re-enabling the -fdata-sections -ffunction-sections optimization

PR #335, opened 2025-07-14
[EV3] Implement a panic exception handler

ArcaneNibble:
If an exception happens, print out all the registers on the debug UART and then reboot the system.

Also clean up and fix a latent bug in the vector table


ArcaneNibble:
I think this and https://github.com/pybricks/pybricks-micropython/pull/334 now collide, so I will rebase whichever one is not merged first.

dlech:
we'll go with this one first.

PR #336, opened 2025-07-14
Remove support for busy-wait millisecond delays with IRQs disabled

ArcaneNibble:
The current implementation of Pybricks does not disable IRQs for any significant length of time, which means that we can remove clock driver code supporting it.

The STM32 code still supports *micro*second delays.


ArcaneNibble:
Hrm, looks like NXT still has blocking delays in its bootup process. Going to keep those in place for now and only remove the mphalport.c support for running with IRQs disabled.

PR #339, opened 2025-07-15
[EV3] Implement LED brightness control via software PWM on PRU1

ArcaneNibble:
This is dependent on https://github.com/pybricks/pybricks-pru/pull/1 , and it needs to be modified to download the PRU binary rather than checking it into git (see related: https://github.com/pybricks/support/issues/2262 ), but I am opening the PR for now as the rest of the code is ready.

### DIFF
--- a/bricks/_common/arm_none_eabi.mk
+++ b/bricks/_common/arm_none_eabi.mk
@@ -564,6 +564,7 @@ OBJ += $(addprefix $(BUILD)/, $(TI_AM1808_USB_SRC_C:.c=.o))
 OBJ += $(addprefix $(BUILD)/, $(EV3_SRC_S:.S=.o))
 $(addprefix $(BUILD)/, $(EV3_SRC_S:.S=.o)): CFLAGS += -D__ASSEMBLY__
 OBJ += $(BUILD)/pru_suart.bin.o
+OBJ += $(BUILD)/pru_ledpwm.bin.o
 endif
 
 # List of sources for qstr extraction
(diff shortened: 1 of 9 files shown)


ArcaneNibble:
PR has been updated to download a released PRU binary (which is no worse than what currently happens for U-boot).

This is now ready for review.

ArcaneNibble:
Updated based on feedback.

Using the linker script incidentally means that overflowing the shared memory region will be caught. However, this PR doesn't yet migrate PRU0 to use it.

lib/pbio/platform/ev3/platform.c
```diff
@@ -87,7 +86,7 @@ const pbdrv_led_dual_platform_data_t pbdrv_led_dual_platform_data[PBDRV_CONFIG_L
 };
 
 static const pbdrv_led_pwm_platform_color_t pbdrv_led_pwm_color = {
-    // TODO: PWM not yet implemented, so these values not used.
+    // TODO: Calibrate these numbers, as well as .scale_factor
```
dlech, line 89: Here is the secret sauce:

```c
static const pbdrv_led_pwm_platform_color_t pbdrv_led_pwm_color = {
    // The red LED is much stronger than the green one, so leave green at 1000
    // and adjust red until yellow looks good.
    .r_factor = 150,
    .g_factor = 1000,
    .b_factor = 0,
    // 1250 allows max brightness for an 8-bit PWM size (with scale_factor = 1).
    // Increase to limit max brightness. Going lower will cause overflow.
    .r_brightness = 1250,
    .g_brightness = 1250,
    .b_brightness = 0,
};
```

bricks/_common/arm_none_eabi.mk
```diff
@@ -637,6 +638,11 @@ $(BUILD)/u-boot.bin:
 	$(Q)mkdir -p $(dir $@)
 	$(Q)curl -sL -o $@ https://github.com/pybricks/u-boot/releases/download/pybricks/v2.0.1/u-boot.bin
 	$(Q)echo "86ddad84f64d8aea85b4315fc1414bdec0bb0d46c92dbd3db45ed599e3a994cb  $@" | sha256sum -c --strict
+$(BUILD)/pru_ledpwm.bin:
+	$(ECHO) "Downloading pru_ledpwm.bin"
+	$(Q)mkdir -p $(dir $@)
+	$(Q)curl -sL -o $@ https://github.com/pybricks/pybricks-pru/releases/download/v0.0.1/pru_ledpwm.bin
```
dlech, line 644: I would already call it v1.0.0.  (I no longer believe in https://0ver.org/) :laughing:

PR #340, opened 2025-07-15
[EV3] New USB driver, batch 1

ArcaneNibble:
This is batch 1 of reimplementing the EV3 USB support using a bespoke USB stack.

It currently provides only standard USB descriptors. When data is sent to the EP1 OUT endpoint, it will be echoed back (with the value 1 added to each byte) on EP1 IN.

The next steps will involve refactoring out some of the shared WebUSB/Microsoft descriptor creation code as well as hooking this up to existing command handlers.


lib/pbio/drv/usb/usb_ev3.c
```diff
@@ -52,500 +28,823 @@
...
+            .pktRetQueue = RX_COMPQ1,
+        },
+        .buf_len = PYBRICKS_EP_PKT_SZ_HS,
+        .buf_ptr = ep1_rx_buf,
+        .next_desc_ptr = 0,
+        .orig_buf_len = PYBRICKS_EP_PKT_SZ_HS,
+        .orig_buf_ptr = ep1_rx_buf,
+    };
+
+    __asm__ volatile ("" ::: "memory");
```
dlech, line 293:
```suggestion
    CP15ICacheFlushBuff((uint32_t)&cppi_descriptors[CPPI_DESC_RX], sizeof(cppi_descriptors[CPPI_DESC_RX]));
```

Is this what we are going for here?

lib/pbio/drv/usb/usb_ev3.c
```diff
@@ -52,500 +28,823 @@
...
+    USBReset(USB0_BASE);
 
-    g_sCDCSerDeviceInfo.sCallbacks.pfnSuspendHandler = g_sCDCSerDeviceInfo.sCallbacks.pfnDisconnectHandler;
-    //
-    // Pass our device information to the USB library and place the device
-    // on the bus.
+    // Reset the PHY
+    HWREG(CFGCHIP2_USBPHYCTRL) |= CFGCHIP2_RESET;
+    for (int i = 0; i < 50; i++) {__asm__ volatile ("");
+    }
```
dlech, line 766:
```suggestion
    for (int i = 0; i < 50; i++) {
        __asm__ volatile (""); // this is a delay loop, don't optimize it out
    }
```

Or could use `pbdrv_clock_busy_delay_us();`

And would be helpful to have a comment explaining why this delay is needed. I'm guessing because "that is how TI did it"?

lib/pbio/drv/usb/usb_ev3.c
```diff
@@ -52,500 +28,823 @@
...
+    // Finally signal a connection
+    USBDevConnect(USB0_BASE);
+
+    // We are basically done. USB is event-driven, and so we don't have to block boot.
+    pbio_os_process_start(&pbdrv_usb_ev3_process, pbdrv_usb_ev3_process_thread, NULL);
 }
 
 pbdrv_usb_bcd_t pbdrv_usb_get_bcd(void) {
-    return g_bUSBConfigured ? PBDRV_USB_BCD_STANDARD_DOWNSTREAM : PBDRV_USB_BCD_NONE;
+    // TODO: Reimplement this
```
dlech, line 826: There isn't USB battery charging on EV3, so no TODO here. We probably don't even need to implement the function `PBSYS_CONFIG_BATTERY_CHARGER` should not be defined (or 0) on EV3, so this function should never be called.

lib/pbio/drv/usb/usb_ev3.c
```diff
@@ -52,500 +28,823 @@
...
-//
-// This function is called by the CDC driver to notify us of any events
-// related to operation of the receive data channel (the OUT channel carrying
-// data from the USB host).
-//
-// \return The return value is event-specific.
-//
-// *****************************************************************************
-unsigned int RxHandler(void *pvCBData, unsigned int ulEvent, unsigned int ulMsgValue, void *pvMsgData) {
+static void usb_device_intr() {
```
dlech, line 349:
```suggestion
static void usb_device_intr(void) {
```

And other places as well.

lib/pbio/drv/usb/usb_ev3.c
```diff
@@ -52,500 +28,823 @@
...
-        //
-        // We are being asked how much unprocessed data we have still to
-        // process. We return 0 if the UART is currently idle or 1 if it is
-        // in the process of transmitting something. The actual number of
-        // bytes in the UART FIFO is not important here, merely whether or
-        // not everything previously sent to us has been transmitted.
-        //
-        case USB_EVENT_DATA_REMAINING: {
-            return 0;
+        // Set up the FIFOs
```
dlech, line 368: This seems like a lot to do in an interrupt handler. Is it something we can defer? i.e. set a flag and poll the protothread/process to do it.

dlech (commented):
This is pretty dense so I had to give up about half way through, but here a  few comments anyway.

lib/pbio/drv/usb/usb_ev3.c
```diff
@@ -52,500 +28,823 @@
...
-const unsigned char *const g_pStringDescriptors[] =
-{
-    g_pLangDescriptor,
-    g_pManufacturerString,
-    g_pProductString,
-    g_pSerialNumberString,
-    g_pControlInterfaceString,
-    g_pConfigString
+enum {
+    CPPI_RX_SUBMIT_QUEUE = 0,
```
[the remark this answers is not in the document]
dlech: Or use a `#define` instead.

lib/pbio/drv/usb/usb_ev3.c
dlech: Doesn't that mean that we need this on every register read/write then?

In the Linux kernel, we have helper functions with various levels of memory barriers exactly for this reason.

ArcaneNibble: No, `HWREG` uses a `volatile` access. All `volatile` accesses are ordered with respect to each other. The only thing that is not ordered is `volatile` vs regular accesses.

lib/pbio/drv/usb/usb_ev3.c
ArcaneNibble: This is *visually* a lot of code, but if you actually look at what it does, it just does a series of register pokes. It's not really meaningfully longer than many of the other interrupt code paths.

lib/pbio/drv/usb/usb_ev3.c
dlech: Ah, so the reasoning here is that we want all struct fields to actually be assigned (committed to memory) before assigning the _pointer_ to the struct to the register.

ArcaneNibble:
Rebased this to be totally up to date with the latest merges. Also uses `pbdrv_compiler_memory_barrier` now.

lib/pbio/drv/usb/usb_ev3.c
```diff
@@ -52,500 +30,840 @@
...
+    // TODO: Reimplement this
+    return 0;
 }
 
 int32_t pbdrv_usb_get_char(void) {
-    uint8_t c;
-    if (USBBufferRead((tUSBBuffer *)&g_sRxBuffer, &c, 1) > 0) {
-        return c;
-    }
+    // TODO: Reimplement this
```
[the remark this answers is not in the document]
laurensvalk: > Some of these won't get re-implemented. Instead, we'll want to use `bricks/_common_stm32/mphalport.c` and drop `bricks/ev3/mphalport.c`. So we only need to implement what is called by `pbsy_host-*`.

Yes. We are heading towards a single `mphalport` for all builds (except perhaps the unix-based Virtual Hub). The IRQ hooks have been unified recently, and i/o will be standardized by `sys/host`.

So we can drop the `_stm32` wording from some of the stuff here.

ArcaneNibble:
I've just implemented all the requested changes.

lib/pbio/drv/usb/usb_ev3.c
```diff
@@ -52,504 +30,897 @@
...
+static volatile bool usb_rx_is_ready;
+static volatile bool usb_tx_response_is_not_ready;
+static volatile bool usb_tx_status_is_not_ready;
+static volatile bool usb_tx_stdout_is_not_ready;
+
+// CPPI DMA support code
+
+// Descriptors must be aligned to a power of 2 greater than or equal to their size.
+// We are using a descriptor of 32 bytes which is also the required alignment.
+#define CPPI_DESCRIPTOR_ALIGN   32
```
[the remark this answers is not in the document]
ArcaneNibble: Only `usb_cppi_hpd_t` needs to be aligned. The actual payload data being transferred can have any alignment.

dlech:
OK! Let's call it good enough and move on to the next one.

PR #341, opened 2025-07-16
[EV3] Minimum-viable speaker driver

ArcaneNibble:
This is a basic speaker driver for the EV3. The API surface exposed to user code is the same `Speaker` class as other hubs.

This runs the hardware PWM module at a fixed frequency using asymmetrical PWM and fetches samples using a FIQ handler on the ARM core.

This implementation struggles to play higher-frequency notes due to interactions with the way the `Speaker` class is implemented. Specifically, the `Speaker` class always generates a waveform of the same number of samples (128 samples) but keeps increasing the sampling rate. This driver always outputs samples to the hardware at a fixed rate and does not implement a proper resampler, so it ends up badly distorting the waveform. However, beeps and notes around octaves 2-4 work.

### DIFF
--- a/bricks/_common/sources.mk
+++ b/bricks/_common/sources.mk
@@ -175,6 +175,7 @@ PBIO_SRC_C = $(addprefix lib/pbio/,\
 	drv/reset/reset_stm32.c \
 	drv/reset/reset_ev3.c \
 	drv/resistor_ladder/resistor_ladder.c \
+	drv/sound/sound_ev3.c \
 	drv/sound/sound_nxt.c \
 	drv/sound/sound_stm32_hal_dac.c \
 	drv/stack/stack_embedded.c \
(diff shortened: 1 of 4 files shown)


dlech:
For beeps on the EV3, we want to just set the PWM period equal to 1 / frequency of the beep with a 50% duty cycle to get a square wave. This is how the official LEGO firmware and ev3dev do it. It will be much louder!

I'm not sure why we put the waveform generator in the MicroPython class. It is specific to SPIKE Prime because it has an actual audio peripheral rather than a PWM. So that need to get moved to the specific driver for that platform and we need to introduce a new `pbdrv_sound_beep()` method that sets the beep frequency. Then use that in the MicroPython Speaker class.

ArcaneNibble:
Ah, but doing that will make it even harder to implement custom wave shapes, which looks to at least be a vaguely-planned feature?

PR #342, opened 2025-07-16
[v2] [EV3] Minimum-viable speaker driver

ArcaneNibble:
This is a basic speaker driver for the EV3. The API surface exposed to user code is the same Speaker class as other hubs.

Unlike [v1](https://github.com/pybricks/pybricks-micropython/pull/341), this driver directly generates a square wave of the desired frequency rather than playing back samples.

As a consequence, it does not support volume control. However, it plays the full note range better than v1.

### DIFF
--- a/lib/pbio/drv/sound/beep_sampled.c
+++ b/lib/pbio/drv/sound/beep_sampled.c
@@ -1,0 +1,55 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2025 The Pybricks Authors
+
+// Converts beeps into an array of samples
+
+#include <pbdrv/config.h>
+
+#if PBDRV_CONFIG_SOUND_BEEP_SAMPLED
+
+#include <stddef.h>
+#include <stdint.h>
+
+#include <pbdrv/sound.h>
+#include <pbio/util.h>
+
+static uint16_t waveform_data[128];
+
+static void pbdrv_sound_generate_square_wave(uint16_t sample_attenuator) {
+    uint16_t lo_amplitude_value = INT16_MAX - sample_attenuator;
+    uint16_t hi_amplitude_value = sample_attenuator + INT16_MAX;
+
+    size_t i = 0;
+    for (; i < PBIO_ARRAY_SIZE(waveform_data) / 2; i++) {
+        waveform_data[i] = lo_amplitude_value;
+    }
+    for (; i < PBIO_ARRAY_SIZE(waveform_data); i++) {
+        waveform_data[i] = hi_amplitude_value;
+    }
+}
+
+// For 0 frequencies that are just flat lines.
+static void pbdrv_sound_generate_line_wave(void) {
+    for (size_t i = 0; i < PBIO_ARRAY_SIZE(waveform_data); i++) {
+        waveform_data[i] = INT16_MAX;
+    }
+}
+
+void pbdrv_beep_start(uint32_t frequency, uint16_t sample_attenuator) {
+    if (frequency == 0) {
+        pbdrv_sound_generate_line_wave();
+    } else {
+        pbdrv_sound_generate_square_wave(sample_attenuator);
+    }
+
+    if (frequency < 64) {
+        frequency = 64;
+    }
+    if (frequency > 24000) {
+        frequency = 24000;
+    }
+
+    pbdrv_sound_start(&waveform_data[0], PBIO_ARRAY_SIZE(waveform_data), frequency * PBIO_ARRAY_SIZE(waveform_data));
+}
+
+#endif
--- a/lib/pbio/drv/sound/sound_ev3.c
+++ b/lib/pbio/drv/sound/sound_ev3.c
@@ -5,18 +5,131 @@
 
 #if PBDRV_CONFIG_SOUND_EV3
 
+#include <math.h>
 #include <stdint.h>
 
-void pbdrv_sound_stop() {
+#include <pbdrv/gpio.h>
 
+#include <tiam1808/ehrpwm.h>
+#include <tiam1808/hw/soc_AM1808.h>
+#include <tiam1808/hw/hw_syscfg0_AM1808.h>
+#include <tiam1808/hw/hw_types.h>
+#include <tiam1808/psc.h>
+
+#include "../drv/gpio/gpio_ev3.h"
+
+// This module covers the frequency range from 64 Hz to 10 kHz.
+// It uses a maximum duty cycle of 1/16, and it controls volume
+// by shortening the duty cycle. The hardware has a resolution
+// of 16 bits for the period counter, and we want 8 bits for the volume,
+// so we use the following timebase division factors to make this work:
+//
+// 64 Hz - 900 Hz   => /40
+// 900 Hz - 8 kHz   => /4
+// 8 kHz - 10 kHz   => /1
+
+// Audio amplifier enable
+static const pbdrv_gpio_t pin_sound_en = PBDRV_GPIO_EV3_PIN(13, 3, 0, 6, 15);
+// Audio output pin
+#define SYSCFG_PINMUX3_PINMUX3_7_4_GPIO0_0 0
+static const pbdrv_gpio_t pin_audio = PBDRV_GPIO_EV3_PIN(3, 7, 4, 0, 0);
+
+void pbdrv_sound_stop() {
+    // Force the output low
+    EHRPWMAQContSWForceOnB(SOC_EHRPWM_0_REGS, EHRPWM_AQCSFRC_CSFB_LOW, EHRPWM_AQSFRC_RLDCSF_IMMEDIATE);
+    // Clean up counter
+    HWREGH(SOC_EHRPWM_0_REGS + EHRPWM_TBCTL) |= EHRPWM_TBCTL_CTRMODE_STOPFREEZE;
+    EHRPWMWriteTBCount(SOC_EHRPWM_0_REGS, 0);
 }
 
 void pbdrv_beep_start(uint32_t frequency, uint16_t sample_attenuator) {
+    // Clamp the frequency into the supported range
+    if (frequency < 64) {
+        frequency = 64;
+    }
+    if (frequency > 10000) {
+        frequency = 10000;
+    }
 
+    // Clamp the volume into the supported range
+    if (sample_attenuator > INT16_MAX) {
+        sample_attenuator = INT16_MAX;
+    }
+    // Extract bits[14:7] for our 8-bit volume resolution
+    sample_attenuator >>= 7;
+
+    // Configure the timebase depending on which bucket the frequency is in.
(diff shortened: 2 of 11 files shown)


lib/pbio/drv/sound/beep_sampled.c
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2025 The Pybricks Authors
+
+// Converts beeps into an array of samples
+
+#include <pbdrv/config.h>
+
+#if PBDRV_CONFIG_SOUND_SAMPLED_ONLY
```
dlech, line 8:
```suggestion
#if PBDRV_CONFIG_SOUND_BEEP_SAMPLED
```
It would be preferable to make the config option match the file name.

lib/pbio/drv/sound/sound_ev3.c
```diff
@@ -0,0 +1,96 @@
...
+void pbdrv_beep_start(uint32_t frequency, uint16_t sample_attenuator) {
+    // This hardware doesn't implement volume control
+    (void)sample_attenuator;
+
+    if (frequency < 64) {
+        frequency = 64;
+    }
+    if (frequency > 24000) {
+        frequency = 24000;
+    }
```
dlech, line 48: In ev3dev our limits were 100 and 10000. It is probably fine to go a bit lower here, but not much point in going higher as I think we start losing volume much above that.

lib/pbio/drv/sound/sound_ev3.c
```diff
@@ -0,0 +1,96 @@
...
+
+    // The stop function performs various initializations
+    pbdrv_sound_stop();
+
+    // Program a timebase division factor of /40
+    // Don't use EHRPWMTimebaseClkConfig because its calculation algorithm
+    // isn't very good and is also tricky to fix.
+    HWREGH(SOC_EHRPWM_0_REGS + EHRPWM_TBCTL) = (HWREGH(SOC_EHRPWM_0_REGS + EHRPWM_TBCTL) & ~(EHRPWM_TBCTL_CLKDIV | EHRPWM_TBCTL_HSPCLKDIV)) |
+        (EHRPWM_TBCTL_CLKDIV_DIVBY4 << EHRPWM_TBCTL_CLKDIV_SHIFT) |
+        (EHRPWM_TBCTL_HSPCLKDIV_DIVBY10 << EHRPWM_TBCTL_HSPCLKDIV_SHIFT);
```
dlech, line 72: We are probably going to need to dynamically set this to be able to get decent volume control across the frequency range.

lib/pbio/drv/sound/sound_ev3.c
```diff
@@ -0,0 +1,96 @@
...
+#include <tiam1808/hw/hw_syscfg0_AM1808.h>
+#include <tiam1808/hw/hw_types.h>
+#include <tiam1808/psc.h>
+
+#include "../drv/gpio/gpio_ev3.h"
+
+// The division factor of 40 gives decent coverage of the frequency range
+// 64 Hz - 24 kHz that the other drivers support.
+// (This is limited by the 16-bit period counter.)
+#define PWM_TIMEBASE_FREQ       (SOC_EHRPWM_0_MODULE_FREQ / 40)
```
dlech, line 23: Not necessarily for this pull request, but we need to have a look at the clock tree.

In the official LEGO firmware and in ev3dev, we have the ASYNC3 clock mux selecting SYSCLK2 from PLL1 rather than PLL2. It is 132 MHz instead of 150MHz.

lib/pbio/drv/sound/sound_ev3.c
```diff
@@ -0,0 +1,96 @@
...
+void pbdrv_sound_stop() {
+    // Turn speaker amplifier off
+    pbdrv_gpio_out_low(&pin_sound_en);
+    // Clean up counter
+    HWREGH(SOC_EHRPWM_0_REGS + EHRPWM_TBCTL) |= EHRPWM_TBCTL_CTRMODE_STOPFREEZE;
+    EHRPWMWriteTBCount(SOC_EHRPWM_0_REGS, 0);
+}
+
+void pbdrv_beep_start(uint32_t frequency, uint16_t sample_attenuator) {
+    // This hardware doesn't implement volume control
```
dlech, line 40: This is how we handle the volume control in ev3dev: https://github.com/ev3dev/lego-linux-drivers/blob/1b387f3bacb4ab8f623494d29a855de6163c8dec/ev3/legoev3_sound.c#L138-L151

lib/pbio/include/pbdrv/sound.h
```diff
@@ -17,6 +17,15 @@
 
 #if PBDRV_CONFIG_SOUND
 
+/**
+ * Starts playing a square wave until pbdrv_sound_stop() is called.
+ *
+ * @param [in]  frequency           The frequency of the wave in Hz
+ * @param [in]  sample_attenuator   The amplitude of the wave in sample LSBs
```
dlech, line 24:
```suggestion
 * @param [in]  sample_attenuator   The normalized attenuation to apply to get the requested volume.
```

We can't control the amplitude of the wave on EV3 should make the language a bit more general.

ArcaneNibble:
Made some changes, but I don't quite understand how the volume control works. Could you please explain it?

lib/pbio/drv/sound/sound_ev3.c
dlech: With a sufficiently narrow PWM pulse, I think the low pass filter starts to kick in and thus attenuates the signal and decreases the volume.

lib/pbio/drv/sound/sound_ev3.c
```diff
@@ -0,0 +1,125 @@
...
+    } else {
+        timebase_div = 1;
+        HWREGH(SOC_EHRPWM_0_REGS + EHRPWM_TBCTL) = (HWREGH(SOC_EHRPWM_0_REGS + EHRPWM_TBCTL) &
+            ~(EHRPWM_TBCTL_CLKDIV | EHRPWM_TBCTL_HSPCLKDIV)) |
+            (EHRPWM_TBCTL_CLKDIV_DIVBY1 << EHRPWM_TBCTL_CLKDIV_SHIFT) |
+            (EHRPWM_TBCTL_HSPCLKDIV_DIVBY1 << EHRPWM_TBCTL_HSPCLKDIV_SHIFT);
+    }
+
+    uint32_t pwm_period = (SOC_EHRPWM_0_MODULE_FREQ / timebase_div + frequency / 2) / frequency;
+    uint32_t pwm_duty_cycle = (pwm_period * sample_attenuator / 256) / 16;
```
dlech, line 85: I have a sound level meter I've been using to try to get an objective measurement on how well this works. Ideally 50% volume would be approx 10dB less than max volume and the sound level all of the max volumes would be about the same for any given frequency (using C weighting to approx human perception).

This seems to be working well in the mid ranges (500 - 1500 Hz), but I found that the low frequencies (100 Hz) got to max volume at about 50% requested volume and the high frequencies (5000Hz) never got near the same max volume. so I think we need a bit of an adjustment based on the frequency. Playing around, this got something decent but could probably use a bit more tuning:

```
    uint32_t freq_adj = powf(2, log10f(frequency) - 3) * 256;

    uint32_t pwm_period = (SOC_EHRPWM_0_MODULE_FREQ / timebase_div + frequency / 2) / frequency;
    uint32_t pwm_duty_cycle = (pwm_period * sample_attenuator / 256) / 16 * freq_adj / 256;
```

lib/pbio/drv/sound/sound_ev3.c
```diff
@@ -0,0 +1,125 @@
...
+
+// Audio amplifier enable
+static const pbdrv_gpio_t pin_sound_en = PBDRV_GPIO_EV3_PIN(13, 3, 0, 6, 15);
+// Audio output pin
+#define SYSCFG_PINMUX3_PINMUX3_7_4_GPIO0_0 0
+static const pbdrv_gpio_t pin_audio = PBDRV_GPIO_EV3_PIN(3, 7, 4, 0, 0);
+
+void pbdrv_sound_stop() {
+    // Turn speaker amplifier off
+    pbdrv_gpio_out_low(&pin_sound_en);
```
dlech, line 38: There is a popping noise from the speaker every time we turn off the amplifier. So we should probably either leave it on all of the time or only turn it off after some timeout period. Otherwise, when we do beeps, there is a pop after every beep.

Or maybe there is some way we can program the PWM, e.g. 50kHz 50% duty cycle to get "0V" or invert the PWM output so that 0% duty cycle is high instead of low, that will minimize the popping sound?

ArcaneNibble: Listening carefully, it's not from turning *off* the amplifier, it's from turning *on* the amplifier.

*sigh* classic audio problems. We should probably indeed just leave it on and program the PWM controller to mute instead.

lib/pbio/drv/sound/sound_ev3.c
ArcaneNibble: Let's just go with this for now.

dlech:
Once again, thanks for the excellent work!

BertLindeman:
running ```ci-build-4014-v3.6.1-281-g16e7a481``` to test the speaker.
Entered this here as it seemed appropriate.
Can make a new issue if that is better.

The "REST" sound is a nasty BRRRR.

Also tested with:
```python

# from pybricks.hubs import EV3Brick

# hub = EV3Brick()
hub = ThisHub()

hub.speaker.volume(100)
notes = []
for octave in range(2, 9):  # 2 to 8
    for note in ["A", "B", "C", "D", "E", "F", "G"]:
        notes.append(note + str(octave) +"/8")
    notes.append("R/1")
for n in range(len(notes) / 8):
    print(notes[n*8:n*8+8])
    hub.speaker.play_notes(notes[n*8:n*8+8])
wait(500)
```

Could not upload the m3 files, so zipped them:
[EV3 speaker.zip](https://github.com/user-attachments/files/21327293/EV3.speaker.zip)

It may be my hearing but it seems on the primehub that on each octave from the A to the B note the pitch increases and to C it drops a bit and then the pitch goes up again.
Not sure though, do not have a musical ear.

ArcaneNibble:
Oh that's fun. Apparently a rest plays a frequency of 0 Hz expecting that to turn the sound off, but that function wasn't implemented.

As for the pitch jumping around, that's just because the API is excessively music-theory-brained:

> The octave changes to the next number at the note C, for example, B3 is the note below middle C (C4).

So A3 and B3 come after G3 and are *far* above C3.

BertLindeman:
Thanks, for the music theory, I am such a noob in that. 
Asked 3 A.I. assistants and they confirmed my "program" was correct going up in pitch.

So I should use C D E F G A B per octave.

Sounds better!
And ~ nothing wrong in my hearing although the last notes of octave 7 and all of octave 8 I hear just the clicks.

PR #343, opened 2025-07-16
[EV3] Implement reset via watchdog timer

ArcaneNibble:
Implements both resetting as well as booting into firmware update mode.

~~Note that the ev3dev documentation is incorrect. The correct magic flag address is 0xFFFF1FF**C** and not 0xFFFF1FF**A**.~~ This issue has been corrected in the ev3dev documentation.

### DIFF
--- a/lib/pbio/drv/reset/reset_ev3.c
+++ b/lib/pbio/drv/reset/reset_ev3.c
@@ -12,7 +12,32 @@
 
 #include "../drv/gpio/gpio_ev3.h"
 
+#include <tiam1808/hw/soc_AM1808.h>
 #include <tiam1808/hw/hw_syscfg0_AM1808.h>
+#include <tiam1808/hw/hw_types.h>
+#include <tiam1808/timer.h>
+
+#define BOOTLOADER_UPDATE_MODE_VALUE    0x5555AAAA
+
+typedef struct {
+    // 0xffff1ff0
+    uint32_t _dummy0;
+    // 0xffff1ff4
+    // Pybricks uses this flag to determine software reset vs other resets
+    uint32_t reset_reason_flag;
+    // 0xffff1ff8
+    // Have not fully investigated this, but the bootloader seems to store
+    // the result of (DDR) memory testing at 0xffff1ffa and 0xffff1ffb
+    uint32_t _bootloader_unk_ram_test;
+    // 0xffff1ffc
+    uint32_t bootloader_update_flag;
+} persistent_data_t;
+// This is defined as an extern variable so that its address can be specified
+// in the platform.ld linker script. This means that the linker script
+// contains information about *all* fixed memory locations.
+//
+// This lives at the very end of the ARM local RAM.
+extern volatile persistent_data_t ev3_persistent_data;
 
 static const pbdrv_gpio_t poweroff_pin = PBDRV_GPIO_EV3_PIN(13, 19, 16, 6, 11);
 
@@ -23,11 +48,17 @@ void pbdrv_reset_init(void) {
 void pbdrv_reset(pbdrv_reset_action_t action) {
     for (;;) {
         switch (action) {
-            case PBDRV_RESET_ACTION_RESET_IN_UPDATE_MODE:
-                // TODO
+            case PBDRV_RESET_ACTION_POWER_OFF:
+                pbdrv_reset_power_off();
                 break;
-            // TODO: implement case PBDRV_RESET_ACTION_RESET
+            case PBDRV_RESET_ACTION_RESET_IN_UPDATE_MODE:
+                ev3_persistent_data.bootloader_update_flag = BOOTLOADER_UPDATE_MODE_VALUE;
+                __attribute__((fallthrough));
             default:
+                // PBDRV_RESET_ACTION_RESET
+
+                // Poke the watchdog timer with a bad value to immediately trigger it
+                HWREG(SOC_TMR_1_REGS + TMR_WDTCR) = 0;
                 break;
         }
     }
--- a/lib/pbio/drv/reset/reset_ev3.c
+++ b/lib/pbio/drv/reset/reset_ev3.c
@@ -7,6 +7,8 @@
 
 #if PBDRV_CONFIG_RESET_EV3
 
+#include <stdbool.h>
+
 #include <pbdrv/reset.h>
 #include <pbdrv/gpio.h>
 
@@ -18,6 +20,11 @@
 #include <tiam1808/timer.h>
 
 #define BOOTLOADER_UPDATE_MODE_VALUE    0x5555AAAA
+
+// 'Pybr'
+#define RESET_REASON_FLAG_WDT           0x50796272
+// 'rboo'
+#define RESET_REASON_FLAG_SOFT_RESET    0x72626f6f
 
 typedef struct {
     // 0xffff1ff0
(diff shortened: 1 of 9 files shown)


lib/pbio/drv/reset/reset_ev3.c
```diff
@@ -23,17 +30,31 @@ void pbdrv_reset_init(void) {
...
         switch (action) {
-            case PBDRV_RESET_ACTION_RESET_IN_UPDATE_MODE:
-                // TODO
+            case PBDRV_RESET_ACTION_POWER_OFF:
+                pbdrv_reset_power_off();
                 break;
-            // TODO: implement case PBDRV_RESET_ACTION_RESET
+            case PBDRV_RESET_ACTION_RESET_IN_UPDATE_MODE:
+                *(volatile uint32_t *)BOOTLOADER_UPDATE_MODE_ADDR = BOOTLOADER_UPDATE_MODE_VALUE;
+            // Fall through
```
dlech, line 38:
```suggestion
            __attribute__((fallthrough));
```

Probably worth starting a `pbio_compiler_*` header that has macros for this, the memory barrier inline assembly, and other common things we use to give hints to the compiler.

dlech (commented):
Do you also plan on creating a watchdog driver for EV3 to implement `pbdrv_watchdog_update()`?

ArcaneNibble:
I actually didn't realize there was also a watchdog framework. I'll revisit this to take that into account.

lib/pbio/drv/reset/reset_ev3.c
ArcaneNibble: I've chosen `lib/pbio/include/pbdrv/compiler.h` since most of these are very low-level tasks most likely to be found in the driver layer.

lib/pbio/include/pbdrv/compiler.h
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2025 The Pybricks Authors
+
+#ifndef _PBDRV_COMPILER_H_
+
+// Marks a switch case that intentionally falls through to the next one
+#define FALL_THROUGH                __attribute__((fallthrough))
```
dlech, line 7:
```suggestion
#define PBDRV_FALL_THROUGH                __attribute__((fallthrough))
```

everything needs pbio prefix to avoid name clash with other libraries

lib/pbio/include/pbdrv/compiler.h
```diff
@@ -0,0 +1,12 @@
...
+
+// Forces the compiler to not reorder memory access around this line
+#define compiler_memory_barrier()     __asm__ volatile ("" ::: "memory")
```
dlech, line 10:
```suggestion
#define pbdrv_compiler_memory_barrier()     __asm__ volatile ("" ::: "memory")
```

lib/pbio/platform/ev3/platform.ld
```diff
@@ -9,11 +9,17 @@ MEMORY
...
     DDR        (rwx) : ORIGIN = 0xC0008000, LENGTH = (64M - 0x8000)
-    ARM_LRAM   (rwx) : ORIGIN = 0xFFFF0000, LENGTH = 8K
+    ARM_LRAM   (rwx) : ORIGIN = 0xFFFF0000, LENGTH = (8K - 16)
 }
 
 _minimal_stack_size = 4M;
 pbdrv_stack_end = ORIGIN(DDR) + LENGTH(DDR) - 4;
+/*
+We declare this in this style rather than creating a section.
+If we do create a section, the loader (U-boot) will always clear it.
```
dlech, line 19: Even for a NOLOAD section?

Don't have a problem doing it this way, just curious.

lib/pbio/drv/reset/reset_ev3.c
```diff
@@ -7,33 +7,85 @@
...
+    // Pybricks uses this flag to determine software reset vs other resets
+    uint32_t reset_reason_flag;
+    // 0xffff1ff8
+    // Have not fully investigated this, but the bootloader seems to store
+    // the result of (DDR) memory testing at 0xffff1ffa and 0xffff1ffb
+    uint32_t _bootloader_unk_ram_test;
+    // 0xffff1ffc
+    uint32_t bootloader_update_flag;
+} persistent_data_t;
+extern volatile persistent_data_t ev3_persistent_data;
```
dlech, line 43: Let's add a comment explaining this is declared in the linker script and that we are using the ARM LRAM to pass info.

lib/pbio/drv/reset/reset_ev3.c
```diff
@@ -7,33 +7,85 @@
...
+static bool second_init;
+static uint32_t saved_reset_reason_flag;
 
 static const pbdrv_gpio_t poweroff_pin = PBDRV_GPIO_EV3_PIN(13, 19, 16, 6, 11);
 
 void pbdrv_reset_init(void) {
     pbdrv_gpio_out_high(&poweroff_pin);
+    if (!second_init) {
+        saved_reset_reason_flag = ev3_persistent_data.reset_reason_flag;
+        // The EV3 happens to call this init function multiple times,
```
dlech, line 54: We shouldn't be. Can we remove `pbdrv_reset_init()` from `SystemInit()` and replace it with something else (e.g. just put the power pin call inline there)?

lib/pbio/platform/ev3/platform.ld
ArcaneNibble: Yeah.

If loading via section headers (the default), sections flagged as `NOBITS` are cleared: https://github.com/pybricks/u-boot/blob/pybricks-ev3/lib/elf.c#L266-L268

If loading via program headers, the difference between filesz and memsz is cleared: https://github.com/pybricks/u-boot/blob/pybricks-ev3/lib/elf.c#L219-L221

lib/pbio/platform/ev3/platform.c
```diff
@@ -487,6 +488,12 @@ void ev3_panic_handler(int except_type, ev3_panic_ctx *except_data) {
     panic_putu32(except_data->spsr);
 
     panic_puts("\r\nSystem will now reboot...\r\n");
+
+    // Poke the watchdog timer with a bad value to immediately trigger it
+    // if it has already been configured. If it has *not* been configured,
+    // that means we are crashing in early boot, and we let the jump back
+    // to the reset vector take care of rebooting the system.
+    HWREG(SOC_TMR_1_REGS + TMR_WDTCR) = 0;
```
dlech, line 496: Should we add a new `pbdrv_reset_reason_t` for panic?

ArcaneNibble: Do we need one? It currently reports as a watchdog reset

dlech: I don't know if it is worth it or not. That is why I ask. :smile: 

There is a slight difference in that the true watchdog reset only happens if the event loops stops running while a panic is a different mode of failure. So could be potentially useful for getting info on the cause of a crash. But would anyone ever actually do that? Maybe not very likely.

ArcaneNibble: Doesn't seem too likely, since right now crash dumps aren't saved anywhere and you need to have the debug console connected to see them.

Perhaps we can change "watchdog" to a generic "fatal error"?

dlech: OK, we'll leave it the way it is. I would prefer to stick with calling it watchdog and add a new reason if we need to differentiate.

dlech:
I rebased this and fixed the conflicts.

PR #344, opened 2025-07-17
Use separate heap for image data on EV3

dlech:
In https://github.com/pybricks/support/issues/2154#issuecomment-3046648358 (cc: @schodet), it came up that we were running out of MicroPython heap space on EV3 when trying to allocate full-screen sized images. We really don't want to use MicroPython heap for this because the image data will contain no pointers to MicroPython objects (other than false positives) and will make allocating and collecting memory in MicroPython much slower than it needs to be.

Instead, we can add a new memory allocator that doesn't use garbage collection. I've picked one created by our old MINDSTORMS pal @rhempel that seem well suited to the task. In addition to being designed for microcontrollers, it also has a nice debug print feature which I have wired up to `hub.system.umm_info()` that works similar to `micropython.mem_info()` that will help us debug memory issues if needed.

The block size and heap size was just an educated guess, but we can tune those as needed based on real-world usage.


laurensvalk:
1. Could we perhaps also use this opportunity to set the following to sensible values? These are just pseudorandom values to get things to build initially. It's no surprise that this was quickly running our of MicroPython heap.

```
#define PBDRV_CONFIG_BLOCK_DEVICE_RAM_SIZE          (10 * 1024)
#define PBDRV_CONFIG_BLOCK_DEVICE_EV3_SIZE          (8 * 1024)
#define PBDRV_CONFIG_BLOCK_DEVICE_EV3_SIZE_USER     (512)
#define PBDRV_CONFIG_BLOCK_DEVICE_EV3_START_ADDRESS (10 * 1024 * 1024)
```

2. While we are working on the `Image` class, it would be nice to also move it to its intended destination in `pybricks.parameters.Image`. This follows the pattern established by `Icon` and `Color` and it ensures we don't have to have a module for just one or two classes. See https://github.com/pybricks/support/issues/2154#issuecomment-3067993338 for rationale.

rhempel:
You were probably running out of contiguous heap space for large objects on the MicroPython heap :-)

umm_malloc() has the same problem, so a simple pool of image buffers would have been fine, but less general.

Of course I'm happy that you were able to find umm_malloc useful - it's surprising where that project ends up sometimes.

I think the OpenMV cameras used it for a while.

--
  Ralph Hempel
  ***@***.***


On Thu, Jul 17, 2025, at 3:35 AM, Laurens Valk wrote:
> *laurensvalk* left a comment (pybricks/pybricks-micropython#344) <https://github.com/pybricks/pybricks-micropython/pull/344#issuecomment-3082949996>
>  1. Could we perhaps also use this opportunity to set the following to sensible values? These are just pseudorandom values to get things to build initially. It's no surprise that this was quickly running our of MicroPython heap.
> `#define PBDRV_CONFIG_BLOCK_DEVICE_RAM_SIZE          (10 * 1024)
> #define PBDRV_CONFIG_BLOCK_DEVICE_EV3_SIZE          (8 * 1024)
> #define PBDRV_CONFIG_BLOCK_DEVICE_EV3_SIZE_USER     (512)
> #define PBDRV_CONFIG_BLOCK_DEVICE_EV3_START_ADDRESS (10 * 1024 * 1024)
` 1. While we are working on the `Image` class, it would be nice to also move it to its intended destination in `pybricks.parameters.Image`. This follows the pattern established by `Icon` and `Color` and it ensures we don't have to have a module for just one or two classes. See pybricks/support#2154 (comment) <https://github.com/pybricks/support/issues/2154#issuecomment-3067993338> for rationale.

schodet:
Looks good, thanks!

dlech:
> 1. Could we perhaps also use this opportunity to set the following to sensible values?

I'm not really sure what sensible values would be at this time. I think we should figure out how we plan to use all of the flash and RAM (make a complete map of both), then we can make a better educated guess.

> 2\. it would be nice to also move it to its intended destination in `pybricks.parameters.Image`

Done.

PR #345, opened 2025-07-17
pybricks/ev3devices: Enable remaining device support

ArcaneNibble:
As mentioned on the call earlier today, I decided to test if this module was already working or not. It would appear that it is, at least for basic distance measurements. I don't have a remote or any way to test the other functionality,

### DIFF
--- a/pybricks/ev3devices.h
+++ b/pybricks/ev3devices.h
@@ -12,15 +12,10 @@
 
 extern const mp_obj_type_t pb_type_ev3devices_TouchSensor;
 extern const mp_obj_type_t pb_type_ev3devices_ColorSensor;
+extern const mp_obj_type_t pb_type_ev3devices_InfraredSensor;
+extern const mp_obj_type_t pb_type_ev3devices_UltrasonicSensor;
+extern const mp_obj_type_t pb_type_ev3devices_GyroSensor;
 
 #endif // PYBRICKS_PY_EV3DEVICES
-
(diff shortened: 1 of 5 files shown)


ArcaneNibble:
I've enabled all the `ev3devices` sensors in the build, since they do indeed all compile.

pybricks/ev3devices.h
```diff
@@ -12,15 +12,10 @@
...
 extern const mp_obj_type_t pb_type_ev3devices_ColorSensor;
-
-#endif // PYBRICKS_PY_EV3DEVICES
-
-#if PYBRICKS_PY_EV3DEVDEVICES
-
 extern const mp_obj_type_t pb_type_ev3devices_InfraredSensor;
-extern const mp_obj_type_t pb_type_ev3devices_GyroSensor;
 extern const mp_obj_type_t pb_type_ev3devices_UltrasonicSensor;
+extern const mp_obj_type_t pb_type_ev3devices_GyroSensor;
```
dlech, line 17: Why changing the order?

laurensvalk:
These haven't been updated to the async API yet, so they haven't been enabled yet. I'll see if I can get around to the InfraredSensor soon so you have something to test with :robot: 

The gyro sensor needs some work too. I've already enabled the necessary quirks at the protocol level, but haven't exposed them yet. It's on my list :slightly_smiling_face:  Thanks!

laurensvalk:
Thanks Bert. We'll get there pretty soon! :smile:

PR #346, opened 2025-07-17
pybricks/experimental: Add functions to reboot or intentionally crash the brick

ArcaneNibble:
I expect that this might be somewhat controversial to have in final shipping firmware, but this was very useful for debugging I was doing.

### DIFF
--- a/pybricks/experimental/pb_module_experimental.c
+++ b/pybricks/experimental/pb_module_experimental.c
@@ -11,6 +11,7 @@
 #include "py/runtime.h"
 #include "py/mperrno.h"
 
+#include <pbdrv/reset.h>
 #include <pbio/util.h>
 
 #include <pybricks/util_mp/pb_obj_helper.h>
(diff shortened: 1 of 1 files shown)


ArcaneNibble:
I do think that having "reboot" and "power off" in shipping firmware would be useful, but it would have to be clearly documented that they indeed do what they say (and thus terminates user programs unexpectedly without saving data).

BertLindeman:
We **do** have `hub.system.shutdown()`

dlech:
We use to have `hub.system.reset()` but it got removed in https://github.com/pybricks/pybricks-micropython/commit/a2e3bd7c30101ae812609a683ae8d9625046928e. As it says in the commit message, we can reset via Bluetooth (or soon USB), so that seemed good enough since most users won't every need this.

laurensvalk:
```python
from universe import destroy

destroy(password="So Long, and Thanks for All the Fish!")
```

PR #347, opened 2025-07-17
EV3 SD card support

ArcaneNibble:
This is *far* from complete, but I wanted to make it known that I've hacked on some basic SD card support just to get an understanding of how the MMC/SD controller works.

This initialized a SD*HC* card (no legacy SD support) and prints out a bunch of its internal registers as debug spam.


ArcaneNibble:
Also, I just tried an SD*XC* card and it does seem to work just fine (goes through all of the initialization steps, but I haven't actually tried reading any sectors yet).

The hardware peripheral has no knowledge of what SD card capacity you've inserted -- this is entirely under software control. Data corruption (with ev3dev? with the stock firmware?) is probably just a software bug.

PR #349, opened 2025-07-19
USB stdio v2

dlech:
The less ambitious alternative to https://github.com/pybricks/pybricks-micropython/pull/332.

This time, instead of duplicating a bunch of stuff, we finish the TODO to move the stdin buffer out of pbsys_bluetooth and into pbsys_host. Then we can hook up the USB stdin to this as well.


laurensvalk:
Thanks!

PR #351, opened 2025-07-19
bricks/common/mpconfigport: enable f-strings on level 2 hubs

dlech:
Enable f-strings on level 2 hubs. This is a very useful feature, but also big, so we only enable it on level 2 hubs.


BertLindeman:
> Besides the REPL (where using f-strings should be relatively rare), what are the additional things we could now do?

In micropython f-strings in the REPL are not widely implemented. 
Sometimes you want to do a quick test in the REPL and then bump into: no f-strings. 
It is more important to have f-strings on all hubs, but the REPL could do without I think. 

But like David states, nice to have. 
(Hm, rather not so helpful remark, sorry)

dlech:
> Besides the REPL (where using f-strings should be relatively rare), what are the additional things we could now do?

I was using paste-mode on the REPL since we don't have download and run on the EV3 yet. 

If we ever get a "real" file system and people can load .py files rather than just .mpy, then this would be useful as well.

PR #352, opened 2025-07-19
pbio/sys/storage_data: fix strict aliasing issues

dlech:
Fix strict aliasing issues between block drivers and pbsys storage.

Casting a uint8_t* to a struct pointer violates strict aliasing rules, which can lead to hard-to-find bugs. We need to ensure that the memory being pointed to has the same alignment, etc. as the struct it is being cast to. The usual way to do this is to use a union of the struct and the other type we want to use (in this case, uint8_t array).

To do this, we need to move the storage data struct definition to a header file that can be included by both pbdrv/block_device and pbsys/storage.

Additionally, in block_device_test, the size of the ramdisk block was wrong and was causing the NXT to crash sometime after starting the REPL. It also incorrectly had the .noinit section attribute, but depends on being initialized to zero, so that is removed.

Fixes: https://github.com/pybricks/support/issues/2272
Fixes: ce4253aec ("pbio/sys: Move block device read to driver level.")

### DIFF
--- a/lib/pbio/drv/block_device/block_device_ev3.c
+++ b/lib/pbio/drv/block_device/block_device_ev3.c
@@ -20,13 +20,6 @@
 #include <stdint.h>
 #include <string.h>
 
-#include "../core.h"
-
-#include <pbdrv/block_device.h>
-#include <pbdrv/clock.h>
-#include <pbdrv/compiler.h>
-#include <pbdrv/gpio.h>
-
 #include <tiam1808/edma.h>
 #include <tiam1808/spi.h>
 #include <tiam1808/psc.h>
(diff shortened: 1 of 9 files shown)


BertLindeman:
This fix helps to have a more stable REPL on the NXT, but the EV3 now suffers something like [issue 2272](https://github.com/pybricks/support/issues/2272)

Went back too github hash 96cdd12b to get a usable REPL.

dlech:
> Is there a way to do this without including `pbsys` from `pbdrv`?

We could, but we would lose the strict safety that this provides.

dlech:
Maybe it works to move the union to the pbsys side of things? We would probably need to change the array type from uint8_t to uint32_t though to get the right alignment in the block drivers. And when we have multiple slots, we need to make sure each slot is 4-byte aligned as well. I'll see if I can come up with something.

PR #360, opened 2025-07-29
[EV3] Implement I2C support via PRU1

ArcaneNibble:


ArcaneNibble:
We should probably have a "PRU1" driver to handle shared init, *and* we should figure out how to coordinate the PRU->ARM interrupt between PRU0 and PRU1

dlech:
> We should probably have a "PRU1" driver to handle shared init

Yes, a new `rproc` folder is what I had in mind (for remote processor).

> we should figure out how to coordinate the PRU->ARM interrupt between PRU0 and PRU1

Probably `ev3/platform.c` is the right place for this kind of shared resource.

dlech:
> > we should figure out how to coordinate the PRU->ARM interrupt between PRU0 and PRU1
> 
> Probably `ev3/platform.c` is the right place for this kind of shared resource.

Although there are 8 different `PRU_EVTOUTx` interrupts, so do we really need to share between the two PRUs? Or do you mean something else?

ArcaneNibble:
I was referring to the *setting up* of the interrupt channel mapping ( https://github.com/pybricks/pybricks-micropython/blob/master/lib/pbio/drv/uart/uart_ev3_pru_lib/suart_api.c#L2324 )

dlech:
Ah, I would pull out all of the code for the PRU's interrupt controller to the new PRU rproc driver as well. Since we don't need to make this general purpose, we can just do all of the setup for both PRUs in one place and hard-code it. 

The various drivers that depend on the PRUs can then just wait for a "kick" from the PRU when it is actually up and running to know when they can actually start using the PRU.

ArcaneNibble:
Note that I am currently testing with this patch, because I2CDevice isn't ported yet:

```diff
diff --git a/lib/pbio/drv/i2c/i2c_ev3.c b/lib/pbio/drv/i2c/i2c_ev3.c
index 853994a8..5bad271f 100644
--- a/lib/pbio/drv/i2c/i2c_ev3.c
+++ b/lib/pbio/drv/i2c/i2c_ev3.c
@@ -13,6 +13,7 @@
 #include <string.h>
 
 #include <tiam1808/armv5/am1808/interrupt.h>
+#include <tiam1808/hw/hw_syscfg0_AM1808.h>
 #include <tiam1808/hw/hw_types.h>
 #include <tiam1808/pruss.h>
 
@@ -22,8 +23,10 @@
 #include <pbio/util.h>
 
 #include <pbdrv/cache.h>
+#include <pbdrv/gpio.h>
 #include <pbdrv/i2c.h>
 
+#include "../drv/gpio/gpio_ev3.h"
 #include "../drv/rproc/rproc.h"
 #include "../rproc/rproc_ev3.h"
 
@@ -171,9 +174,115 @@ pbio_error_t pbdrv_i2c_placeholder_operation(
     PBIO_OS_ASYNC_END(PBIO_SUCCESS);
 }
 
+static pbdrv_gpio_t test_2_scl = PBDRV_GPIO_EV3_PIN(0, 15, 12, 0, 12);
+static pbdrv_gpio_t test_2_sda = PBDRV_GPIO_EV3_PIN(2, 7, 4, 1, 14);
+static pbdrv_gpio_t test_3_scl = PBDRV_GPIO_EV3_PIN(1, 27, 24, 0, 1);
+static pbdrv_gpio_t test_3_sda = PBDRV_GPIO_EV3_PIN(2, 3, 0, 1, 15);
+
+static pbio_os_process_t ev3_i2c_wip_ultrasonic_process;
+static pbio_os_process_t ev3_i2c_wip_rpi_pico_process;
+
+pbio_error_t ev3_i2c_wip_ultrasonic_process_thread(pbio_os_state_t *state, void *context) {
+    static pbio_os_timer_t timer;
+    static pbio_os_state_t sub;
+    static int i;
+    static uint8_t buf[7];
+
+    PBIO_OS_ASYNC_BEGIN(state);
+
+    while (1) {
+        memset(buf, 0, sizeof(buf));
+        buf[0] = 0x10;
+
+        pbio_error_t err;
+        PBIO_OS_AWAIT(state, &sub, err = pbdrv_i2c_placeholder_operation(
+            &sub,
+            &i2c_devs[2],
+            0x01,
+            &buf[0],
+            1,
+            &buf[1],
+            6,
+            true
+            ));
+
+        if (err) {
+            debug_pr("i2c ultrasonic err %d\r\n", err);
+        } else {
+            for (i = 0; i < 6; i++) {
+                debug_pr("i2c ultrasonic get %02x\r\n", buf[1 + i]);
+            }
+        }
+
+        PBIO_OS_AWAIT_MS(state, &timer, 1000);
+    }
+
+    PBIO_OS_ASYNC_END(PBIO_SUCCESS);
+}
+
+pbio_error_t ev3_i2c_wip_rpi_pico_process_thread(pbio_os_state_t *state, void *context) {
+    static pbio_os_timer_t timer;
+    static pbio_os_state_t sub;
+    static int i;
+    static uint8_t buf[7];
+
+    PBIO_OS_ASYNC_BEGIN(state);
+
+    while (1) {
+        memset(buf, 0, sizeof(buf));
+        buf[0] = 0x10;
+        buf[1] = 0xaa;
+        buf[2] = 0x55;
+        buf[3] = 0x12;
+        buf[4] = 0x34;
+
+        pbio_error_t err;
+        PBIO_OS_AWAIT(state, &sub, err = pbdrv_i2c_placeholder_operation(
+            &sub,
+            &i2c_devs[3],
+            0x17,
+            &buf[0],
+            5,
+            0,
+            0,
+            false
+            ));
+
+        if (err) {
+            debug_pr("i2c rpi pico #1 err %d\r\n", err);
+        }
+
+        buf[0] = 0x0e;
+        PBIO_OS_AWAIT(state, &sub, err = pbdrv_i2c_placeholder_operation(
+            &sub,
+            &i2c_devs[3],
+            0x17,
+            &buf[0],
+            1,
+            &buf[1],
+            6,
+            false
+            ));
+
+        if (err) {
+            debug_pr("i2c rpi pico #2 err %d\r\n", err);
+        } else {
+            for (i = 0; i < 6; i++) {
+                debug_pr("i2c rpi pico get %02x\r\n", buf[1 + i]);
+            }
+        }
+
+        PBIO_OS_AWAIT_MS(state, &timer, 1000);
+    }
+
+    PBIO_OS_ASYNC_END(PBIO_SUCCESS);
+}
+
 static pbio_os_process_t ev3_i2c_init_process;
 
 pbio_error_t ev3_i2c_init_process_thread(pbio_os_state_t *state, void *context) {
+    static pbio_os_timer_t timer;
+
     PBIO_OS_ASYNC_BEGIN(state);
 
     // Need rproc to be initialized, because it sets up the PRU INTC
@@ -209,6 +318,22 @@ pbio_error_t ev3_i2c_init_process_thread(pbio_os_state_t *state, void *context)
 
     pbio_busy_count_down();
 
+
+
+    pbdrv_gpio_out_low(&test_2_scl);
+    pbdrv_gpio_input(&test_2_scl);
+    pbdrv_gpio_out_low(&test_2_sda);
+    pbdrv_gpio_input(&test_2_sda);
+    pbdrv_gpio_out_low(&test_3_scl);
+    pbdrv_gpio_input(&test_3_scl);
+    pbdrv_gpio_out_low(&test_3_sda);
+    pbdrv_gpio_input(&test_3_sda);
+
+    PBIO_OS_AWAIT_MS(state, &timer, 200);
+
+    pbio_os_process_start(&ev3_i2c_wip_ultrasonic_process, ev3_i2c_wip_ultrasonic_process_thread, NULL);
+    pbio_os_process_start(&ev3_i2c_wip_rpi_pico_process, ev3_i2c_wip_rpi_pico_process_thread, NULL);
+
     PBIO_OS_ASYNC_END(PBIO_SUCCESS);
 }
 
diff --git a/lib/pbio/platform/ev3/platform.c b/lib/pbio/platform/ev3/platform.c
index 42add184..fbbe3143 100644
--- a/lib/pbio/platform/ev3/platform.c
+++ b/lib/pbio/platform/ev3/platform.c
@@ -367,7 +367,7 @@ const pbdrv_ioport_platform_data_t pbdrv_ioport_platform_data[PBDRV_CONFIG_IOPOR
             .adc_p1 = 10,
             .adc_p6 = 9,
         },
-        .supported_modes = PBIO_PORT_MODE_UART | PBIO_PORT_MODE_LEGO_DCM,
+        .supported_modes = 0,
     },
     {
         .port_id = PBIO_PORT_ID_4,
@@ -389,7 +389,7 @@ const pbdrv_ioport_platform_data_t pbdrv_ioport_platform_data[PBDRV_CONFIG_IOPOR
             .adc_p1 = 12,
             .adc_p6 = 11,
         },
-        .supported_modes = PBIO_PORT_MODE_UART | PBIO_PORT_MODE_LEGO_DCM,
+        .supported_modes = 0,
     },
 };
```

lib/pbio/drv/i2c/i2c_ev3.c
```diff
@@ -31,21 +32,19 @@
...
-    const pbdrv_i2c_ev3_platform_data_t *pdata;
-    //
-    // TODO: i2c state goes here.
-    //
+    bool init;
+    uint8_t pru_i2c_idx;
 };
 
 static pbdrv_i2c_dev_t i2c_devs[PBDRV_CONFIG_I2C_EV3_NUM_DEV];
+_Static_assert(PBDRV_CONFIG_I2C_EV3_NUM_DEV <= PBDRV_RPROC_EV3_PRU1_NUM_I2C_CHANNELS);
```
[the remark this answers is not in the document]
ArcaneNibble: Well, we can make it _less_ than `PBDRV_RPROC_EV3_PRU1_NUM_I2C_CHANNELS`. I don't know if that's useful though?

lib/pbio/drv/i2c/i2c_ev3.c
```diff
@@ -105,6 +113,12 @@ pbio_error_t ev3_i2c_init_process_thread(pbio_os_state_t *state, void *context)
     // Need rproc to be initialized, because it sets up the PRU INTC
     PBIO_OS_AWAIT_UNTIL(state, pbdrv_rproc_is_ready());
 
+    // Set up the buffer pointers
+    for (int i = 0; i < PBDRV_CONFIG_I2C_EV3_NUM_DEV; i++) {
+        pbdrv_i2c_dev_t *i2c = &i2c_devs[i];
+        pbdrv_rproc_ev3_pru1_shared_ram.i2c[i].buffer = i2c->buffer;
```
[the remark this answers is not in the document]
ArcaneNibble: Right now the entire firmware assumes a mostly identity map in the MMU, including all of the other DMA-using drivers.

lib/pbio/drv/rproc/rproc_ev3_pru1.h
```diff
@@ -11,6 +11,45 @@
...
+    (((rlen) & 0xff) << 16) |                                           \
+    (((wlen) & 0xff) << 8) |                                            \
+    ((flags) & 0xff))
+
+typedef struct {
+    // bit[7:0]     status and flags
+    // bit[15:8]    write length
+    // bit[23:16]   read length
+    // bit[31:24]   device address (unshifted)
+    uint32_t flags;
```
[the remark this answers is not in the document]
ArcaneNibble: The current macro does a read/write of all fields at once. How do you do that with a bitfield?

lib/pbio/drv/i2c/i2c_ev3.c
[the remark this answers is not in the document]
dlech: Oh, right. I would still make it not configurable unless there is a good reason to do so.

lib/pbio/drv/gpio/gpio_ev3.c
```diff
@@ -31,13 +34,30 @@ static void pbdrv_gpio_alt_gpio(const pbdrv_gpio_t *gpio) {
...
+// If the pin is not in banks 0/1, the PRU is not involved.
+// Otherwise, if the PRU is initialized, do direction changes
+// via the PRU in order to prevent race conditions.
+// If the PRU is not initialized (i.e. doing direction changes
+// during early boot), also set the direction directly via
+// the ARM.
+#define PBDRV_GPIO_EV3_ARM_OWNS_GPIO_BANK(pin_index)    \
+    ((pin_index) > 32 || !pbdrv_rproc_is_ready())
+
 static void gpio_write(const pbdrv_gpio_t *gpio, uint8_t value) {
```
laurensvalk, line 46: This is currently called for every `pbdrv_gpio_out_low` and `pbdrv_gpio_out_high`, even if it is already an output. Shouldn't we skip setting the direction if it is already set, especially now that it involves asking the PRU?

lib/pbio/drv/gpio/gpio_ev3.c
```diff
@@ -31,13 +34,30 @@ static void pbdrv_gpio_alt_gpio(const pbdrv_gpio_t *gpio) {
...
     pbdrv_gpio_alt_gpio(gpio);
     uint32_t pin_index = get_pin_index(gpio);
-    GPIODirModeSet(SOC_GPIO_0_REGS, pin_index, GPIO_DIR_OUTPUT);
+    if (PBDRV_GPIO_EV3_ARM_OWNS_GPIO_BANK(pin_index)) {
+        GPIODirModeSet(SOC_GPIO_0_REGS, pin_index, GPIO_DIR_OUTPUT);
+    } else {
+        uint32_t val = 1 << (pin_index - 1);
+        pbdrv_rproc_ev3_pru1_shared_ram.gpio_bank_01_dir_clr = val;
+        while (pbdrv_rproc_ev3_pru1_shared_ram.gpio_bank_01_dir_clr) {
+            // Wait for the PRU to process the command
```
laurensvalk, line 58: I suppose this is very fast, but just wanted to check if we can afford a blocking loop. Or maybe add a comment to indicate why it isn't an issue.

See also comment above. I suppose it is even less of an issue if we don't do this over and over. The color sensor will keep toggling a pin high and low hundreds of times per second, so call this loop as many times (not sure which bank these are on).

lib/pbio/drv/gpio/gpio_ev3.c
```diff
@@ -55,7 +75,15 @@ uint8_t pbdrv_gpio_input(const pbdrv_gpio_t *gpio) {
     }
     pbdrv_gpio_alt_gpio(gpio);
     uint32_t pin_index = get_pin_index(gpio);
-    GPIODirModeSet(SOC_GPIO_0_REGS, pin_index, GPIO_DIR_INPUT);
+    if (PBDRV_GPIO_EV3_ARM_OWNS_GPIO_BANK(pin_index)) {
+        GPIODirModeSet(SOC_GPIO_0_REGS, pin_index, GPIO_DIR_INPUT);
+    } else {
```
laurensvalk, line 80: Same here, could skip this if it is already an input?

ArcaneNibble:
You're right, done.

lib/pbio/drv/gpio/gpio_ev3.c
```diff
@@ -31,13 +34,32 @@ static void pbdrv_gpio_alt_gpio(const pbdrv_gpio_t *gpio) {
...
 static void gpio_write(const pbdrv_gpio_t *gpio, uint8_t value) {
     if (!gpio) {
         return;
     }
     pbdrv_gpio_alt_gpio(gpio);
     uint32_t pin_index = get_pin_index(gpio);
-    GPIODirModeSet(SOC_GPIO_0_REGS, pin_index, GPIO_DIR_OUTPUT);
+    if (PBDRV_GPIO_EV3_ARM_OWNS_GPIO_BANK(pin_index)) {
+        GPIODirModeSet(SOC_GPIO_0_REGS, pin_index, GPIO_DIR_OUTPUT);
+    } else if (GPIODirModeGet(SOC_GPIO_0_REGS, pin_index) != GPIO_DIR_OUTPUT) {
```
laurensvalk, line 56: Shouldn't this check go as the first `if`, so we don't re-set it for pins owned by the ARM either? (I know it used to do this before.)

ArcaneNibble: I figured that code path was cheap enough (a single register write) that it wasn't worth bothering

laurensvalk:
I'll move it to a dedicated issue.

PR #361, opened 2025-07-30
[EV3] Implement Pybricks USB protocol

ArcaneNibble:
This fully implements the rest of the Pybricks USB protocol, allowing `pybricksdev` to run.

Screen Shot 2025-07-30 at 10 56 25 PM

### DIFF
--- a/lib/pbio/drv/usb/usb_ev3.c
+++ b/lib/pbio/drv/usb/usb_ev3.c
@@ -883,6 +883,7 @@ static pbio_os_process_t pbdrv_usb_ev3_process;
 static pbio_error_t pbdrv_usb_ev3_process_thread(pbio_os_state_t *state, void *context) {
     static pbio_os_timer_t delay_timer;
     static pbio_os_timer_t tx_timeout_timer;
+    static pbio_os_timer_t keepalive_timer;
     static bool was_transmitting = false;
     static bool is_transmitting = false;
     static uint32_t prev_status_flags = ~0;
@@ -910,6 +911,7 @@ static pbio_error_t pbdrv_usb_ev3_process_thread(pbio_os_state_t *state, void *c
             if (usb_rx_sz && !usb_tx_response_is_not_ready) {
                 switch (ep1_rx_buf[0]) {
                     case PBIO_PYBRICKS_OUT_EP_MSG_SUBSCRIBE:
+                        pbio_os_timer_set(&keepalive_timer, 1000);
                         pbdrv_usb_is_events_subscribed = ep1_rx_buf[1];
                         ep1_tx_response_buf[0] = PBIO_PYBRICKS_IN_EP_MSG_RESPONSE;
                         pbio_set_uint32_le(&ep1_tx_response_buf[1], PBIO_PYBRICKS_ERROR_OK);
@@ -936,7 +938,8 @@ static pbio_error_t pbdrv_usb_ev3_process_thread(pbio_os_state_t *state, void *c
 
         // Send status flags if they've changed (and we can)
         new_status_flags = pbsys_status_get_flags();
-        if (pbdrv_usb_is_events_subscribed && !usb_tx_status_is_not_ready && new_status_flags != prev_status_flags) {
+        if (pbdrv_usb_is_events_subscribed && !usb_tx_status_is_not_ready &&
+            (new_status_flags != prev_status_flags || pbio_os_timer_is_expired(&keepalive_timer))) {
             ep1_tx_status_buf[0] = PBIO_PYBRICKS_IN_EP_MSG_EVENT;
             uint32_t usb_status_sz = PBIO_PYBRICKS_USB_MESSAGE_SIZE(pbsys_status_get_status_report(&ep1_tx_status_buf[1]));
 
@@ -944,6 +947,15 @@ static pbio_error_t pbdrv_usb_ev3_process_thread(pbio_os_state_t *state, void *c
             usb_setup_tx_dma_desc(CPPI_DESC_TX_STATUS, ep1_tx_status_buf, usb_status_sz);
 
             prev_status_flags = new_status_flags;
+
+            if (new_status_flags != prev_status_flags) {
+                // If we are sending a status because the flags have changed,
+                // we can bump out the keepalive timer.
+                pbio_os_timer_set(&keepalive_timer, 1000);
+            } else {
+                // Otherwise, we want to send keepalives at a particular rate.
+                pbio_os_timer_extend(&keepalive_timer);
+            }
         }
 
         // Handle timeouts


lib/pbio/drv/usb/usb_ev3.c
```diff
@@ -179,13 +184,15 @@ static const pbdrv_usb_ev3_conf_1_union_t configuration_1_desc_fs = {
...
+// global-lifetime buffer for sending dynamic data in response
+// to control transfers. This is used for the serial number string
+// and for Pybricks protocol requests.
 static union {
-    uint8_t b[STRING_DESC_MAX_SZ];
-    uint32_t u[STRING_DESC_MAX_SZ / 4];
-} usb_string_desc_buffer;
+    uint8_t b[EP0_BUF_SZ];
+    uint32_t u[EP0_BUF_SZ / 4];
+} pbdrv_usb_ep0_buffer;
```
dlech, line 196:
```suggestion
} pbdrv_usb_ev3_ep0_buffer;
```

lib/pbio/drv/usb/usb_ev3.c
```diff
@@ -179,13 +184,15 @@ static const pbdrv_usb_ev3_conf_1_union_t configuration_1_desc_fs = {
...
+// This dynamic buffer is needed in order to have an aligned,
+// global-lifetime buffer for sending dynamic data in response
+// to control transfers. This is used for the serial number string
+// and for Pybricks protocol requests.
 static union {
-    uint8_t b[STRING_DESC_MAX_SZ];
-    uint32_t u[STRING_DESC_MAX_SZ / 4];
-} usb_string_desc_buffer;
+    uint8_t b[EP0_BUF_SZ];
+    uint32_t u[EP0_BUF_SZ / 4];
```
dlech, line 195:
```suggestion
    uint32_t u[EP0_BUF_SZ / sizeof(uint32_t)];
```

lib/pbio/include/pbio/protocol.h
```diff
@@ -427,6 +427,11 @@ extern const uint8_t pbio_nus_tx_char_uuid[];
 /** USB bDeviceProtocol for Pybricks hubs */
 #define PBIO_PYBRICKS_USB_DEVICE_PROTOCOL 0xF5
 
+/** USB bRequest to retrieve GATT characteristics */
+#define PBIO_PYBRICKS_INTERFACE_READ_CHARACTERISTIC_GATT 0x01
```
dlech, line 431: `PBIO_PYBRICKS_USB_INTERFACE_READ_CHARACTERISTIC_GATT`

lib/pbio/include/pbio/protocol.h
```diff
@@ -427,6 +427,11 @@ extern const uint8_t pbio_nus_tx_char_uuid[];
...
+/** USB bRequest to retrieve Pybricks characteristics */
+#define PBIO_PYBRICKS_INTERFACE_READ_CHARACTERISTIC_PYBRICKS 0x02
```
dlech, line 433: We can make these an enum.

lib/pbio/drv/usb/usb_ev3.c
```diff
@@ -214,11 +216,15 @@ static uint32_t pbdrv_usb_setup_misc_tx_byte;
...
 
+// Whether there is a host app listening to events
+static bool pbdrv_usb_is_events_subscribed;
+
 // Buffers, used for different logical flows on the data endpoint
 static uint8_t ep1_rx_buf[PYBRICKS_EP_PKT_SZ_HS];
 static uint8_t ep1_tx_response_buf[PYBRICKS_EP_PKT_SZ_HS];
 static uint8_t ep1_tx_status_buf[PYBRICKS_EP_PKT_SZ_HS];
 static uint8_t ep1_tx_stdout_buf[PYBRICKS_EP_PKT_SZ_HS];
+#define USB_RESPONSE_ACTUAL_SZ      (1 + sizeof(uint32_t))
```
dlech, line 227: Isn't this something that should be shared by all implementations? I.e. something defined in protocol.h.

Could even make it more general and make a USB message size wrapper macro to wrap the existing macros.

```c
/**
 * Size of USB messages for Pybricks USB interface.
 *
 * USB has one extra byte header for a message type discriminator
 * compared to BLE messages.
 */
#define PBIO_PYBRICKS_USB_MESSAGE_SIZE(n) (1 + n)
```

lib/pbio/drv/usb/usb_ev3.c
```diff
@@ -875,41 +881,97 @@ static void usb_device_intr(void) {
...
 
 static pbio_error_t pbdrv_usb_ev3_process_thread(pbio_os_state_t *state, void *context) {
+    static pbio_os_timer_t delay_timer;
+    static pbio_os_timer_t tx_timeout_timer;
+    static bool was_transmitting = false;
+    static bool is_transmitting = false;
+    static uint32_t prev_status_flags = ~0;
+    static uint32_t new_status_flags = 0;
+
+    (void)ep1_tx_stdout_buf;
```
dlech: What is this for?

lib/pbio/drv/usb/usb_ev3.c
```diff
@@ -875,41 +881,97 @@ static void usb_device_intr(void) {
...
+
+            usb_tx_status_is_not_ready = true;
+            usb_setup_tx_dma_desc(CPPI_DESC_TX_STATUS, ep1_tx_status_buf, usb_status_sz);
+
+            prev_status_flags = new_status_flags;
+        }
+
+        // Handle timeouts
+        is_transmitting = usb_tx_response_is_not_ready || usb_tx_status_is_not_ready || usb_tx_stdout_is_not_ready;
+        if (was_transmitting && is_transmitting) {
```
dlech, line 957: Perhaps can use `pbio_oneshot()`?

lib/pbio/drv/usb/usb_ev3.c
```diff
@@ -875,41 +881,97 @@ static void usb_device_intr(void) {
...
+            if (pbio_os_timer_is_expired(&tx_timeout_timer)) {
```
dlech, line 958: Are we checking for timeout before setting the timer?

lib/pbio/drv/usb/usb_ev3.c
```diff
@@ -1079,22 +1077,71 @@ bool pbdrv_usb_connection_is_active(void) {
...
+    if (usb_tx_stdout_is_not_ready) {
+        return 0;
+    }
+
+    // Subtract 2 bytes for header
+    if (pbdrv_usb_is_usb_hs) {
+        return PYBRICKS_EP_PKT_SZ_HS - 2;
+    } else {
+        return PYBRICKS_EP_PKT_SZ_FS - 2;
+    }
```
dlech, line 1145:
```suggestion

     return PYBRICKS_EP_PKT_SZ_FS - 2; 
```

else is redundant

lib/pbio/drv/usb/usb_ev3.c
```diff
@@ -937,14 +939,17 @@ static pbio_error_t pbdrv_usb_ev3_process_thread(pbio_os_state_t *state, void *c
...
+            (new_status_flags != prev_status_flags || pbio_os_timer_is_expired(&keepalive_timer))) {
             ep1_tx_status_buf[0] = PBIO_PYBRICKS_IN_EP_MSG_EVENT;
             uint32_t usb_status_sz = 1 + pbsys_status_get_status_report(&ep1_tx_status_buf[1]);
 
             usb_tx_status_is_not_ready = true;
             usb_setup_tx_dma_desc(CPPI_DESC_TX_STATUS, ep1_tx_status_buf, usb_status_sz);
 
             prev_status_flags = new_status_flags;
+
+            pbio_os_timer_set(&keepalive_timer, 1000);
```
dlech, line 952: `pbio_os_timer_extend()`?

dlech (commented):
Not the most thorough review and I won't be able to test until sometime next week, but looks pretty good to me.

laurensvalk:
Amazing! 

> and possibly some fixes in pybricksdev? I was noticing some strange issues

At a glance, Pybricksdev seems to still always start slot 0 rather than start the selected slot. So code loaded to other slots does not appear to run. It does run if you start them with the button.

> With this in place (...) we might be able to call this an early alpha release?

Let's wait a while. Anyone who is able to run it in the current state is probably also able to grab the nightly release.

I'd like to turn off some of the developer hacks (the hard reset; sensor 1 disabled, repl on 0) and get the official devices enabled (except I2C).

With the rapid progress here, I think I'll put aside some of the internal BLE refactoring and return focus to EV3 sensors and UX.

lib/pbio/drv/usb/usb_ev3.c
ArcaneNibble: No, the timer gets set in the `else if` later, which this code has to pass through before both flags can become true.

laurensvalk:
Would you say we should merge it and further improve later? This has worked well for the SPI flash/adc code, so I'm happy to go this way here too.

It would be pretty useful to have #361, #363, #364 merged to facilitate other developments.

laurensvalk:
Done, thanks! Yes, let's build on this in follow up commits as needed.

PR #362, opened 2025-08-01
pbio/drv/uart/uart_ev3.c: Clear entire RX FIFO on interrupt

ArcaneNibble:
This seems to improve pasting large amounts of data into the debug REPL to the point where it will no longer lock up. It still drops bytes as buffers fill up but does not freeze.

Related to https://github.com/pybricks/support/issues/2301 cc @laurensvalk


dlech:
Can we use DMA for RX as well? We should be able to set up a cyclic DMA with a decent sized buffer so that we never lose any data.

ArcaneNibble:
I would have to think about how to get interrupts on *any* data received, and also how to update circular buffers in a race-free way...

PR #363, opened 2025-08-01
[EV3] Initial MMU enablement

ArcaneNibble:
This code enables the MMU and configures it to trap very-invalid memory accesses. It also currently marks the entire address space as uncacheable until drivers can be updated to properly flush data out of cache.

Previously, U-boot was leaving the MMU *enabled*, with page tables at the end of DRAM at 0xC3FF0000

This also fixes some invalid pointers that were discovered along the way.


lib/pbio/drv/gpio/gpio_ev3.c
```diff
@@ -32,7 +32,7 @@ static void pbdrv_gpio_alt_gpio(const pbdrv_gpio_t *gpio) {
 }
 
 static void gpio_write(const pbdrv_gpio_t *gpio, uint8_t value) {
-    if (!gpio) {
+    if (!gpio || !gpio->bank) {
```
dlech, line 35: Technically, there is a GPIO pin for `pbdrv_ioport_platform_data_vcc_pin` - it is just shared with the ADC and the USB 1.1 host port. So I would be inclined to add some compile-time options to show that EV3 is different in this regard rather than than having a non-value assigned to a GPIO.

lib/pbio/platform/ev3/platform.c
```diff
@@ -489,6 +490,17 @@ void ev3_panic_handler(int except_type, ev3_panic_ctx *except_data) {
     panic_puts("\r\nSPSR: 0x");
     panic_putu32(except_data->spsr);
 
+    uint32_t r;
+    __asm__ volatile ("mrc p15, 0, %0, c5, c0, 0" : "=r" (r));
```
dlech, line 510: These need comment to explain what the assembly code does. (Or even better, use `cp15.h` functions instead of inline assembly).

lib/pbio/platform/ev3/platform.c
```diff
@@ -633,6 +645,82 @@ static void Edma3CCErrHandlerIsr(void) {
...
+    }
+
+    // ARM local RAM, interrupt controller
+    l1_page_table[0xFFF00000 >> MMU_SECTION_SHIFT] = MMU_L1_SECTION(0xFFF00000, 0, 1, 0, 0);
+
+    // Make sure this all makes its way into memory
+    pbdrv_compiler_memory_barrier();
+
+    // Set TTBR
+    __asm__ volatile ("mcr p15, 0, %0, c2, c0, 0" "" : : "r" (l1_page_table));
```
dlech, line 744: Would be nice to use `cp15.h` instead of having inline assembly here (and add functions there if needed).

lib/pbio/platform/ev3/platform.c
```diff
@@ -422,6 +422,12 @@ static const char *const panic_types[] = {
     "Prefetch Abort",
     "Data Abort",
 };
+enum {
+    EV3_PANIC_UNKNOWN,
```
dlech, line 426: Would be nice to share these defs with `lib/pbio/platform/ev3/exceptionhandler.S` instead of redefining them. Or at least add a comment explaining where the values come from.

(also could use some blank lines before and after to keep things spaced out.)

ArcaneNibble:
Done. I added a `PBDRV_CONFIG_HAS_PORT_VCC_CONTROL` config setting and moved most of the "one-liner" CP15 operations to CP15.c. I didn't move the TLB lockdown sequence though, since that's already mostly in a separate function and isn't a "generic" operation.

PR #364, opened 2025-08-02
[EV3] Enable caching

ArcaneNibble:
This depends on the in-progress USB and MMU work, but this enables the caches on the EV3 (caching the main system RAM).

### DIFF
--- a/lib/pbio/include/pbdrv/cache.h
+++ b/lib/pbio/include/pbdrv/cache.h
@@ -24,6 +24,6 @@ void pbdrv_cache_prepare_after_dma(const void *buf, size_t sz);
 #define PBDRV_CACHE_LINE_SZ         32
 
 // Align data to a cache line, which is needed for clean RX DMA
-#define PBDRV_CACHE_LINE_ALIGNED    __attribute__((aligned(PBDRV_CACHE_LINE_SZ)))
+#define PBDRV_DMA_BUF               __attribute__((aligned(PBDRV_CACHE_LINE_SZ), section(".dma")))
 
 #endif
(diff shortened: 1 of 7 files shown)


ArcaneNibble:
fwiw this seems to improve system performance *so* much that https://github.com/pybricks/support/issues/2303 doesn't manifest anymore (at least, not in a way that is visible to the naked eye)

laurensvalk:
> fwiw this seems to improve system performance _so_ much that [pybricks/support#2303](https://github.com/pybricks/support/issues/2303) doesn't manifest anymore (at least, not in a way that is visible to the naked eye)

Can confirm visually that this is way better!

If you look carefully you can still see it with the NXT color sensor but there should be other ways to smooth it out if we want to.

laurensvalk:
The new 3 commits rebase cleanly. Let me know if you'd like me to rebase and merge this yet or refine further.

lib/pbio/include/pbdrv/cache.h
```diff
@@ -0,0 +1,29 @@
...
+// can be read by DMA peripherals. This cleans the relevant cache lines
+// and flushes the write buffer.
+void pbdrv_cache_prepare_before_dma(const void *buf, size_t sz);
+
+// Makes sure that we (the CPU) are able to read data written
+// by DMA peripherals. This invalidates the relevant cache lines.
+void pbdrv_cache_prepare_after_dma(const void *buf, size_t sz);
+
+// Accesses a variable via the uncached memory alias
+#define PBDRV_UNCACHED(x)           (*(volatile __typeof__(x) *)((uint32_t)(&(x)) + 0x10000000))
```
dlech, line 18: This seems very specific to EV3, so we should rename things accordingly. Or just put this in the tiam1808 library.

ArcaneNibble: Definitely not in tiam1808, since it's specific to the way we've specifically programmed the MMU. `PBDRV_EV3_UNCACHED`?

PR #366, opened 2025-08-05
Initial "remote coprocessor" driver

ArcaneNibble:
This was mentioned [here](https://github.com/pybricks/pybricks-micropython/pull/360#issuecomment-3134432589) and is going to be important as more functionality gets added to the PRUs.


dlech:
Very nice, thanks!

PR #368, opened 2025-08-05
[EV3] Remove SUART code which we will not use

ArcaneNibble:
Because we are already using PRU1, we cannot use dual-PRU mode for the SUART library. Remove support for this mode so that the code is easier to maintain and understand.

### DIFF
--- a/lib/pbio/drv/uart/uart_ev3_pru_lib/suart_api.c
+++ b/lib/pbio/drv/uart/uart_ev3_pru_lib/suart_api.c
@@ -477,15 +477,8 @@ int16_t pru_softuart_setbaud
     /* channel starts from 0 and uart instance starts from 1 */
     chNum = (hUart->uartNum * SUART_NUM_OF_CHANNELS_PER_SUART) - 2;
 
-    if (hUart->uartNum <= 4) {
-        /* PRU0 */
-        pruOffset = PRU_SUART_PRU0_CH0_OFFSET;
-    } else {
-        /* PRU1 */
-        pruOffset = PRU_SUART_PRU1_CH0_OFFSET;
-        /* First 8 channel corresponds to PRU0 */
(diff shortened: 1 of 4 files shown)


ArcaneNibble:
I don't know how much further we might want to specialize this interface, but the functions removed here will likely never be re-enabled.

lib/pbio/drv/uart/uart_ev3_pru_lib/suart_api.c
```diff
@@ -2319,168 +1520,83 @@ int16_t arm_to_pru_intr_init(void) {
...
-            (uint32_t)pru_arm_iomap.pru_io_addr | (PRU_INTC_CHANMAP10 &
-                0xFFFF);
-        u32value = 0x03020908;
-        s16retval =
-            pru_ram_write_data_4byte(u32offset, (uint32_t *)&u32value, 1);
-        if (-1 == s16retval) {
-            return -1;
-        }
+    /* Sets the channel for the system interrupt
+    * MAP channel 5 to SYS_EVT40	SUART3-Tx
```
dlech, line 1556:
```suggestion
    * MAP channel 5 to SYS_EVT40    SUART3-Tx
```

tabs vs. spaces

dlech: There is also some line wrapping that could be improved on lines we change, but don't spend too much time on it.

dlech (commented):
Nice to remove all of this dead code.

lib/pbio/drv/uart/uart_ev3_pru_lib/suart_api.c
ArcaneNibble: I made a reasonable effort to fix the most wonky of the line wrapping.

PR #371, opened 2025-08-06
pbio/platform/ev3/exceptionhandler.S: Fix bug in panic handler

ArcaneNibble:
The bic opcode was supposed to be used here in order to *clear* the mode bits, rather than keep them.

This bug did not surface because we've mostly been hitting panics in System mode, which has a mode bits value of all 1s.


lib/pbio/platform/ev3/exceptionhandler.S
```diff
@@ -68,7 +68,7 @@ CommonPanicHandler:
         @ NOTE: Assumes we never use User mode
```
dlech, line 68: Is this comment still true?

PR #375, opened 2025-08-06
pbio/drv/battery/battery_ev3.c: Implement EV3 battery driver

ArcaneNibble:
This reads the values from the ADC and scales them according to the ideal scaling factors.

Voltage was tested and reads in the correct order of magnitude, slightly lower (~100 mV) when compared against a cheap DMM without traceable calibration.

Current appears reasonable (116 mA) and increases when I squeeze a running motor (up to 500 mA).

Fixes https://github.com/pybricks/support/issues/2311


lib/pbio/drv/battery/battery_ev3.c
```diff
@@ -15,22 +16,40 @@
...
 
 pbio_error_t pbdrv_battery_get_voltage_now(uint16_t *value) {
-    // TODO. Calculate battery voltage based on ADC channel 4.
-    // For now, return nominal value for now so we don't trigger low battery shutdown.
-    *value = 7200;
+    pbio_error_t err;
+    uint16_t value_raw;
+    if ((err = pbdrv_adc_get_ch(PBDRV_EV3_BATTERY_VOLTAGE_ADC_CH, &value_raw)) != PBIO_SUCCESS) {
+        return err;
+    }
```
dlech, line 34:
```suggestion
    uint16_t value_raw;
    pbio_error_t err = pbdrv_adc_get_ch(PBDRV_EV3_BATTERY_VOLTAGE_ADC_CH, &value_raw);
    if (err != PBIO_SUCCESS) {
        return err;
    }
```

lib/pbio/drv/battery/battery_ev3.c
```diff
@@ -15,22 +16,40 @@
...
-    return PBIO_ERROR_NOT_IMPLEMENTED;
+    uint16_t value_raw;
+    pbio_error_t err = pbdrv_adc_get_ch(PBDRV_EV3_BATTERY_CURRENT_ADC_CH, &value_raw);
+    if (err != PBIO_SUCCESS) {
+        return err;
+    }
+    // Battery current is read across a 0.05 ohm equivalent shunt resistor
+    // which is then connected to an op-amp configured with a gain of 16
+    // (non-inverting). This yields 1 A = 0.8 V.
+    *value = ((uint32_t)(value_raw) * 5000 * 10) / (1023 * 8);
```
laurensvalk, line 51: It would be nice to move scaling to mV to the ADC driver on all platforms. Can be a different PR though. I think David might be working on ADC at the moment.

dlech:
I tried this with 3 EV3s so far and the error is kind of all over the place.

image

lib/pbio/drv/battery/battery_ev3.c
```diff
@@ -15,22 +16,40 @@
...
-    *value = 7200;
+    uint16_t value_raw;
+    pbio_error_t err = pbdrv_adc_get_ch(PBDRV_EV3_BATTERY_VOLTAGE_ADC_CH, &value_raw);
+    if (err != PBIO_SUCCESS) {
+        return err;
+    }
+    // Battery voltage is read through a voltage divider consisting of two
+    // 100 K resistors, which halves the voltage. The ADC returns 10 LSBs,
+    // where full scale is 5 V.
+    *value = ((uint32_t)(value_raw) * 2 * 5000) / 1023;
```
dlech, line 38: In the Linux kernel, for the battery voltage, I also took into account the V<sub>CE</sub> of Q19B that I somehow figured to be 50 mV (don't remember how I came up with that number). And I also added a bit more to account for the shunt resistor which depends on the current.

laurensvalk:
@dlech , what is the status on this one? Would you like to make the ADC fixes first before we merge this?

dlech:
Last weekend, I did some tests on the current measurement using this PR + changing SCLK rate to 1 MHz. The results show that it is not very accurate.

Methodology:

```python
from pybricks.ev3devices import Motor
from pybricks.hubs import EV3Brick
from pybricks.parameters import Port
from pybricks.tools import multitask, run_task, wait

hub = EV3Brick()

motors = [Motor(p) for p in (Port.A, Port.B, Port.C, Port.D)]

for m in motors:
    m.dc(20)

avg_current = hub.battery.current()
RATIO = 0.98


async def average():
    global avg_current

    while True:
        await wait(10)
        avg_current = int(RATIO * avg_current + (1 - RATIO) * hub.battery.current())


async def print_info():
    while True:
        print(hub.battery.voltage(), "mv", avg_current, "mA")
        await wait(1000)


run_task(multitask(average(), print_info()))
```

Used this program to compare current measured with a digital multi-meter (DMM) compared to the on-board ADC. The duty cycle was adjusted manually to get data points for various current usages. The X axis of the graphs is duty cycle percent. The Y axis is milliamps. This was repeated with 3 different EV3 bricks.

image

It can't be seen in the graph, but at higher currents, the ADC signal for the current measurement is quite noisy even with the averaging filter applied.

There is some clear non-linearity in the ADC measurement, but it doesn't seem to be consistent between EV3 bricks. So I'm not really sure what we could do to make this better, if anything.

lib/pbio/drv/battery/battery_ev3.c
dlech: After lowering the SCLK rate to 1MHz, this 50 mV offset seems to go away, so I might have figured that wrong in the Linux driver.

As far as the shunt resistor goes, it would probably be better to include that in the low battery shutdown battery level, but not in the battery voltage that we use for motor control. So for now, I guess we should just leave that out.

dlech:
I rebased and added a couple of extra commits. One to slow down the ADC SPI SCLK to get more accurate measurements. (I suspect this may help with the long standing issue of large motors being detected as medium motors). Another is some more battery monitoring code we had from ev3dev and is also in the official LEGO firmware for a high temperature shutdown.

PR #377, opened 2025-08-09
pybricks.tools: Fix crash when cross() args are wrong type.

dlech:
Add type checking to fix a crash if the arguments to `cross()` are not `Matrix` objects.


laurensvalk:
Thanks!

PR #378, opened 2025-08-16
Line drawing, tests and circles improvement

schodet:
Hi,

here are some improvements for the Image code:

- line and thick lines drawing,
- unit tests,
- small improvement on the circle shape.


lib/pbio/src/image/image.c
```diff
@@ -262,6 +262,142 @@ void pbio_image_draw_vline(pbio_image_t *image, int x, int y, int l,
     }
 }
 
+/**
+ * Draw a line with a flat slope (less or equal to 1).
+ * @param [in,out] image      Image to draw into.
```
dlech, line 267:
```suggestion
 * @param [in] image      Image to draw into.
```

`[in,out]` would require `pbio_image_t **image` and would return a different pointer than the one that was passed in.

lib/pbio/src/image/image.c
```diff
@@ -262,6 +262,142 @@ void pbio_image_draw_vline(pbio_image_t *image, int x, int y, int l,
...
+            err -= dx * 2;
+            y += ydir;
+        }
+        x++;
+    } while (x <= x2);
+}
+
+/**
+ * Draw a line with a steep slope (more than 1).
+ * @param [in,out] image      Image to draw into.
```
dlech, line 335:
```suggestion
 * @param [in] image      Image to draw into.
```

And everywhere else too.

lib/pbio/test/src/test_image.c
```diff
@@ -0,0 +1,835 @@
...
+        "MP stamp ST.......mp"
+        "P stamp ST.......mp "
+        " stamp ST.......mp S"
+        "stamp ST.......mp ST"
+        "tamp ST.......mp STA"
+        "amp ST.......mp STAM"
+        "mp ST.......mp STAMP");
+
+    test_image_prepare_images(&outer, &inner, &full);
+    srand(1234);
```
dlech, line 224: One thing I've done before to make tests random but also reproducible is make the seed number come from an environment variable. Then have some global init function that runs once that checks the environment variable. If it is not set, generate a random seed and print it. This way, in CI runs, we can see the seed that was used and can use the same seed locally to reproduce a failed test if needed. And it gets us a bit more coverage by actually using different numbers on each CI run.

In once case, I think I used the github CI variable for the CI run number as the seed so I didn't even have to make the global init function.

dlech (commented):
Very nice. I like the way you are doing the tests. It makes it very easy to visualize what is going on.

lib/pbio/src/image/image.c
schodet: I do not agree, the image is read and written, so it’s both in and out.

Random people on the internet seems to agree: https://stackoverflow.com/questions/47732665/is-that-an-in-or-in-out-parameter-doxygen-c :slightly_smiling_face: 

BTW, I find those in/out redundant for C, it’s usually clear from the function prototype.

lib/pbio/test/src/test_image.c
schodet: I’m sorry, I do not agree too :-) what is the point of adding non-determinism?

If I run a CI for a given commit, I want to have the exact same result one month later with the same commit, it should not depend on external data.

About code coverage, you can check that every code path is tested. Adding non-determinism would mean that possibly a CI run would report a buggy reduced coverage for a unrelated commit.

If the global init function runs once, it means that when I select only a part of the test to run (using `./build/test-pbio src/image/..` for example), it gets different numbers. So it should at least be done at the start of each test.

Previous tests use rand without using srand so the random number generator was not initialized.

I have made the same kind of system (env variable and print) in the past, but I think it brings complexity for no gain.

Did I convince you?

lib/pbio/src/image/image.c
dlech: Sure, there is room for interpretation. I learned C# before C, so I've always used the C# semantics for in/out, which are well defined. 

> BTW, I find those in/out redundant for C, it’s usually clear from the function prototype.

This is exactly why I don't consider `[in,out]` correct here. It is only useful for something like:

```c
int read(void* data, size_t *size);
```

Where the precondition is that `*size` is set to the size of `data` in bytes and the post-condition is that `*size` is set to the number of bytes actually read, which may be less than the input size. Just going by the signature alone, one would normally expect `size_t *size` to be an out parameter so `[in,out]` makes it a bit more obvious that something unusual is going on.

Anyway, I'm not going to make a big deal about it. I don't think we've used this 100% consistently throughout the code anyway. But, in the majority of cases, we have only used `[in]` for cases where a struct is modified by a function. So it would be more consistent to stick with just `[in]`.

dlech: > Sure, there is room for interpretation. I learned C# before C, so I've always used the C# semantics for in/out, which are well defined.

I guess for C# structs though, we would need to call it `[in,out]` like you are doing here since otherwise C# would pass the struct by value. :smile: It's just that normally something like the image type would be an object in C# rather than a struct.

lib/pbio/test/src/test_image.c
schodet: I’m not sure you would be happy to have 460000 random numbers in the source code.

If the number of loop is large enough, using a different seed in different runs brings no benefit, but brings the drawbacks I mentioned.

If I increase 10000 to 1000000, it’s the equivalent of running 100 CI, but it takes 2 seconds.

My feeling was that 10000 loops was enough to find those edge cases (and it found some edge cases that I fixed), and also, it generates nice pictures, but I can not actually prove it.

I do not want to impose anything, but I should at least try to convince you, if not, I can adapt. I hope you do not feel bad about it.

dlech: I am convinced. (And I don't feel bad at all. :smile:)

lib/pbio/src/image/image.c
schodet: OK, made the change for the whole file.

lib/pbio/src/image/image.c
schodet: Done.

lib/pbio/test/src/test_image.c
schodet: Written form + non native language, it’s sometime easy to sound rude without having the intention. I’m happy this is not the case.

Thanks :)

dlech:
Thanks!

PR #381, opened 2025-09-08
pbio/sys/light: stop animation when drawing on screen

schodet:
Stop the "running" animation on the first drawing to the screen, and clear screen when the user program exits.

Refs: https://github.com/pybricks/support/issues/2154

First proposal to get the ball rolling.

### DIFF
--- a/lib/pbio/include/pbsys/light.h
+++ b/lib/pbio/include/pbsys/light.h
@@ -17,8 +17,18 @@ extern pbio_color_light_t *pbsys_status_light_main;
 #endif
 
 #if PBSYS_CONFIG_HUB_LIGHT_MATRIX
+
 #include <pbio/light_matrix.h>
+
 extern pbio_light_matrix_t *pbsys_hub_light_matrix;
+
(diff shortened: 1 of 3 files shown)


schodet:
I had the same feeling about `pbio/sys` not being the right place.

The idea is to have the animation and screen control at the same place that the user program start/stop handling.

Depending on your refactoring work landing time, would you mind merging anyway so it solves my problem until this is moved elsewhere?

laurensvalk:
> Would you mind merging anyway so it solves my problem until this is moved elsewhere?

Yes, we can do that :smile:

laurensvalk:
And simplifying further... Considering that we only have the light matrix display so that we can have a UI for now, we can avoid this problem entirely by just not starting the animation in the first place and clearing the display when the program starts and ends. I've added a commit to #382.

Tested by drawing on the screen from the user program. It is erased when the program ends, and a clean screen appears when the program begins. If that works well enough for you, we could close this PR.

PR #386, opened 2025-09-13
USB temporary fix

dlech:
I've been experimenting with Prime hub USB on Windows and the transmit time is just timing out all of the time causing the hub to unsubscribe from USB notifications, effectively losing communication with the hub.

I don't know what we could do in it's place yet to handle the case when a user program is blocked on printing and and application disappears without sending the unsubscribe message. But the timer doesn't seem like a viable solution. Maybe we just leave it like this for a while and see how often this really happens? If it does happen, the user just has to unplug the USB cable (and optionally plug it back in) to get unstuck.

On EV3, removing the timer wasn't so trivial, so I haven't done that yet. Maybe something @ArcaneNibble could help with?


ArcaneNibble:
Thinking about this further, we might still want a timeout (block the user program for some interval before giving up)

The critical difference between using an endpoint halt vs another mechanism is that it signals *to the host* that the device thinks the "is subscribe" state changed

PR #387, opened 2025-09-16
pbio/drv/bluetooth_stm32_cc2640: Use loops to set params.

laurensvalk:
This is an independent spin-off from other Bluetooth progress. This has not yet been tested. 

---------------

When inspecting the map.elf, the init_task was well over 4K. Since much of it is repetition with different parameters, we can vastly reduce the number of different entry and yield points by looping over a list of parameters instead.

This reduces the firmware size by nearly 2K, which is helpful since we are running out of space on Technic Hub. :tada: 

Almost everything is still set in the same order so should work the same if we didn't copy things incorrectly. Only the bond manager configuration is reordered slightly so we can use a single loop.


dlech:
> This reduces the firmware size by nearly 2K, which is helpful since we are running out of space on Technic Hub.

Holy object code, Batman! Nice find!

PR #390, opened 2025-09-29
Split HMI to functional variants

laurensvalk:
**HMI**
The Human Machine Interface (HMI) is split into functional variants. This way we can make a dedicated UI for platforms with an LCD and buttons such as the EV3 and NXT. They have a dedicated init and deinit which we can use for boot and shutdown animations, possibly including sound.

**Poweroff**
This generalizes the power down button so it doesn't have to be the center button. The EV3 will use the back button to stop programs. The button can be held for power down. We may additionally add an explicit shutdown option to the UI in the future.

**Simplified light_matrix system API**
There were 3 different light_matrix.h and light_matrix.c that were doing slightly different things. We really have only one light matrix. It is used by the HMI and by user programs. Simplifying this makes it much more transparent to build the HMI.

**Disable EV3 hard reset by default**
Since the back button is now used for normal poweroff, the hard reset is disabled by default. Developers can still turn it on if needed.


lib/pbio/platform/ev3/pbsysconfig.h
```diff
@@ -1,17 +1,18 @@
...
+// Copyright (c) 2020-2025 The Pybricks Authors
 
 #define PBSYS_CONFIG_FEATURE_BUILTIN_USER_PROGRAM_REPL             (1)
 #define PBSYS_CONFIG_FEATURE_BUILTIN_USER_PROGRAM_PORT_VIEW        (0)
 #define PBSYS_CONFIG_FEATURE_BUILTIN_USER_PROGRAM_IMU_CALIBRATION  (0)
 #define PBSYS_CONFIG_FEATURE_PROGRAM_FORMAT_MULTI_MPY_V6           (1)
 #define PBSYS_CONFIG_FEATURE_PROGRAM_FORMAT_MULTI_MPY_V6_3_NATIVE  (0)
 #define PBSYS_CONFIG_BATTERY_TEMP_ESTIMATION        (1)
+#define PBSYS_CONFIG_HMI                            (1)
+#define PBSYS_CONFIG_HMI_STOP_BUTTON                (1 << 7) // top left, back button
```
dlech, line 11: Why not using macro names instead of hard-coding numbers?

lib/pbio/src/os.c
```diff
@@ -69,50 +68,42 @@ pbio_error_t pbio_port_process_none_thread(pbio_os_state_t *state, void *context
...
+    pbio_os_process_t **pp = &process_list;
 
-    // Add the new process to the end of the list.
-    pbio_os_process_t *last = process_list;
-    if (!last) {
-        process_list = process;
-    } else {
-        while (last->next) {
-            last = last->next;
+    while (*pp) {
```
dlech, line 77: Looks like a for loop. 😉

lib/pbio/platform/ev3/pbsysconfig.h
laurensvalk: Including `pbio` which includes `pbdrv` here seems to mess things up, leading them to be always zero.

lib/pbio/src/os.c
laurensvalk: What do you mean?

PR #392, opened 2025-09-30
pbio/platform/ev3: Enable more storage and heap.

laurensvalk:
Opening this for discussion. @dlech @ArcaneNibble. Anything we should be aware of? Should we just max it out instead of picking these small values (though still far larger than SPIKE Prime).

I suppose garbage collection will happen later but take more time this way. We've previously added hooks to upstream MicroPython to drive our event loop during garbage collection. But user code will block during collection.


dlech:
Hmm... looks like we had 8MiB on Pybricks v2.0. https://github.com/pybricks/micropython/commit/ecbd58aae37d290cccf93eb0dda3b2a622db8fa4

So maybe the performance hit wasn't so noticeable with the faster processor? We could try it and try to measure how long a GC cycle takes. Just gut feeling, I think a 1MiB size might be more reasonable.

PR #393, opened 2025-10-06
pbio/drv/button_gpio: Wait for release and delay on power down.

laurensvalk:
Attempted fix for https://github.com/pybricks/support/issues/2176, but cannot reproduce locally.


dlech:
I don't understand what this actually does and there is no explanation in the commit message.

laurensvalk:
Hmm, this wasn't supposed to be merged yet.

laurensvalk:
But what it does is:
- Use `pbdrv_button_deinit()` to request the gpio debounce process to gracefully stop.
- It waits for release, it waits another `200 ms` before considering deinit complete, so as to be sure it is really released before we proceed with the rest of poweroff.

PR #394, opened 2025-10-07
pbio/light_animation: Convert to pbio/os.

laurensvalk:
This one is slightly less trivial to convert from Contiki, so could use a closer look to be safe.

### DIFF
--- a/lib/pbio/test/src/test_animation.c
+++ b/lib/pbio/test/src/test_animation.c
@@ -15,8 +15,6 @@
 
 #define TEST_ANIMATION_TIME 10
 
-PROCESS_NAME(pbio_light_animation_process);
-
 static uint8_t test_animation_set_hsv_call_count;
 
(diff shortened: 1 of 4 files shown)


lib/pbio/test/src/test_animation.c
```diff
@@ -15,45 +15,49 @@
...
 static uint8_t test_animation_set_hsv_call_count;
 
 static uint32_t test_animation_next(pbio_light_animation_t *animation) {
     test_animation_set_hsv_call_count++;
     return TEST_ANIMATION_TIME;
 }
 
-static PT_THREAD(test_light_animation(struct pt *pt)) {
-    PT_BEGIN(pt);
+#define YIELD(state)      \
```
laurensvalk, line 25: We've only needed it in this test so far, but we should probably move this to the `pbio/os` module.

PR #457, opened 2026-01-25
USB remove timeout

dlech:
This is just for testing for now. This is probably the first step towards unbreaking USB on slow computers.

### DIFF
--- a/lib/pbio/drv/usb/usb.c
+++ b/lib/pbio/drv/usb/usb.c
@@ -294,14 +294,14 @@ static pbio_error_t pbdrv_usb_process_thread(pbio_os_state_t *state, void *conte
                 pbdrv_usb_respond_soon = false;
 
                 // Send the response.
-                PBIO_OS_AWAIT(state, &sub, err = pbdrv_usb_tx_response(&sub, error_code));
+                PBIO_OS_AWAIT(state, &sub, err = pbdrv_usb_tx_response(&sub, error_code, pbdrv_usb_process.request == PBIO_OS_PROCESS_REQUEST_TYPE_CANCEL));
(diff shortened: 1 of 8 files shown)


laurensvalk:
One reason this is failing sometimes even while Pybricks Code is in focus, is that nominal messages may take close to 50 ms. 

Added this UART debug line to `pbdrv_usb_tx_event`:

```
USB event transmitted in 0 ms
USB event transmitted in 0 ms
USB event transmitted in 0 ms
USB event transmitted in 1 ms
USB event transmitted in 30 ms
USB event transmitted in 49 ms
USB event transmitted in 49 ms
USB event transmitted in 49 ms
USB event transmitted in 51 ms   <----- would have triggered a reset but nothing is wrong
USB event transmitted in 44 ms
USB event transmitted in 27 ms
USB event transmitted in 32 ms
USB event transmitted in 25 ms
USB event transmitted in 27 ms
USB event transmitted in 21 ms
USB event transmitted in 17 ms
USB event transmitted in 0 ms
USB event transmitted in 0 ms
USB event transmitted in 0 ms
USB event transmitted in 0 ms
USB event transmitted in 0 ms
USB event transmitted in 0 ms
USB event transmitted in 0 ms
```

So at the very least we can use a timeout more representative of something actually being wrong, like `500` ms.

laurensvalk:
> So at the very least we can use a timeout more representative of something actually being wrong, like `500` ms.

In fact, the issues caused by going to another tab also appear resolved this way. When printing in a tight loop, 99% of messages take ~0ms to send, but occasionally it will be slightly above 50ms.

dlech:
I guess this isn't needed anymore since we changed to serial interface.

PR #465, opened 2026-03-25
Rebuild generated files when media files change

schodet:
Several time, my build failed because of changes in media files when switching branch. When used in an frozen python script, the build does not fail, but it fails at run time.

This fixes it.

### DIFF
--- a/bricks/_common/common.mk
+++ b/bricks/_common/common.mk
@@ -552,9 +552,31 @@ ifneq ($(PB_MCU_FAMILY),TIAM1808)
 SRC_S += lib/pbio/platform/$(PBIO_PLATFORM)/startup.s
 endif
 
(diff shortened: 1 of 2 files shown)


dlech (commented):
I guess this works, but it would seem more natural to me to list the output files in the Makefile rather than the input files. And have a recipe that converts individual files instead of all at once. Then make should be able to do the right thing without any special handling.

schodet:
Hi!

I pushed a new version addressing your comments. The list of media files is now given explicitly.

Note that if you want to quickly iterate on a change, you still can temporarily patch the list with a `$(wildcard blah/blah/*)`.

schodet:
Hi, I just rebased the branch.

laurensvalk:
Thanks!

PR #468, opened 2026-04-03
added keyword argument to UARTDevice to set power on one of the power…

### LINKED ISSUE #2655: [Feature] Allow to set powered pin for UARTDevice.
## Problem description
The UARTDevice iodevice is a nice generic way to communicate with external devices using plain uart communciation. When external devices have more power needs than the 3v3 line can deliver (e.g. when driving NeoPixels or Servo motors), it would be nice when such a device can use 8V power coming from one of the power lines P1 or P2.

The PUPDevice iodevice allows for setting power on either P1 or P2 depending how that is negotiatied in the PUP protocol. For I2CDevices, there is a keyword argument `powered` that allows for powering P1 (not P2). Unfortunately, I2Cdevice is not present for prime hub or technic hub, only for EV3 hub (and there the `powered` does work, but is not effective, as the P1 pin is connected through a 330Ohm resistor, and the voltage drops sharply when connecting a device that draws some current.)

## Related issue

[Feature: UARTdevice support for PoweredUp hubs](https://github.com/pybricks/support/issues/220) 

## Proposed solution 
I propose to add a keyword argument `power_pin` to the UARTDevice  `__init__` method where the argument can be 0 (no power), 1 (P1 powered) or 2 (P2 powered).

In a pybricks program that would look like:
```
uart = UARTDevice(Port.A, power_pin = 1)
```

resulting in P1 powered and

```
uart = UARTDevice(Port.A, power_pin = 0)
```
resulting in P1 nor P2 powered.

## Proposed changes to `pb_type_uart_device.c` 

- an extra keyword argument `power_pin` is added to `__init__`

```
static mp_obj_t pb_type_uart_device_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *args) {
    PB_PARSE_ARGS_CLASS(n_args, n_kw, args,
        PB_ARG_REQUIRED(port),
        PB_ARG_DEFAULT_INT(baudrate, 115200),
        PB_ARG_DEFAULT_NONE(timeout),
        PB_ARG_DEFAULT_INT(power_pin, 0)
        );
```
- depending on the value of `power_pin` the corresponding pin is powered, or none, when not 1 or 2.
```
    if (mp_obj_get_int(power_pin_in) == 1) {
       pbio_port_p1p2_set_power(self->port, PBIO_PORT_POWER_REQUIREMENTS_BATTERY_VOLTAGE_P1_POS);
    } else if (mp_obj_get_int(power_pin_in) == 2) {
       pbio_port_p1p2_set_power(self->port, PBIO_PORT_POWER_REQUIREMENTS_BATTERY_VOLTAGE_P2_POS);
    } else
       pbio_port_p1p2_set_power(self->port, PBIO_PORT_POWER_REQUIREMENTS_NONE);
```

<details>
<summary> Click triangle to see the full code with changes for pb_type_uart_device.c </summary>

```
/pybricks-micropython/pybricks/iodevices/pb_type_uart_device.c
```

```
// SPDX-License-Identifier: MIT
// Copyright (c) 2018-2025 The Pybricks Authors

#include "py/mpconfig.h"

#if PYBRICKS_PY_IODEVICES

#include "py/mphal.h"
#include "py/objstr.h"
#include "py/runtime.h"

#include <pbdrv/uart.h>
#include <pbio/port_interface.h>

#include <pybricks/common.h>
#include <pybricks/parameters.h>
#include <pybricks/tools/pb_type_async.h>

#include <pybricks/util_mp/pb_kwarg_helper.h>
#include <pybricks/util_mp/pb_obj_helper.h>
#include <pybricks/util_pb/pb_error.h>

// pybricks.iodevices.uart_device class object
typedef struct _pb_type_uart_device_obj_t {
    mp_obj_base_t base;
    pbio_port_t *port;
    pbdrv_uart_dev_t *uart_dev;
    uint32_t timeout;
    pb_type_async_t *write_iter;
    mp_obj_t write_obj;
    pb_type_async_t *read_iter;
    mp_obj_str_t *read_obj;
    const byte *wait_data;
    size_t wait_len;
} pb_type_uart_device_obj_t;

// pybricks.iodevices.UARTDevice.set_baudrate
static mp_obj_t pb_type_uart_device_set_baudrate(mp_obj_t self_in, mp_obj_t baudrate_in) {
    pb_type_uart_device_obj_t *self = MP_OBJ_TO_PTR(self_in);

    int32_t baud_rate = pb_obj_get_int(baudrate_in);
    if (baud_rate < 1) {
        pb_assert(PBIO_ERROR_INVALID_ARG);
    }
    pbdrv_uart_set_baud_rate(self->uart_dev, baud_rate);
    return mp_const_none;
}
static MP_DEFINE_CONST_FUN_OBJ_2(pb_type_uart_device_set_baudrate_obj, pb_type_uart_device_set_baudrate);


// pybricks.iodevices.UARTDevice.__init__
static mp_obj_t pb_type_uart_device_make_new(const mp_obj_type_t *type, size_t n_args, size_t n_kw, const mp_obj_t *args) {
    PB_PARSE_ARGS_CLASS(n_args, n_kw, args,
        PB_ARG_REQUIRED(port),
        PB_ARG_DEFAULT_INT(baudrate, 115200),
        PB_ARG_DEFAULT_NONE(timeout),
        PB_ARG_DEFAULT_INT(power_pin, 0)
        );
    // Get device, which inits UART port
    pb_type_uart_device_obj_t *self = mp_obj_malloc(pb_type_uart_device_obj_t, type);

    if (timeout_in == mp_const_none) {
        // In the uart driver implementation, 0 means no timeout.
        self->timeout = 0;
    } else {
        // Timeout of 0 is often perceived as partial read if the requested
        // number of bytes is not available. This is not supported, so don't
        // make it appear that way.
        if (pb_obj_get_int(timeout_in) < 1) {
            pb_assert(PBIO_ERROR_INVALID_ARG);
        }
        self->timeout = pb_obj_get_int(timeout_in);
    }

    pbio_port_id_t port_id = pb_type_enum_get_value(port_in, &pb_enum_type_Port);
    pb_assert(pbio_port_get_port(port_id, &self->port));
    pbio_port_set_mode(self->port, PBIO_PORT_MODE_UART);

    if (mp_obj_get_int(power_pin_in) == 1) {
       pbio_port_p1p2_set_power(self->port, PBIO_PORT_POWER_REQUIREMENTS_BATTERY_VOLTAGE_P1_POS);
    } else if (mp_obj_get_int(power_pin_in) == 2) {
       pbio_port_p1p2_set_power(self->port, PBIO_PORT_POWER_REQUIREMENTS_BATTERY_VOLTAGE_P2_POS);
    } else
       pbio_port_p1p2_set_power(self->port, PBIO_PORT_POWER_REQUIREMENTS_NONE);

    pb_assert(pbio_port_get_uart_dev(self->port, &self->uart_dev));
    pb_type_uart_device_set_baudrate(MP_OBJ_FROM_PTR(self), baudrate_in);
    pbdrv_uart_flush(self->uart_dev);

    // Awaitables associated with reading and writing.
    self->write_iter = NULL;
    self->read_iter = NULL;
    self->wait_len = 0;

    return MP_OBJ_FROM_PTR(self);
}

static pbio_error_t pb_type_uart_device_write_iter_once(pbio_os_state_t *state, mp_obj_t self_in) {
    pb_type_uart_device_obj_t *self = MP_OBJ_TO_PTR(self_in);
    GET_STR_DATA_LEN(self->write_obj, data, data_len);
    return pbdrv_uart_write(state, self->uart_dev, (uint8_t *)data, data_len, self->timeout);
}

static mp_obj_t pb_type_uart_device_write_return_map(mp_obj_t self_in) {
    pb_type_uart_device_obj_t *self = MP_OBJ_TO_PTR(self_in);
    // Write always returns none, but this is effectively a completion callback.
    // So we can use it to disconnect the write object so it can be garbage collected.
    self->write_obj = MP_OBJ_NULL;
    return mp_const_none;
}

// pybricks.iodevices.UARTDevice.write
static mp_obj_t pb_type_uart_device_write(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {

    PB_PARSE_ARGS_METHOD(n_args, pos_args, kw_args,
        pb_type_uart_device_obj_t, self,
        PB_ARG_REQUIRED(data));

    // Assert that data argument are bytes
    if (!(mp_obj_is_str_or_bytes(data_in) || mp_obj_is_type(data_in, &mp_type_bytearray))) {
        pb_assert(PBIO_ERROR_INVALID_ARG);
    }

    // Prevents this object from being garbage collected while the write is in progress.
    self->write_obj = data_in;

    pb_type_async_t config = {
        .iter_once = pb_type_uart_device_write_iter_once,
        .parent_obj = MP_OBJ_FROM_PTR(self),
        .return_map = pb_type_uart_device_write_return_map,
    };
    return pb_type_async_wait_or_await(&config, &self->write_iter, true);
}
static MP_DEFINE_CONST_FUN_OBJ_KW(pb_type_uart_device_write_obj, 1, pb_type_uart_device_write);

// pybricks.iodevices.UARTDevice.waiting
static mp_obj_t pb_type_uart_device_waiting(mp_obj_t self_in) {
    pb_type_uart_device_obj_t *self = MP_OBJ_TO_PTR(self_in);
    return mp_obj_new_int(pbdrv_uart_in_waiting(self->uart_dev));
}
static MP_DEFINE_CONST_FUN_OBJ_1(pb_type_uart_device_waiting_obj, pb_type_uart_device_waiting);

static pbio_error_t pb_type_uart_device_read_iter_once(pbio_os_state_t *state, mp_obj_t self_in) {
    pb_type_uart_device_obj_t *self = MP_OBJ_TO_PTR(self_in);
    return pbdrv_uart_read(state, self->uart_dev, (uint8_t *)self->read_obj->data, self->read_obj->len, self->timeout);
}

static mp_obj_t pb_type_uart_device_read_return_map(mp_obj_t self_in) {
    pb_type_uart_device_obj_t *self = MP_OBJ_TO_PTR(self_in);
    mp_obj_str_t *result = self->read_obj;
    self->read_obj = NULL;
    return pb_obj_new_bytes_finish(result);
}

// pybricks.iodevices.UARTDevice.read
static mp_obj_t pb_type_uart_device_read(size_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {

    PB_PARSE_ARGS_METHOD(n_args, pos_args, kw_args,
        pb_type_uart_device_obj_t, self,
        PB_ARG_DEFAULT_INT(length, 1));

    // Allocate new buffer that we'll read into.
    self->read_obj = pb_obj_new_bytes_prepare(pb_obj_get_positive_int(length_in));

    pb_type_async_t config = {
        .iter_once = pb_type_uart_device_read_iter_once,
        .parent_obj = MP_OBJ_FROM_PTR(self),
        .return_map = pb_type_uart_device_read_return_map,
    };
    return pb_type_async_wait_or_await(&config, &self->read_iter, true);
}
static MP_DEFINE_CONST_FUN_OBJ_KW(pb_type_uart_device_read_obj, 1, pb_type_uart_device_read);

// pybricks.iodevices.UARTDevice.read_all
static mp_obj_t pb_type_uart_device_read_all(mp_obj_t self_in) {

    pb_type_uart_device_obj_t *self = MP_OBJ_TO_PTR(self_in);
    uint32_t in_waiting = pbdrv_uart_in_waiting(self->uart_dev);

    if (in_waiting == 0) {
        return mp_const_empty_bytes;
    }

    mp_obj_str_t *result = pb_obj_new_bytes_prepare(in_waiting);

    // We know we can read this in one go, so all data will be copied without
    // intermediate yields.
    pbio_os_state_t state = 0;
    pb_assert(pbdrv_uart_read(&state, self->uart_dev, (uint8_t *)result->data, in_waiting, 0));

    return pb_obj_new_bytes_finish(result);
}
static MP_DEFINE_CONST_FUN_OBJ_1(pb_type_uart_device_read_all_obj, pb_type_uart_device_read_all);

// pybricks.iodevices.UARTDevice.clear
static mp_obj_t pb_type_uart_device_clear(mp_obj_t self_in) {
    pb_type_uart_device_obj_t *self = MP_OBJ_TO_PTR(self_in);
    pbdrv_uart_flush(self->uart_dev);
    return mp_const_none;
}
static MP_DEFINE_CONST_FUN_OBJ_1(pb_type_uart_device_clear_obj, pb_type_uart_device_clear);

static pbio_error_t pb_type_uart_device_wait_until_iter_once(pbio_os_state_t *state, mp_obj_t self_in) {

    pb_type_uart_device_obj_t *self = MP_OBJ_TO_PTR(self_in);

retry:

    // Yield if not enough to read yet.
    if (pbdrv_uart_in_waiting(self->uart_dev) < self->wait_len) {
        return PBIO_ERROR_AGAIN;
    }

    // We can read the full amount of bytes without blocking now.
    for (size_t i = 0; i < self->wait_len; i++) {
        // Read at most one byte since the viewing window may not overlap pattern.
        pbio_os_state_t sub = 0;
        uint8_t rx;
        pb_assert(pbdrv_uart_read(&sub, self->uart_dev, &rx, 1, 0));

        if (rx != self->wait_data[i]) {
            // Not the character we expected, so start over, yielding if there
            // is not enough to read.
            goto retry;
        }
    }
    return PBIO_SUCCESS;
}

static mp_obj_t pb_type_uart_device_wait_until_return_map(mp_obj_t self_in) {
    pb_type_uart_device_obj_t *self = MP_OBJ_TO_PTR(self_in);
    self->wait_len = 0;
    self->wait_data = NULL;
    return mp_const_none;
}

static mp_obj_t pb_type_uart_device_wait_until(mp_obj_t self_in, mp_obj_t pattern_in) {

    pb_type_uart_device_obj_t *self = MP_OBJ_TO_PTR(self_in);

    if (self->wait_len) {
        pb_assert(PBIO_ERROR_BUSY);
    }

    self->wait_data = (const uint8_t *)mp_obj_str_get_data(pattern_in, &self->wait_len);
    if (self->wait_len == 0) {
        pb_assert(PBIO_ERROR_INVALID_ARG);
    }

    pb_type_async_t config = {
        .iter_once = pb_type_uart_device_wait_until_iter_once,
        .parent_obj = MP_OBJ_FROM_PTR(self),
        .return_map = pb_type_uart_device_wait_until_return_map,
    };
    return pb_type_async_wait_or_await(&config, &self->read_iter, true);
}
static MP_DEFINE_CONST_FUN_OBJ_2(pb_type_uart_device_wait_until_obj, pb_type_uart_device_wait_until);

// dir(pybricks.iodevices.uart_device)
static const mp_rom_map_elem_t pb_type_uart_device_locals_dict_table[] = {
    { MP_ROM_QSTR(MP_QSTR_read),         MP_ROM_PTR(&pb_type_uart_device_read_obj)         },
    { MP_ROM_QSTR(MP_QSTR_read_all),     MP_ROM_PTR(&pb_type_uart_device_read_all_obj)     },
    { MP_ROM_QSTR(MP_QSTR_write),        MP_ROM_PTR(&pb_type_uart_device_write_obj)        },
    { MP_ROM_QSTR(MP_QSTR_waiting),      MP_ROM_PTR(&pb_type_uart_device_waiting_obj)      },
    { MP_ROM_QSTR(MP_QSTR_wait_until),   MP_ROM_PTR(&pb_type_uart_device_wait_until_obj)   },
    { MP_ROM_QSTR(MP_QSTR_set_baudrate), MP_ROM_PTR(&pb_type_uart_device_set_baudrate_obj) },
    { MP_ROM_QSTR(MP_QSTR_clear),        MP_ROM_PTR(&pb_type_uart_device_clear_obj)        },
};
static MP_DEFINE_CONST_DICT(pb_type_uart_device_locals_dict, pb_type_uart_device_locals_dict_table);

// type(pybricks.iodevices.uart_device)
MP_DEFINE_CONST_OBJ_TYPE(pb_type_uart_device,
    MP_QSTR_uart_device,
    MP_TYPE_FLAG_NONE,
    make_new, pb_type_uart_device_make_new,
    locals_dict, &pb_type_uart_device_locals_dict);

#endif // PYBRICKS_PY_IODEVICES

```

</details>

ste7anste7an:
The UARTDevice iodevice is a nice generic way to communicate with external devices using plain uart communciation. When external devices have more power needs than the 3v3 line can deliver (e.g. when driving NeoPixels or Servo motors), it would be nice when such a device can use 8V power coming from one of the power lines P1 or P2.

The PUPDevice iodevice allows for setting power on either P1 or P2 depending how that is negotiatied in the PUP protocol. For I2CDevices, there is a keyword argument powered that allows for powering P1 (not P2). Unfortunately, I2Cdevice is not present for prime hub or technic hub, only for EV3 hub (and there the powered does work, but is not effective, as the P1 pin is connected through a 330Ohm resistor, and the voltage drops sharply when connecting a device that draws some current.)

# Proposed solution
I propose to add a keyword argument power_pin to the UARTDevice __init__ method where the argument can be 0 (no power), 1 (P1 powered) or 2 (P2 powered).

In a pybricks program that would look like:

```
uart = UARTDevice(Port.A, power_pin = 1)
```
resulting in P1 powered and

```
uart = UARTDevice(Port.A, power_pin = 0)
```
resulting in P1 nor P2 powered.

Fixes https://github.com/pybricks/support/issues/2655

### DIFF
--- a/pybricks/iodevices/pb_type_uart_device.c
+++ b/pybricks/iodevices/pb_type_uart_device.c
@@ -53,8 +53,9 @@ static mp_obj_t pb_type_uart_device_make_new(const mp_obj_type_t *type, size_t n
     PB_PARSE_ARGS_CLASS(n_args, n_kw, args,
         PB_ARG_REQUIRED(port),
         PB_ARG_DEFAULT_INT(baudrate, 115200),
-        PB_ARG_DEFAULT_NONE(timeout));
-
+        PB_ARG_DEFAULT_NONE(timeout),
+        PB_ARG_DEFAULT_INT(power_pin, 0)
+        );
     // Get device, which inits UART port
     pb_type_uart_device_obj_t *self = mp_obj_malloc(pb_type_uart_device_obj_t, type);
 
@@ -74,6 +75,14 @@ static mp_obj_t pb_type_uart_device_make_new(const mp_obj_type_t *type, size_t n
     pbio_port_id_t port_id = pb_type_enum_get_value(port_in, &pb_enum_type_Port);
     pb_assert(pbio_port_get_port(port_id, &self->port));
     pbio_port_set_mode(self->port, PBIO_PORT_MODE_UART);
+
+    if (mp_obj_get_int(power_pin_in) == 1) {
+       pbio_port_p1p2_set_power(self->port, PBIO_PORT_POWER_REQUIREMENTS_BATTERY_VOLTAGE_P1_POS);
+    } else if (mp_obj_get_int(power_pin_in) == 2) {
+       pbio_port_p1p2_set_power(self->port, PBIO_PORT_POWER_REQUIREMENTS_BATTERY_VOLTAGE_P2_POS);
+    } else
+       pbio_port_p1p2_set_power(self->port, PBIO_PORT_POWER_REQUIREMENTS_NONE);
+
     pb_assert(pbio_port_get_uart_dev(self->port, &self->uart_dev));
     pb_type_uart_device_set_baudrate(MP_OBJ_FROM_PTR(self), baudrate_in);
     pbdrv_uart_flush(self->uart_dev);


laurensvalk:
The tricky part about allowing this is that we won't have a way to automatically turn off the power when a custom device is unplugged.

We _could_ turn power off at the end of the program for safety reasons, but I imagine you'd want to keep it on between runs just like the Ultrasonic Sensor.

A few ideas:
- We could require some kind of keep-alive message to keep the power on. But this does require a background process on the hub too and some kind of task on the sensor side.
- We could add a safety prompt the first time the user applies this. Something like: `Using custom electronics may damage your LEGO hub. Press 'y' to proceed or 'n' to cancel`. We could make the user choice persist so we only ask the first time. This is simpler for both the hub and the custom sensor.

laurensvalk:
I have pushed an alternative branch [here](https://github.com/pybricks/pybricks-micropython/tree/power-pin) to implement the second option discussed above.

The ready made firmware files can be download [here](https://github.com/pybricks/pybricks-micropython/actions/runs/26219173184).

The `UARTDevice` now takes a `power_in` argument.

I do not currently have the equipment to test this. @ste7anste7an, could you please test the following on all hubs? 

Please confirm the polarity for `1` and `2`. Please also test that it returns to off when you choose `0`. It should ask for permission only the first time after you update the firmware. On EV3, only `1` should do something, `2` turns it off.

I think we can include this in the next release after full testing.

ste7anste7an:
Thank you for looking a this pull request. The warning is a good option. I have a Spike prime, a Lego technic hub and an EV3 hub. I will try it for all these hubs and do some tests with polarity and switching external power on and off (will do within a few days).

dlech:
> * We could add a safety prompt the first time the user applies this. Something like: `Using custom electronics may damage your LEGO hub. Press 'y' to proceed or 'n' to cancel`. We could make the user choice persist so we only ask the first time. This is simpler for both the hub and the custom sensor.

People don't always have an interactive terminal, e.g. when running with `pybricksdev`, so I'm not sure implementing this on the hub itself would be the most robust option.

Maybe we could just add another argument for controlling if power stays on after the program exits/device is disconnected?

laurensvalk:
It's a one-off safety confirmation that persists, so Pybricksdev users won't have to deal with this at runtime.

It's not really a convenience/technical issue, more of an awareness issue.

I think that the stock sensors are all protected against this, but other third-party sensors might not be.

ste7anste7an:
I checked the `power_pin` keyword for `UARTDevice` on primehub, technichub and EV3. the power pins are called M1 and M2. M1 is pin1 on the LPF2 Prime hub connector; M2 is pin 2.

For both primehub and technichub the keyword works as expected.
- It shows the warning 'Using custom electronics may damage your LEGO hub' the first time
- it accepts the values 0 (no power on M1 or M2), 1 (power on M1) and 2 (power on M2). All other values result in `ValueError: Invalid argument`.

When the programs stops, power is also turned off on any of the pins two poiwer pins.

For EV3, the keyword is not supported, as expected, because the EV3 has seperate ports (A..D) for motor output.

Thanks for building the firmwares. This feature allows for raw UART communicartion while not being dependent on the PUP protocol to feed power to a sensor.

laurensvalk:
Thanks for testing!

> For EV3, the keyword is not supported, as expected, because the EV3 has seperate ports (A..D) for motor output.

It should provide battery voltage on P1 of the sensor ports when enabled. Does this not work?

> When the programs stops, power is also turned off on any of the pins two poiwer pins.

Correct. We could potentially allow this in the future, but it sounds like your sensors work fine as proposed?

ste7anste7an:
I tested it with EV3 and I noticed that I got an error because the keyworkd 'power_pin' was not imlemented. I will double check to see that I flahed the EV3 with the correct firmware.

laurensvalk:
Thanks for the update. All candidate firmware with this feature [are here](https://github.com/pybricks/pybricks-micropython/actions/runs/26442757064). So for the EV3 it is [this file](https://github.com/pybricks/pybricks-micropython/actions/runs/26442757064/artifacts/7213010239).

ste7anste7an:
Tested again on EV3. Now loaded correct firmware. These are the voltages measured on P1 and P2:

```
u=UARTDevice(Port.S1,power_pin=0)
P1: 5.05
P2: 3.24

u=UARTDevice(Port.S1,power_pin=1)
P1: 6.63
P2: 3.24

u=UARTDevice(Port.S1,power_pin=1)
P1: 5.05
P2: 3.24
```

I want to do some tests under load. I think, looking at the schematics, that these pins can not really deliver significant amounts of current and are merely be meant for probing sensors.

laurensvalk:
In your script, was your third test really with `1` again or was that a typo for `2`?

It is supposed to give battery voltage:

image

So it looks like this feature needs some more work. Although maybe your battery was just that low? It did increase for `power_pin=1`.

Maybe we can get @JorgePe to test https://github.com/ev3dev/ev3dev/issues/913 but on the new firmware with the `I2CDevice` class :smile: 

And if we are very lucky, we can even reproduce the following:

> @dlech David, you are my hero!!! It works!!!

ste7anste7an:
This is the schematic of the sensor port of the EV3. As you can see, R123 limits the current when I_ONA is high, and 9V gets put on the right side of R123. The current that can be drawn is very limited, in the order of 10-20 mA. But measuring the voltage while drawing a current is the best way to find out. Will do later.
image

ste7anste7an:
I connected a device via a 3v3 voltage regulator to pin 1, while applying `u=UARTDevice(Port.S1,power_pin=1)`. The device consumes less that 100mA. The voltage on P1 drops from 6.63V to less then 3V.

laurensvalk:
What was the battery voltage before and after that, perhaps as measured or reported by hub.battery.voltage?

ste7anste7an:
Before the hub voltage is 7.114V. While powering via power_pin=1, almost the same: 7.104V

ste7anste7an:
Did a small additional test. Configured the UART: `u=UARTDevice(Port.S1,power_pin=1)`
I connected a resistor of 100 Ohm between P1 and GND. When the voltage is 6.6V without load, now the voltage drops to 1.6V. The current is 16mA. This indicates exactly the problem when pin P1 would be used as a power source. Obviously, it is not meant to be used as such, and is originally used for probing different kind of sensors.

Conclusion: on EV3, the sensor ports can not deliver significant power through pin P1. All sensors are powered by the 4.7V power pin 4. This pin can deliver 100's of mA.

dlech:
> Obviously, it is not meant to be used as such, and is originally used for probing different kind of sensors.

Actually, it is for the NXT Ultrasonic sensor. You can read about this in the NXT hardware docs. 

You are right, clearly it wasn't designed to be used as a power supply.


> All sensors are powered by the 4.7V power pin 4

This should nominally be 5V. Internally, it is the same power supply as the USB VBUS.

laurensvalk:
But not only the NXT Ultrasonic, right? Like https://github.com/ev3dev/ev3dev/issues/913

Just so we can make some decisions here then --- I was hoping to unify these APIs, but maybe we shouldn't apply it to EV3 then?

We'll still want to keep it on the `I2CDevice` class though, so that's why I initially thought we could keep it the same between them for consistency.

laurensvalk:
Since the Powered Up portion of this works as intended and has been tested, I would be OK to merge the power-pin branch.

We can still refine the API for EV3 if we want to change this since it won't be part of the upcoming stable release. Any objections?

antonvh:
Great work. I tested it too on SPIKE. It asked confirmation once and then kept working. No objections.

dlech:
> But not only the NXT Ultrasonic, right?

Ah yes, I forgot about that one. And maybe it is also used for RCX sensors on NXT (RCX doesn't work on EV3 since pin 2 is not GND)

laurensvalk:
@antonvh and @ste7anste7an Thanks for testing. Can you help shed some more light on the use cases for this so we can better understand it?

I'm assuming that this feature is primarily useful for boards such as servo drivers. Am I right in assuming that their VCC power is fed from the hub port VCC and battery power is only used for the motors?

In other words, is it not an inconvenience but in fact a good thing that we turn the power off at the end of a program? If so, we should probably leave it at this and call it good :smile:

Or are there also cases where the power should ideally kept on in between runs, to keep powering a more power hungry processor for example?

laurensvalk:
While we are all here, let's document it once and for all. Does anyone want to review this draft? Please quadruple check the orientation of the Powered Up connector :smile: 

I suppose we should also add the voltages for the data pins. Are they 3.3V on Powered UP? And is it 3.3V for EV3 despite the 5V vcc?

image

Sirus-7:
We should mention the following regarding the EV3:

“The output voltage is limited by the 330-ohm source resistor.”

Otherwise, questions will arise due to the current-dependent output voltage.

antonvh:
I can confirm. I wired OpenMV clock/p4/tx to PUP wire 6 (hub RX/sensor TX)
And openmv data/p5/rx to PUP wire 5  (hub TX/sensor RX)
And UART worked like this with the new UARTDevice!

For I2C, the OpenMV wires have to be switched around then :/ 
Is I2C still on the roadmap for PoweredUp?

https://docs.openmv.io/_images/pinout-openmv-ae3-pag7936.png

ste7anste7an:
The main use case for enabling power on the motor pins (Pin 1 and Pin 2) is to provide proprietary sensors or boards connected to a LEGO hub with a suitable power source. The VCC pin (Pin 4) on the LEGO SPIKE Prime and Technic Hub only provides 3.3 V and is limited in the amount of current it can supply. For applications that require higher voltages or higher currents, the motor outputs can provide sufficient power. Examples include servo motors and NeoPixels (e.g. WS2812 LEDs).

In one application, we use a buck converter to regulate the 8 V motor output voltage to a stable 5 V.

Until now, power on the motor pins could only be enabled by emulating a PUP sensor and requesting power on one of the pins through the PUP protocol. This requires the connected proprietary equipment to emulate the PUP protocol when interfacing with the SPIKE hub.

With the addition of the `power_pin` keyword, power can now also be provided when using UARTDevice. This simplifies communication considerably.

For some devices (e.g. NeoPixels), it would be beneficial if power remained enabled between runs. This also prevents connected processors (for example, in cameras) from having to restart every time a Pybricks program starts. Therefore, I would prefer power to remain enabled once it has been turned on.

If this behavior could cause issues for other applications, perhaps an additional keyword such as `keep_power` could be introduced to control this behavior. Alternatively, users could explicitly disable power in their program before exiting.

ste7anste7an:
> While we are all here, let's document it once and for all. Does anyone want to review this draft? Please quadruple check the orientation of the Powered Up connector 😄
> 
> I suppose we should also add the voltages for the data pins. Are they 3.3V on Powered UP? And is it 3.3V for EV3 despite the 5V vcc?
> 

Checked the numbering of pins, and they are correct. Maybe you can add also the pinout for EV3?

Furthermore, you could add a desciption of wthat the keyword 'power_pin' actually does:

The `power_pin` keyword controls power delivery on the motor connector pins and can take the following values:

0 – No power is supplied on either Pin 1 or Pin 2.
1 – Power is supplied on Pin 1.
2 – Power is supplied on Pin 2.

When power is enabled on either pin, the selected pin becomes the positive power output and can be used to supply external devices connected to the port. The voltage provided is the hub's motor supply voltage. 
Only one of the two pins can be powered at a time. Specifying 0 disables power output on both pins.

laurensvalk:
Right, something like that has been added (just not in the screenshot).

Could you add some discussion about GND in this case? I.e. the implications and/or mistakes for using the other P1/P2 pin vs. the dedicated ground pin.

ste7anste7an:
In our application, we use only P2 to power the external device and connect the GND of the device to GND Pin 3 of the Spike prime. I assume this will not cause any problems. The Spike prime uses LB1836 (https://www.onsemi.com/pub/Collateral/LB1836M-D.PDF) dual motor drivers, and the motors are conected between P1 and P2. I assume that when either P1 and P2 is not driven, they are effectively connected to ground. Furthermore I assume that Pin 3 can handle the current flowing back when a device is connected between P2 and GND.

antonvh:
> Furthermore I assume that Pin 3 can handle the current flowing back when a device is connected between P2 and GND.

Actually, that's something we could verify tomorrow in our current and voltage tests.

laurensvalk:
A new full beta release at https://beta.pybricks.com/ is out. For you it contains:
- The new `power_pin` feature landed in the included firmware.
- LPF2 protocol fixes (would still be great to know if this resolves your earlier issues even if you switch to raw uart).
- Draft documentation and autocomplete for `UARTDevice`, including a first draft of the screenshot above.

:tada:

antonvh:
> LPF2 protocol fixes (would still be great to know if this resolves your earlier issues even if you switch to raw uart).

Yes. it fixes earlier issues. LPF2 worked fine now. I think it even works after creating an UARTDevice on that port. But I'll doublecheck that. It might have been that the program stopped, UARTDevice got garbagecollected and then LPF2 succeeded.

laurensvalk:
Glad to hear the original issue was fixed!

Right - Instantiating a `UARTDevice` will disable the LPF2 protocol, as well as the auto-id algorithm that pokes at the same pins to figure out what is attached. LEGO mode is restored as soon as you initialize any class that uses it (`ColorSensor`, `...`, `PUPDevice`) They can be used in the same program in principle, but not at the same time.

Sirus-7:
Applies only to EV3:

As the schematic shows and @ste7anste7an  describes, P1 is not really suitable as a power supply. 
Under normal conditions, P1 is connected to 5 VDC via a 10k resistor. This voltage can be measured using an ADC. By connecting it externally to GND = 0 V or a resistor (e.g., 10k = 2.5 V), it can be used to detect different sensors. 

When power_pin=1, VBat is applied to P1 via the 330-ohm resistor R123. The voltage can be measured via the ADC.

P2 is only a digital input. When unconnected, 3.2 VDC is applied here via a voltage divider (AM1802 EMA_WEN DQM1/GP2.2).

Therefore, only P1 can be used as a power supply, taking into account the limited current draw of approx. 10 mA at 6 VDC.

_Translated with DeepL.com (free version)_

laurensvalk:
This has been completed, merged separately.

PR #469, opened 2026-04-14
bricks/ev3: Add NXT ultrasonic sensor on Portview

Sirus-7:
I hope I did that right. It's my first try.
It works on my EV3.


laurensvalk:
Thanks for giving this a try! 

There is however a reason why the NXT Ultrasonic Sensor was not yet included. 

Automatic detection for I2C devices isn't fully finished yet.

PR #489, opened 2026-09-04
micropython: Update submodule.

dlech:
Pull in an optimization to avoid unused TextWrapper qstr.


dlech:
I might see if upstream MicroPython accepts this first. It is a pretty small win, so not a big deal.

dlech:
I think this actually got picked up when we updated to MicroPython 1.29.

PR #496, opened 2026-09-07
MicroPython v1.29

dlech:
Since MicroPython v1.29 just came out recently, it made me realize we are a few versions behind.

Unfortunately, movehub loses a few hundred bytes that we just gained, but all other hubs go down in size.

Haven't tested on an actual hub yet.


laurensvalk:
For future micropython bumps, can we include the full path to their shas in the pybricks-micropython commit?

GitHub is not good at displaying it, and you have to jump through many clicks just to find out what changed.

image

laurensvalk:
Merged, thanks!

PR #498, opened 2026-09-08
bricks/buildhat: Set MICROPY_QSTRDEFS_PORT for shared qstrs.

### LINKED ISSUE #2824: [Bug] Build HAT: "import pybricks.hubs" fails, only pybricks_dot_hubs exists
**Describe the bug**
On the Raspberry Pi Build HAT firmware, the `pybricks.*` modules are registered under the wrong names: the qstr `MP_QSTR_pybricks_dot_hubs` resolves to the literal string `"pybricks_dot_hubs"` instead of `"pybricks.hubs"`. So `import pybricks.hubs` raises `ImportError`, while `import pybricks_dot_hubs` works.

It affects every dotted module in the build, not just `hubs`: in `build/genhdr/qstrdefs.generated.h` from master, all seven `MP_QSTR_pybricks_dot_*` entries (experimental, geometry, hubs, iodevices, messaging, parameters, tools) map to `"pybricks_dot_*"` strings and no `"pybricks.*"` string exists.

**To reproduce**
1. Build `bricks/buildhat` from master and load it into the HAT with `pybricks-build-hat`.
2. In the REPL over the Pi UART, run `import pybricks.hubs`.
3. See `ImportError: no module named 'pybricks.hubs'`.
4. Run `import pybricks_dot_hubs` — it succeeds, showing the module is present but registered under that name.

**Expected behavior**
`import pybricks.hubs` (and `from pybricks.hubs import BuildHat`) should work as on other hubs.

**Cause**
The Build HAT is the only CMake-built target. Makefile targets get the shared Pybricks qstr definitions from `bricks/_common/qstrdefs.h` via `QSTR_DEFS` in `common.mk`. The CMake build never sets the equivalent variable, `MICROPY_QSTRDEFS_PORT`, so that file is not read. Without it, the qstr generator falls back to the identifier name and stores the string as `"pybricks_dot_hubs"`. Setting the variable is a one-line fix, verified on hardware; PR to follow.

davelongdev:
The Build HAT is the only CMake-built target. Makefile targets receive the shared Pybricks qstr definitions via QSTR_DEFS in bricks/_common/common.mk, but the CMake build never set the equivalent MICROPY_QSTRDEFS_PORT. The qstr generator then used "pybricks_dot_*" strings for all seven MP_QSTR_pybricks_dot_* identifiers, so "import pybricks.hubs" failed and only the "pybricks_dot_hubs" spelling worked.

Verified on a Build HAT loaded from a Raspberry Pi 5: all seven dotted modules in the build (experimental, geometry, hubs, iodevices, messaging, parameters, tools) now import, and "from pybricks.hubs import BuildHat" works.

Firmware size: 152520 -> 152488 bytes (-32).

Fixes https://github.com/pybricks/support/issues/2824

### DIFF
--- a/bricks/buildhat/CMakeLists.txt
+++ b/bricks/buildhat/CMakeLists.txt
@@ -45,6 +45,11 @@ set(MICROPY_PORT_DIR ${MICROPY_BOARD_DIR})
 
 set(MICROPY_USER_FROZEN_MANIFEST ${MICROPY_FROZEN_MANIFEST})
 
+# Shared Pybricks qstrs (e.g. "pybricks.hubs"). Without this, only the
+# auto-collected "pybricks_dot_hubs" style names exist and
+# "import pybricks.hubs" fails.
+set(MICROPY_QSTRDEFS_PORT ${TOP_DIR}/bricks/_common/qstrdefs.h)
+
 string(TOLOWER ${MICROPY_BOARD} PICO_BOARD)
 
 # Set the amount of C heap, if it's not already set.


laurensvalk:
Thanks!

PR #499, opened 2026-09-08
pbio/sys: Add Build HAT HMI that starts the status light animation.

### LINKED ISSUE #2825: [Bug] Build HAT: status light never starts the ready animation at boot
**Describe the bug**
When the Build HAT first powers up, the bootloader shows a steady red LED while it waits for firmware. After loading Pybricks, the LED goes off, even though `lib/pbio/platform/build_hat/pbsysconfig.h` enables `PBSYS_CONFIG_STATUS_LIGHT_STATE_ANIMATIONS` with `PBSYS_CONFIG_STATUS_LIGHT_STATE_ANIMATIONS_HUE` set to green (120). The HAT gives no clear indication that Pybricks was successfully loaded.

**To reproduce**
1. Build `bricks/buildhat` from master and load it into the HAT with `pybricks-build-hat`.
2. Watch the status LED: it turns dark, yet the REPL over the Pi UART works, so the firmware is running.

**Expected behavior**
The LED breathes green (the configured hue) once the firmware is ready.

**Cause**
The Build HAT uses `lib/pbio/sys/hmi_none.c` (no buttons, no display), whose `pbsys_hmi_await_program_selection()` requests the REPL directly. Unlike `hmi_pup.c`, it never calls `pbio_color_light_start_breathe_animation()` first, so nothing ever starts the animation. `PBSYS_CONFIG_HMI_NONE` is set only by the Build HAT, so other hubs are unaffected. Fix verified on hardware; PR to follow.

**Note on future light states**
The Build HAT currently has no idle state.  With no buttons and no program storage, `hmi_none.c` goes straight to the REPL, so the "ready" breathe is the only state its light has. In the future, we may want three states: waiting for a host connection (through the Pi), connected with no program running, and program running. I think that is better addressed when the Build HAT port is more mature. This fix only starts the animation that is already configured.

davelongdev:
The Build HAT used the "none" HMI, which requested the REPL without first starting the status light animation, so the LED stayed off after boot even though the platform enables state animations with a green hue.

Rather than change hmi_none.c, which should stay the variant for hubs with no user interface at all, add hmi_buildhat.c selected by a new PBSYS_CONFIG_HMI_BUILDHAT. It is a copy of hmi_none.c that starts the breathe animation (or turns the light off when animations are disabled) before requesting the REPL, as hmi_pup.c does after program selection. The Build HAT has a status light but no buttons, display, Bluetooth or USB, so its user interface will need its own design later; this only starts the animation that is already configured.

Verified on a Build HAT loaded from a Raspberry Pi 5: the LED breathes green as soon as the firmware is ready.

Firmware size: 152188 -> 152316 bytes (+128).

Fixes https://github.com/pybricks/support/issues/2825

Edited 2026-09-09: reworked after review. The first version changed hmi_none.c directly; this version adds hmi_buildhat.c instead, as suggested in the comments below.

### DIFF
--- a/lib/pbio/sys/hmi_buildhat.c
+++ b/lib/pbio/sys/hmi_buildhat.c
@@ -1,0 +1,54 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2018-2026 The Pybricks Authors
+
+// Provides the Human Machine Interface (HMI) for the Build HAT.
+// There is nothing to select, so the REPL starts right after boot.
+
+#include <pbsys/config.h>
+
+#if PBSYS_CONFIG_HMI_BUILDHAT
+
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#include <pbio/button.h>
+#include <pbio/light.h>
+#include <pbio/os.h>
+#include <pbsys/light.h>
+#include <pbsys/main.h>
+#include <pbsys/status.h>
+
+void pbsys_hmi_init(void) {
+}
+
+void pbsys_hmi_deinit(void) {
+}
+
+void pbsys_hmi_stop_animation(void) {
+}
+
+void pbsys_hmi_connection_changed_handler(void) {
+}
+
+pbio_error_t pbsys_hmi_await_program_selection(void) {
+
+    do {
+        if (pbsys_status_test(PBIO_PYBRICKS_STATUS_SHUTDOWN_REQUEST)) {
+            return PBIO_ERROR_CANCELED;
+        }
+        pbio_os_run_processes_and_wait_for_event();
+    } while (pbdrv_button_get_pressed());
+
+    // Start the "ready" animation, same as hmi_pup.c does after program
+    // selection.
+    #if PBSYS_CONFIG_STATUS_LIGHT_STATE_ANIMATIONS
+    pbio_color_light_start_breathe_animation(pbsys_status_light_main, PBSYS_CONFIG_STATUS_LIGHT_STATE_ANIMATIONS_HUE);
+    #elif PBSYS_CONFIG_STATUS_LIGHT
+    pbio_color_light_off(pbsys_status_light_main);
+    #endif
+
+    return pbsys_main_program_request_start(PBIO_PYBRICKS_USER_PROGRAM_ID_REPL, PBSYS_MAIN_PROGRAM_START_REQUEST_TYPE_BOOT);
+}
+
+#endif // PBSYS_CONFIG_HMI_BUILDHAT
--- a/bricks/_common/sources.mk
+++ b/bricks/_common/sources.mk
@@ -242,6 +242,7 @@ PBIO_SRC_C = $(addprefix lib/pbio/,\
 	sys/battery.c \
 	sys/command.c \
 	sys/core.c \
+	sys/hmi_buildhat.c \
 	sys/hmi_ev3.c \
 	sys/hmi_ev3_ui.c \
 	sys/hmi_none.c \
--- a/lib/pbio/platform/build_hat/pbsysconfig.h
+++ b/lib/pbio/platform/build_hat/pbsysconfig.h
@@ -10,7 +10,7 @@
 #define PBSYS_CONFIG_BATTERY_CHARGER                (0)
 #define PBSYS_CONFIG_HMI                            (1)
 #define PBSYS_CONFIG_HMI_STOP_BUTTON                (0) // does not exist but pbsys_main() requires this to be defined
-#define PBSYS_CONFIG_HMI_NONE                       (1)
+#define PBSYS_CONFIG_HMI_BUILDHAT                   (1)
 #define PBSYS_CONFIG_HMI_NUM_SLOTS                  (0)
 #define PBSYS_CONFIG_HOST                           (1)
 #define PBSYS_CONFIG_HOST_EVENT_OUT_SIZE            (512)


dlech:
We probably want to save `hmi_none` for when there isn't any user interface at all. So instead would be better to make our own hmi for the build hat instead (so start a new `hmi_buildhat` file). 

We can start with a copy of `hmi_none` with this change though and then discuss what the actual UX should be later. Since it doesn't have Bluetooth or USB or any buttons, it is going to have to be different from anything else we have done before.

davelongdev:
Sounds good.  Thanks.

dlech:
FYI, you can just force-push to this same branch to update the PR instead of creating a new one when you are ready. (`git push -f`)

davelongdev:
Updated the PR.

BertLindeman:
Probably not this PR but there is no REPL and a simple program like

```python
# print("PR#500")
from pybricks import version
print(version)
```
Disconnects without any traceback.

dev console:
```
main.b6b2f3a2.js:2 Attempting to connect to City PR#500 (attempt 1)
[Violation] Forced reflow while executing JavaScript took 37ms
main.b6b2f3a2.js:2 Downloading 103 bytes
main.b6b2f3a2.js:2 Attempting to connect to City PR#500 (attempt 1)
498.d505946f.chunk.js:2 [Violation] 'mousedown' handler took 489ms
7346.cbc7d460.chunk.js:1 copied PR500_print.py to emscripten fs
7346.cbc7d460.chunk.js:1 copied PR500_print.py to emscripten fs
7346.cbc7d460.chunk.js:1 copied PR500_print.py to emscripten fs
7346.cbc7d460.chunk.js:1 copied PR500_print.py to emscripten fs
main.b6b2f3a2.js:2 Downloading 78 bytes
498.d505946f.chunk.js:2 [Violation] 'mousedown' handler took 536ms
7346.cbc7d460.chunk.js:1 copied PR500_print.py to emscripten fs
7346.cbc7d460.chunk.js:1 copied PR500_print.py to emscripten fs
main.b6b2f3a2.js:2 Downloading 88 bytes
498.d505946f.chunk.js:2 [Violation] 'mousedown' handler took 525ms
```

Will test ble first. My ble_broadcast-observe.py program runs OK.
But it does not end in a regular way, the program is not saved.
So for the not-connected case I need to use something like:
```
pybricksdev run --no-start ble --no-stay-connected ble_broadcast_transmit_demo.py
```
And loaded with pybricksdev I can run the program multiple times by pressing the hub button.

dlech:
@BertLindeman, I see "City" in your log. Did you reply to the wrong PR? This one is about the Raspberry Pi Build HAT.

BertLindeman:
> @BertLindeman, I see "City" in your log. Did you reply to the wrong PR? This one is about the Raspberry Pi Build HAT.

Should have been in [PR 500](https://github.com/pybricks/pybricks-micropython/pull/500)
Have hidden the post.

laurensvalk:
I'm not opposed to adding this, but I guess it's worth doing some brainstorming about what the buildhat should eventually do. Would we want to run Pybricks programs on it, so the hat is semi-autonomous, sending relevant user-made data back and forth?

Or would we rather have it become a remote i/o platform where the host (real Python, or any other app or language) would continuously receive sensor info, and send commands to drive the motors?

Both are possible (and both have their unique merits), but this choice would roughly guide what the design would look like.

dlech:
> Would we want to run Pybricks programs on it,

Yes. 

> Or would we rather have it become a remote i/o platform

I think this would have too much latency. This is how RPi Foundation did it already and user feedback isn't great.

davelongdev:
Thanks Laurens, that's a helpful perspective. It seems like there are basically 3 options we might consider?  I would ideally like to build something like all 3 for different purposes at least to try them out as prototypes.  I say basically here because I think the boundaries are not totally clear cut: some aspects of what goes on the Pi vs. the BH might be shifted around within options 1 and 2.  

My thinking at this point is to start with option 2 below as David suggests because it seems like the best starting point (least work, the most practical, etc.), but I'm open to shifting gears if something else seems like a better option.

**1. BH as smart I/O device: the Pi does most of the work, the BH does as little as possible.**

**Look:** User's program runs on the Pi. The BH handles I/O. Pybricks or some of its components could be built for Linux (like the virtual hub).

**Pros:** One host program sees all BH ports, so two HATs on one Pi become one 8-port robot. Any host language could work, although options 2 and 3 could carry a different language on top of pbio too, since pbio itself is plain C. No program storage needed on the HAT. The BH firmware could be a slimmed-down option 2.

**Cons:** This is the most similar to what the official Build HAT firmware already does, but we could probably do it better. As David noted, latency complaints on this existing approach have been voiced.  I think some of this could likely be alleviated with better design choices and coding, but I think some things will probably run better / faster on the BH than on the Pi. 


**2. Pybricks on the HAT, Pi as loader and auxiliary.**

**Look:** User's program, Pybricks, and REPL run on the BH.  Pi loads the firmware into BH's 264 KB RAM and runs some aux functionality.  The BH's 2 MB flash is not used today, so nothing persists on it between power-ups.  Flash could be used, but would add the bricking risk discussed below.

**Pros:**

- **Lowest latency.** The program and the motor control loops run on the same chip with no serial link in the way. Motors and sensors would respond similarly to SPIKE or Technic hubs.
- **Least new work.** This is what the port already does today, so it builds on code that is written and tested on hardware.
- **Works like other Pybricks hubs.** Not as much of a special case.
- **Parts of option 2 can be reused in options 1 and 3.** Build HAT drivers are being written now for this option. The drivers and pbio's existing motor control that they enable, could stay on the HAT in all three. Option 1 could move MicroPython and other things to the Pi?
- **Could potentially use Pi's Bluetooth and wire buttons / other things into the Pi instead of the BH.**  
- **The Pi is free for heavy work.** Camera, vision, AI, logging or a dashboard could run on the Pi and talk to the program over the UART without slowing the robot.

**Cons:** A two HAT setup would be more like two hubs, rather than one hub with 8 ports; coordination would go through the Pi like a multi-hub setup. Without using   the 2 MB flash memory on BH, nothing would persist on BH after shutdown, and Pi would always need to be present.


**3. Standalone HAT, no Pi.**

**Look:** Pybricks and programs run on HAT alone in some combination of BH's 2 MB flash and 264 KB RAM.  Buttons possible with force sensor through LPF2 port or over UART.  Bluetooth / wifi possible over UART with external hardware.

**Pros:** Cheapest option / least hardware involved (at least w/o hardware beyond the BH).  Saving across BH power down would be possible.  Could access the full 2 MB flash memory on the BH.

**Cons:** Bricking risk that can only be remedied through debugging probe - the BH bootloader (BHBL) lives in flash, so it could be overwritten once flash is touched. This limits this option to more advanced users.  Startup would mean auto-run on power-up without extra hardware.

laurensvalk:
> > Would we want to run Pybricks programs on it,
> 
> Yes. 
> 
> > Or would we rather have it become a remote i/o platform
> 
> I think this would have too much latency. This is how RPi Foundation did it already and user feedback isn't great.

That's not really a fair comparison though. Its implementation ran at 4Hz under some conditions. 

We should be able to do this much, much faster. 

And with the time critical control loops running in our firmware on the hat, we should be able to reach the right balance so that throughput is never the critical factor. 

For reference, even a ball balancing robot runs great at 40 hz when the sensor integration runs internally at maximum sensor speed.

We might even be able to run the Pybricks API in real Python, async and all, with a lightweight link in between. 

All of this seems like it could work well for any hub connected to a computer, not just the build hat. Including daisy chained EV3 bricks.

dlech:
There really isn't a reason we couldn't do both either.

Just run a special built-in remote I/O program if you want that (started just like we start the special REPL program now) or send your own Python programs directly to the hub.

It isn't any more work either because the the latter is already done.

laurensvalk:
We'll soon have both in the firmware, no special built in remote I/O program required :wink: 

So back to the original question on the HMI, which does need some redoing. The `none_hmi` always starts the REPL, whereas we'd want to allow any program to be deployed and run, with the REPL being one option. So I guess it depends on what the host application will look like. That would be up to the user, but we could provide some glue to make it easier.

dlech:
I'm in favor of merging this as-is for now so we an move on to getting motors and sensors working first and then come back to the "how are users going to use this" question later.

laurensvalk:
Works for me!

Might want to remove the part that waits for buttons though if there aren't any buttons.

lib/pbio/sys/hmi_buildhat.c
```diff
@@ -0,0 +1,54 @@
...
+}
+
+pbio_error_t pbsys_hmi_await_program_selection(void) {
+
+    do {
+        if (pbsys_status_test(PBIO_PYBRICKS_STATUS_SHUTDOWN_REQUEST)) {
+            return PBIO_ERROR_CANCELED;
+        }
+        pbio_os_run_processes_and_wait_for_event();
+    } while (pbdrv_button_get_pressed());
```
dlech, line 41: As Laurens said, we don't have a button on this hub, so we should be able to drop this loop.

dlech (changes requested):

PR #500, opened 2026-09-09
drv/bluetooth: Scan at a 50% duty cycle so broadcasting works.

dlech:
All three Bluetooth drivers scanned with the window equal to the interval, 30ms of every 30ms, so the radio was never free and a hub that broadcasts while observing could hardly transmit. Measurements in https://github.com/orgs/pybricks/discussions/2822 show a City Hub dropping from 28 packets/s to under 4 when it broadcasts and observes at the same time.

Curiously it was fine when connected to a computer. The reason is that GAP_DeviceDiscoveryRequest() picks TGAP_CONN_SCAN_INT and _WIND when a connection exists and the general discovery parameters when it does not. We only set the latter, so the chip fell back to TI's defaults of 300ms interval and 150ms window, a 50% duty cycle, whenever a computer was connected. Using a 50% duty cycle always is what this does.

The scan interval is also kept shorter than the advertising interval of a peer hub. When the two periods are close, the phase between them drifts slowly, misses come in runs and reception latency is spiky rather than merely averaging out.

Broadcasting on the BTstack hubs went from every 100ms to every 30ms. 100ms was the Bluetooth 4.x minimum for non-connectable advertising, but the CC2564C is 5.1, where the minimum is 20ms, so the old value only limited how quickly other hubs could hear it. The Move Hub keeps 100ms because BlueNRG-MS is a 4.1 controller and cannot go faster; it is the most limited hub, so it is allowed to be the slow one rather than holding the others back.

Expected reception latency is now roughly 50 to 60ms between any pair of hubs, except from the Move Hub at about 200ms.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@
 - On the Move Hub, `bytes()` and `bytes.find()` now truncate out-of-range
   values instead of raising `ValueError`.
 
+### Fixed
+- Fixed slow broadcasting while observing at the same time ([support#2822]).
+- Fixed slow observing while connected to a computer or app ([support#2822]).
+
+[support#2822]: https://github.com/orgs/pybricks/discussions/2822
+
 [Unreleased]: https://github.com/pybricks/pybricks-micropython/compare/v4.1.0b3...HEAD
 
 ### Fixed
--- a/lib/pbio/drv/bluetooth/bluetooth_btstack.c
+++ b/lib/pbio/drv/bluetooth/bluetooth_btstack.c
@@ -1156,7 +1156,8 @@ pbio_error_t pbdrv_bluetooth_start_broadcasting_func(pbio_os_state_t *state, voi
     }
 
     bd_addr_t null_addr = { };
-    gap_advertisements_set_params(0xA0, 0xA0, PBIO_BLUETOOTH_AD_TYPE_ADV_NONCONN_IND, 0, null_addr, 0x7, 0);
+    // Advertise every 30ms so that other hubs receive broadcasts quickly.
+    gap_advertisements_set_params(0x30, 0x30, PBIO_BLUETOOTH_AD_TYPE_ADV_NONCONN_IND, 0, null_addr, 0x7, 0);
     recorded_events.advertise_enable_complete = false;
     gap_advertisements_enable(true);
 
@@ -1175,7 +1176,10 @@ pbio_error_t pbdrv_bluetooth_start_observing_func(pbio_os_state_t *state, void *
     PBIO_OS_ASYNC_BEGIN(state);
 
     if (!pbdrv_bluetooth_is_observing) {
-        gap_set_scan_params(0, 0x30, 0x30, 0);
+        // 70ms interval, 35ms window. The 50% duty cycle leaves radio time
+        // for broadcasting while observing, and the longer interval spends
+        // less of the radio on starting and ending scans.
+        gap_set_scan_params(0, 0x70, 0x38, 0);
         gap_start_scan();
         pbdrv_bluetooth_is_observing = true;
         // REVISIT: use callback to await operation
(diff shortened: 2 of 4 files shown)


laurensvalk:
It sounds like this was quite extensively tested. Thank you! I would be happy to merge without further testing from my side.

> Measurements in https://github.com/orgs/pybricks/discussions/2822 show a City Hub dropping from 28 packets/s to under 4 when it broadcasts and observes at the same time.

I've been trying to find the video corresponding to [this test](https://github.com/pybricks/pybricks-projects/issues/96) somewhere in the issue pages but haven't found it yet. 

It was showing that using the Powered Up remote with one City train and broadcasting to the other was essentially instant. It would be nice to reproduce something like that as a quick extra test. Maybe it as just that fast because each hub had only one role?

dlech:
> Maybe it was just that fast because each hub had only one role?

Yes, things worked really great before if one hub was only broadcasting and one hub was only observing.

And broadcasting while connected worked really good before too.

So that demo was doing all of the things that work really well. It was all of the other cases that didn't work so well.

dlech:
https://github.com/user-attachments/assets/921259bf-91ea-4ba5-970c-bdee41b4e896

https://github.com/user-attachments/assets/baab350a-c00e-401f-a453-4ad6035ca0f9

dlech:
In one of these videos, both hubs are connected to Pybricks Code and in one they are not. Can anyone tell the difference?

(And yes, the remote is connected to the technic hub, which is broadcasting and the city hub is only observing.)

dlech:
```py
# Remote control over broadcast, the sending half. Broadcasts only.
#
# This hub is connected to a Powered Up Handset and broadcasts the motor speed
# the buttons are asking for. It never observes. Run remote_motor_rx.py on a
# second hub, which drives the motor.
#
# Worth knowing what this exercises: the Handset is a real Bluetooth
# connection, so this hub is broadcasting while connected even though no
# computer is involved. That is the case the connected scan settings fix
# addresses, and pressing a button is a good way to feel the difference.
#
# Buttons:
#   left +/-    full speed forwards or backwards
#   right +/-   a third of that, for finer control
#   nothing     stop

from pybricks.parameters import Button, Color
from pybricks.pupdevices import Remote

CHANNEL = 1
SPEED = 900

hub = ThisHub(broadcast_channel=CHANNEL)

hub.light.on(Color.YELLOW)
remote = Remote()
hub.light.on(Color.GREEN)

last = None

while True:
    pressed = remote.buttons.pressed()

    if Button.LEFT_PLUS in pressed:
        speed = SPEED
    elif Button.LEFT_MINUS in pressed:
        speed = -SPEED
    elif Button.RIGHT_PLUS in pressed:
        speed = SPEED // 3
    elif Button.RIGHT_MINUS in pressed:
        speed = -SPEED // 3
    else:
        speed = 0

    # Only send on a change. The radio keeps advertising the last value on its
    # own, so repeating it would just spend Bluetooth commands to say the same
    # thing, and setting the remote's light every cycle would do the same.
    if speed != last:
        hub.ble.broadcast(speed)
        remote.light.on(
            Color.GREEN if speed > 0 else Color.RED if speed < 0 else Color.NONE
        )
        last = speed

    wait(10)
```

```py
# Remote control over broadcast, the receiving half. Observes only.
#
# Drives a motor at whatever speed remote_motor_tx.py is broadcasting. This hub
# never transmits, so all of its radio time goes to listening.
#
# The status light shows what the link is doing: green while driving, yellow
# while stopped on request, red when nothing has been heard for a second.

from pybricks.parameters import Color, Port

CHANNEL = 1

hub = ThisHub(observe_channels=[CHANNEL])
motor = Motor(Port.A)

hub.light.on(Color.RED)
applied = None

while True:
    speed = hub.ble.observe(CHANNEL)

    if speed is None:
        # observe() gives None when nothing has arrived for a second, which
        # means the link is gone rather than someone asking for a stop. Coast
        # instead of braking, so a moving robot is not stopped dead by a hub
        # going out of range.
        if applied is not None:
            motor.stop()
            hub.light.on(Color.RED)
            applied = None

    elif isinstance(speed, int) and speed != applied:
        # Anyone can broadcast on this channel, so ignore anything that is not
        # a speed rather than raising in the middle of a movement.
        if speed == 0:
            motor.brake()
            hub.light.on(Color.YELLOW)
        else:
            motor.run(speed)
            hub.light.on(Color.GREEN)
        applied = speed

    wait(10)
```

Note: These are AI generated programs (because time is valuable). I think the first one is pretty good. The second one is mostly good but I would just do `else:` instead of `elif isinstance(speed, int) and speed != applied:` (and remove all of the `applied` stuff). It seems a bit _extra_. It doesn't hurt to call motor or light methods over and over. The hub has nothing better to do with it's time anyway. And silently ignoring the wrong type of data isn't great. Wrong data should fail loudly, IMHO.

BertLindeman:
WOW!

PR #501, opened 2026-09-11
pbio/platform/build_hat: Reclaim bootloader RAM for the GC heap.

### LINKED ISSUE #2484: [Feature] Build HAT RAM usage
Currently, the way the Raspberry Pi Build HAT works is that when it starts up, it runs the the Build HAT Bootloader (BHBL) from the flash memory on the Build HAT. (Refer to #2482 for discussion about flash memory.) The BHBL is then used to load the actual firmware into RAM on the Build HAT and run it from there.

There is 256KiB RAM, so this should be plenty for both the Pybricks firmware and user programs and a decent heap memory size for user programs. The Build HAT doesn't have things like Bluetooth, so the firmware size won't be quite as big as on other hubs. Currently, the Pybricks firmware size is sitting at less than 150KiB.

We copied the linker script from the official Build HAT source code. This reserves part of the RAM for the bootloader. However, I don't think that this is necessary. We need to make sure that the firmware size fits in the RAM minus the bootloader size so that when the bootloader is loading the firmware, it doesn't write over itself.  But after the bootloader jumps to the firmware, that area of the RAM previously occupied by the bootloader should be free to use.

One way we could do this to to have 2 RAM sections in the linker script and put uninitialized data in the 2nd one. Or we might just be able to have one RAM section and just have an assert to catch if .text + .data size gets over the allowable size.

davelongdev:
The Build HAT linker script limited RAM to 240k because the bootloader (BHBL) uses the top 16 KiB while it loads the firmware. Once BHBL jumps to the firmware it is no longer running, so the firmware is free to use that RAM.

Checked against the bootloader source (`bhbl-pico/bhbl.ld` in the official firmware repo): BHBL's RAM region is `ORIGIN = 0x2003c000, LENGTH = 0x3f00`.

Changes to `rpi_build_hat.ld`:

- `RAM` becomes `LENGTH = 256k`, so the GC heap, which runs to the end of the RAM region, gains the 16 KiB.
- A link-time `ASSERT(__data_end__ <= 0x2003C000)` makes the build fail if the loaded image (`.text` through `.data`) ever grows into the RAM that BHBL is using while it copies the image in. The `.bss` section and heap may extend above that address since the firmware's own startup code zeroes `.bss` after the jump.
- A header comment explaining the layout, replacing the "limit RAM use to 240K" note.

One region rather than a separate `.bss` region gives the whole 16 KiB to the heap and allows the `.bss` region to be sized as needed.

Verified on a Build HAT loaded from a Raspberry Pi 5: `micropython.mem_info()` GC total 80256 -> 96256 bytes.

Firmware size: 152188 -> 152188 bytes (unchanged).

Fixes https://github.com/pybricks/support/issues/2484

Edit: the first version kept the last 256 bytes out of the heap to preserve the firmware signature BHBL writes there. Dropped after review, since it's not needed.

### DIFF
--- a/lib/pbio/platform/build_hat/rpi_build_hat.ld
+++ b/lib/pbio/platform/build_hat/rpi_build_hat.ld
@@ -1,7 +1,19 @@
 /*
 Modified from
 pico-sdk/src/rp2_common/pico_standard_link/memmap_no_flash.ld
-to limit RAM use to 240K for use with bootloader.
+for use with the Build HAT bootloader (BHBL).
+
+The firmware runs entirely from RAM. While loading it, BHBL uses the top
+16 KiB of RAM (0x2003C000-0x20040000) for its own stack and data.
+
+Once BHBL hands control to the firmware, it is no longer running, so that
+16 KiB is unused. Thus, the RAM entry in the MEMORY code block in this file ends at
+the end of RAM rather than at 0x2003C000. Since the GC heap runs to the end
+of that entry, the reclaimed 16 KiB becomes heap.
+
+The loaded image (.text through .data) must still end below 0x2003C000, or
+BHBL would overwrite itself during loading. An ASSERT at the end of this
+file checks that at link time.
 */
 /* Based on GCC ARM embedded samples.
    Defines the following symbols for use by code:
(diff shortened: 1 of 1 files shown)


lib/pbio/platform/build_hat/rpi_build_hat.ld
```diff
@@ -28,7 +41,7 @@ to limit RAM use to 240K for use with bootloader.
 
 MEMORY
 {
-    RAM(rwx) : ORIGIN =  0x20000000, LENGTH = 240k      /* <--- changed here */
+    RAM(rwx) : ORIGIN =  0x20000000, LENGTH = 256k - 256
```
dlech, line 44: Why do we need to keep the firmware signature?

davelongdev: It was so small I figured we might as well leave it in case it turned out to be useful for something, but having thought it over I can't see how it would be. I'll update the PR.

davelongdev:
Updated the PR.

dlech:
Thanks!

PR #502, opened 2026-09-12
lib/aeabi_div: Add compact integer division helpers.

dlech:
Cortex-M0 has no divide instruction, so every 32-bit integer division calls a libgcc helper. Since GCC 11, the v6-m libgcc only ships speed-optimized helpers: __udivsi3 is fully unrolled and __divsi3 contains its own second copy of the unsigned core, together around 730 bytes. Provide small loop-based replacements that produce identical results, including on division by zero, and enable them on the Move Hub and City Hub. This saves around 640 bytes on each, at the cost of roughly 3 to 5 times slower division, which stays well below 1% of the CPU for the heaviest user (motor control).

The new file is compiled without LTO because the compiler generates references to these helpers during LTO code generation, after LTO symbol resolution, so definitions inside an LTO object fail to link.


dlech:
@laurensvalk, any bright ideas on how to test this to make sure it doesn't affect motor control in a bad way?

PR #504, opened 2026-09-12
pbio/drv/bluetooth: Broadcast on the Move Hub while connected.

### LINKED ISSUE #2826: [Bug] movehub at stable v4.0.1 firmware does not broadcast if PC connected
**Describe the bug**
A Move Hub running Pybricks v4.0.1 does no BLE broadcasting when it is connected to a PC via BLE.

The same program works correctly when the Move Hub is running standalone. 

A City Hub running the same firmware and the same broadcast program, broadcasts correctly while connected to a PC.

The Move Hub can still observe broadcasts while connected to the PC.

**To reproduce**
1. BLE connect the movehub with FW v4.0.1 to a PC
2. Run the example broadcast program (see below) on the movehub
2. run on any other hub the observe program
3. look at the color of the observer hub: 
    - GREEN = broadcast received
    - RED = no broadcast received

Repeat without the PC BLE connection. 

In that case the Move Hub broadcasts correctly and the observer shows GREEN.

**Expected behavior**
BLE broadcasting should continue while the hub has an active PC connection. 

If broadcasting is intentionally disabled while connected, this should be documented and/or indicated to the application and raise some kind of error..

**Additional information**

- Move Hub: Pybricks v4.0.1
- City Hub: Pybricks v4.0.1 - broadcasting works with PC connected.
- Move Hub observing: works both with and without a PC connection.

- The same behavior was also reproduced with the firmware built for https://github.com/pybricks/pybricks-micropython/pull/500.
- A separate passive BLE scanner running on a Raspberry Pi Pico 2 W confirms that the Move Hub's LEGO manufacturer-data advertisements disappear when the PC BLE connection is active. 
    This does not appear to be an issue in BLERadio.observe() on the receiving hub.

<details><summary>broadcast script </summary>

```python
from pybricks.messaging import BLERadio
from pybricks.parameters import Port

# Initialize the hub.
radio = BLERadio(broadcast_channel=1)

# Initialize the motors.
left_motor = Motor(Port.A)
right_motor = Motor(Port.B)

while True:
    # Read the motor angles to be sent to the other hub.
    left_angle = left_motor.angle()
    right_angle = right_motor.angle()

    # Set the broadcast data and start broadcasting if not already doing so.
    data = (left_angle, right_angle)
    # print(data)
    wait(10)  # needed due to https://github.com/pybricks/support/issues/2751
    radio.broadcast(data)

    # Broadcasts are only sent every 100 milliseconds, so there is no reason
    # to call the broadcast() method more often than that.
    wait(100)
```
</details>

<details><summary>observe script</summary>

```python
from pybricks.messaging import BLERadio

# Run this on the hub that is OBSERVING.
hub = ThisHub()
radio = BLERadio(observe_channels=[1])

while True:
    data = radio.observe(1)
    hub.light.on(Color.GREEN if data is not None else Color.RED)
    wait(100)
```
</details>

<details><summary>ble_sniffer_simple.py</summary>

```python
"""
ble_sniffer_simple.py
Minimal passive BLE scanner for Pico 2 W (MicroPython) to prove whether a
Pybricks hub is broadcasting (LEGO manufacturer data, company ID 0x0397).
Same script works unmodified for any hub - flash a hub with Pybricks
v4.0.1, start its broadcast program, run this, and watch for output.

No external dependencies - the LEGO manufacturer ID check (company ID
0x0397, i.e. bytes 151, 3) is inlined below.
"""
import time

import bluetooth
from micropython import const

_IRQ_SCAN_RESULT = const(5)
_ADV_TYPE_MANUFACTURER_DATA = const(0xFF)


def _is_lego(manu):
    try:
        return manu[2] == 151 and manu[3] == 3
    except IndexError:
        return False

_count = 0
_stop = False
_start_ms = time.ticks_ms()


def _find_adv_structure(payload, adv_type):
    i = 0
    n = len(payload)
    while i + 1 < n:
        length = payload[i]
        if length == 0:
            break
        if payload[i + 1] == adv_type:
            return bytes(payload[i:i + 1 + length])
        i += 1 + length
    return None


def _bt_irq(event, data):
    global _count, _stop
    try:
        if event == _IRQ_SCAN_RESULT:
            _, _, _, _, adv_data = data
            manu = _find_adv_structure(adv_data, _ADV_TYPE_MANUFACTURER_DATA)
            if manu and _is_lego(manu):
                channel = manu[4]
                _count += 1
                elapsed = time.ticks_diff(time.ticks_ms(), _start_ms)
                print("{:6d}ms  channel {}  count {}".format(elapsed, channel, _count))
    except KeyboardInterrupt:
        # Ctrl-C can land here instead of in the main loop's sleep_ms();
        # without this it's swallowed and the summary below never runs.
        _stop = True


ble = bluetooth.BLE()
ble.active(True)
ble.irq(_bt_irq)

print("Scanning for Pybricks LEGO broadcasts (Ctrl-C to stop)...")
ble.gap_scan(0, 30000, 30000, False)  # continuous, passive

try:
    while not _stop:
        time.sleep_ms(500)
except KeyboardInterrupt:
    pass
finally:
    ble.gap_scan(None)
    ble.active(False)
    elapsed_s = time.ticks_diff(time.ticks_ms(), _start_ms) / 1000
    print()
    if _count:
        print("RESULT: {} LEGO broadcast packets in {:.1f}s -> Received broadcasts".format(_count, elapsed_s))
    else:
        print("RESULT: 0 LEGO broadcast packets in {:.1f}s -> NO broadcast received".format(elapsed_s))
```
</details>

dlech:
The BlueNRG-MS GAP layer refuses to enter any advertising mode while a computer is connected, so aci_gap_set_non_connectable() answered BLE_STATUS_FAILED and nothing was broadcast. The driver mapped that one error to success so that programs using broadcasting could at least run.

Drive the link layer directly instead, with the set advertising parameters, set advertising data and set advertise enable sequence from UM1865 section 3.3. The chip is already configured for role mode 4, which supports advertising alongside connections and scanning, and it accepts all three commands while a computer is connected. Asking the chip first was no help: HCI_LE_Read_Supported_States answers with bit 20 clear, but it also claims not to support advertising while scanning and not to support scanning while a slave, both of which demonstrably work, so its answer understates what this stack can do.

Advertising is now enabled behind the GAP layer's back, so stopping has to disable it the same way, and the discoverable mode has to be left before the parameters can be set. The link layer advertises exactly the payload it is given, so the service UUID and TX power AD types left over from discovery no longer have to be deleted to make room. Broadcasts therefore no longer carry the Flags AD type, which is what the BTstack hubs already send.

The advertising interval has to be requested as a range. The chip fits a new activity to the anchor period it is already running for the connection and for any scanning, and answers 0x85
BLE_STATUS_INTERVAL_TOO_LARGE when no multiple of that period falls between the requested minimum and maximum. A fixed 100ms therefore failed the moment the hub was also observing or connected, which is precisely the case being fixed here. The GAP layer picked its own range internally and so never ran into it. 100ms is the fastest this Bluetooth 4.1 controller allows for non-connectable advertising and is the bottom of the range; 200ms at the top measures the same as the GAP layer managed, 2.45 against 2.53 broadcasts a second reaching a third hub, which is inside the run to run spread.

Drop the note about broadcasting interfering with observing as well. Measured on a Move Hub with the 50% scan duty cycle, reception is unchanged by broadcasting at the same time: 13.4 advertisements a second from a reference transmitter either way, with the same distribution of gaps. That measurement depends on the duty cycle, so this belongs after that change.

Fixes https://github.com/pybricks/support/issues/2826


dlech:
Note, this doesn't completely fix things on it's own. It will need to be rebased once #500 is merged.

So until then, use https://github.com/pybricks/pybricks-micropython/actions/runs/34675375120 for test firmware instead of the link in the comment below.

dlech:
> It will need to be rebased once #500 is merged.

Done

dlech:
Need more work. Advertising to connect after any program with broadcast is broken by this.

dlech:
https://github.com/user-attachments/assets/44846ede-e5f2-4682-8726-ad4221642058

dlech:
Some noticeable lag compared to https://github.com/pybricks/pybricks-micropython/pull/500#issuecomment-5642109058, but not too bad!

